PR #10, opened 2018-08-21
Wrong  declaration of case "miopenActivationPATHTRU" . It should be "…

pramenku:
…miopenActivationPASTHRU"

As per https://github.com/ROCmSoftwarePlatform/MIOpen/blob/master/include/miopen/miopen.h Line : 350.,
enum is defined as "miopenActivationPASTHRU" but in hipDNN , it is miopenActivationPATHTRU.
Due to this issue, hipDNN building is failing.

src/hcc_detail/hipDNN_miopen.cpp:374:10: error: use of undeclared identifier 'miopenActivationPATHTRU'; did you mean
      'miopenActivationPASTHRU'?
    case miopenActivationPATHTRU:
         ^~~~~~~~~~~~~~~~~~~~~~~
         miopenActivationPASTHRU
/opt/rocm/miopen/include/miopen/miopen.h:352:5: note: 'miopenActivationPASTHRU' declared here
    miopenActivationPASTHRU  = 0, /*!< No activation, pass through the data */
    ^
1 error generated.

### DIFF
--- a/src/hcc_detail/hipDNN_miopen.cpp
+++ b/src/hcc_detail/hipDNN_miopen.cpp
@@ -371,7 +371,7 @@ hipdnnStatus_t miopenTohipActivationMode(miopenActivationMode_t in,
         *out = HIPDNN_ACTIVATION_TANH;
         break;
 
-    case miopenActivationPATHTRU:
+    case miopenActivationPASTHRU:
         *out = HIPDNN_ACTIVATION_PATHTRU;
         break;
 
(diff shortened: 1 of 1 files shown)


streamhsa:
https://github.com/ROCmSoftwarePlatform/hipDNN/commit/5ed2d1666b42ad8f9f4aa1132b9a6cc0647d5c27
In this commit, wrong declaration was created which was reverted back to match with miopen declaration. Then, only it's building on rocm stack. 

Pramendra Kumar (streamhsa)

PR #18, opened 2018-10-11
Fusion api support

pradeepisro:
Added Fusion API support for AMD and fixes in CMake Script.


pradeepisro:
@NEELMCW please review it.

PR #20, opened 2018-10-25
Fusion api support

pradeepisro:
Added Unit tests For new layers a part of MIOpen 1.5


pradeepisro:
@NEELMCW. please review.

PR #53, opened 2025-05-15
Create initial MIOpen legacy plugin build

adickin-amd:
## Proposed changes

- Added new Ubuntu 22 dockerfile.  The RHEL8 dockerfile remains but isnt ideal for development due to its gdb version.
- Created a stub for MIOpen Legacy plugin.
- Made it so Cmake can build the plugin as part of the hipdnn build tree or as a Standalone plugin.

## Checklist


CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ set(CMAKE_CXX_STANDARD 20)
 option(HIP_DNN_BUILD_BACKEND "Build the backend code" ON)
 option(HIP_DNN_BUILD_FRONTEND "Build the frontend code" ON)
 option(HIP_DNN_SKIP_TESTS "Skips building all tests" OFF)
+option(HIP_DNN_BUILD_PLUGINS "Builds the plugins as part of the main cmake" ON)
```
adickin-amd: im thinking we want this on while we build the legacy plugin, wondering if I wait to turn it on while were still getting everything else in place.

dockerfiles/scripts/install_cmake.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# Copyright 2022 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
averinevg, line 6: We'll have to add `Apache-2.0 WITH LLVM-exception` license to the project because of this.

adickin-amd: This is directly copied from the rock: https://github.com/ROCm/TheRock/blob/main/dockerfiles/cpubuilder/install_cmake.sh.  I can change how were getting cmake to not use the script if we think this will cause a problem.

averinevg: I'm not a licensing expert, but I looked into this and got some advice.
To comply with a license requirements of the third-party scripts which are under Apache 2.0, we should keep the original license headers in those scripts, and add a `NOTICE` file to the project root that lists the files under Apache 2.0 and includes the necessary attribution. There is no need to modify our main `LICENSE` file, as long as the Apache-licensed files are properly attributed in the `NOTICE` file and retain their original license headers. This approach should keep us compliant with both MIT and Apache 2.0 licenses.

Example `NOTICE` file content:
```
This project includes third-party code licensed under the Apache License 2.0 with LLVM Exceptions:

- dockerfiles/scripts/install_cmake.sh
  Originally from https://github.com/ROCm/TheRock/blob/main/dockerfiles/cpubuilder/install_cmake.sh
  Copyright 2022 The IREE Authors
  Licensed under the Apache License v2.0 with LLVM Exceptions.

...
```

averinevg: I found the original file at https://github.com/iree-org/base-docker-images/blob/main/build_tools/install_cmake.sh

BrianHarrisonAMD: Oh we should add this for flatbuffers usage as well.

adickin-amd: ive added a notice for the 3rd party files I added in this PR.  I am making a jira ticket to figure out how to properly attribute to any 3rd party dependencies

CMakeLists.txt
adickin-amd: these need to be off for now as CI cannot support them yet.

plugins/miopen_legacy_plugin/README.md
```diff
@@ -0,0 +1,18 @@
+# MIOpen Legacy Plugin
+A plugin wrapping MIOpen in order to provide engines to solve some hipDNN graphs.
+
+## Building
+This plugin can be built as part of the MIOpen project or as a standalone plugin.
+
+### Building as part of hipDNN
+1. Follow the build instructions for hipDNN defined [Building hipDNN](../../docs/Building.md).
+2. Currently the plugin build is defaulted to on.  Eventually it will be always built separately.
```
averinevg: HIP_DNN_BUILD_PLUGINS is defaulted to OFF

plugins/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+# Copyright © Advanced Micro Devices, Inc., or its affiliates.
+# SPDX-License-Identifier:  MIT
+
+# Extra flag to be used for the plugins when building them as part of the main hipdnn build
+option(BUILD_PLUGIN_AS_DEPENDENCY "Flag to build the plugins as part of the main hipdnn build" ON)
```
averinevg, line 5: The flag should probably be just a variable, not an option.

NOTICE.md
averinevg, line 1: The file should not have an extension.

plugins/miopen_legacy_plugin/README.md
adickin-amd: ah will update.  I need it off for CI until we can get support there

plugins/CMakeLists.txt
adickin-amd: I think an option is the right choice since its either ON or OFF.  If we do a variable we need to document on/off.  Eventually we would want this gone as we will build the plugins elsewhere.

NOTICE.md
adickin-amd: fixed

sdk/include/hipdnn_sdk/plugin/plugin_api_data_types.h
```diff
@@ -10,7 +10,11 @@
  * This file contains the definitions of the data types used in the hipDNN Plugin API.
  */
 
+#include <stddef.h>
+#include <stdint.h>
```
averinevg, line 14: These header files are not needed here (at least for now).

adickin-amd: if I dont have these, i get undefined errors on size_t and int64_t

adickin-amd: I can add the includes where I use this but I think its better to have the header include what it needs.

averinevg: A header file should include only what is used in that file.

averinevg: Although in this case, it might be better to leave it as is. Resolved.

averinevg (approved):

BrianHarrisonAMD (approved):

PR #54, opened 2025-05-15
Use gmocks for backend descriptor mocks

JonathanLichtnerAMD:
## Proposed changes

This change converts the existing custom backend descriptor mocks to use gmocks.

## Checklist


BrianHarrisonAMD (approved):
Looks good to me, but you will need to fix the CI issues.
I think there is some warnings that need to be disabled as a result of using mocks.

JonathanLichtnerAMD:
> Looks good to me, but you will need to fix the CI issues. I think there is some warnings that need to be disabled as a result of using mocks.

Yes, fixed.  This warning was introduced by running the auto-formating, which moved the // NOLINT comment that I had to a different line.  (It doesn't like the naming of SetArg4ToInt64, but that matches the gmock style.)

adickin-amd (approved):
LGTM

PR #55, opened 2025-05-16
Engine plugin C-API

averinevg:
## Proposed changes

This PR introduces the following changes:

1. C-API for Engine Plugin.
Added new functions to support engine plugin operations:

- `hipdnnPluginGetLastErrorString`, Retrieves the last error string from the plugin.
- `hipdnnEnginePluginCreate`, Creates a new engine plugin handle.
- `hipdnnEnginePluginDestroy`, Destroys an engine plugin handle and releases associated resources.
- `hipdnnEnginePluginSetStream`, Sets the HIP stream for the specified engine plugin handle.
- `hipdnnEnginePluginGetApplicableEngineIds`, Retrieves the IDs of the applicable engines for a given operation graph.
- `hipdnnEnginePluginGetEngineDetails`, Retrieves the details of a specific engine using its ID and the operation graph.
- `hipdnnEnginePluginDestroyEngineDetails`, Destroys the `engine_details` object and releases the associated resources.
- `hipdnnEnginePluginGetWorkspaceSize`, Retrieves the required workspace size for a specific engine configuration and an operation graph.
- `hipdnnEnginePluginCreateExecutionContext`, Creates an execution context for a specific engine configuration and an operation graph.
- `hipdnnEnginePluginDestroyExecutionContext`, Destroys an execution context and releases the associated resources.
- `hipdnnEnginePluginExecuteOpGraph`, Executes an operation graph using a specified execution context.

2. Initial Data Schemas.
Introduced FlatBuffers schemas for the following objects:

- `EngineDetails`
- `EngineConfig`

3. Generated Files
Generated FlatBuffers code for the above schemas to facilitate serialization and deserialization.

## Checklist


backend/src/plugin/engine_plugin.cpp
```diff
@@ -56,10 +57,11 @@ void Engine_plugin::run_engine(unsigned engine_index,
 {
     assert(_initialized);
     auto status = _func_run_engine(engine_index, input, output, size);
-    if(status != hipdnnPluginStatusSuccess)
+    if(status != HIPDNN_PLUGIN_STATUS_SUCCESS)
     {
         throw Hipdnn_exception(HIPDNN_STATUS_PLUGIN_ERROR,
-                               "Failed to run the engine. Status code: " + std::to_string(status));
+                               "Failed to run the engine. Status code: " + std::to_string(status)
```
adickin-amd, line 63: would it make sense to add functions to convert the enum type to its string name?  Does std::to_string(status) convert the enum to the string name?

backend/src/plugin/plugin_core.hpp
```diff
@@ -36,6 +36,9 @@ class Plugin_base
     hipdnnPluginType_t type() const;
 
 protected:
+    // This function must not throw as it is used during error handling.
```
adickin-amd, line 39: I think the noexcept keyword is enough to imply this shouldnt throw.

backend/src/plugin/engine_plugin.cpp
averinevg: > would it make sense to add functions to convert the enum type to its string name?

I think this should be in the SDK. But I don't think we need it right now, since the status code is mostly used just to indicate whether it's OK or not. We can do without it for now.

> Does std::to_string(status) convert the enum to the string name?

`enum` is converted to a string as a number.

backend/src/plugin/plugin_core.hpp
averinevg: I'd keep this comment, since it explains why this is implemented this way. `noexcept` by itself doesn't guarantee or describe anything.

adickin-amd (approved):

PR #56, opened 2025-05-20
Unified logging to 1 output

SamuelReeder:
## Proposed changes
- All logs originating from the frontend, plugins, or samples, initialize a callback sink that formats and sinks their logs to a backend API call, where the message is logged to the destination sinks.
- The backend uses it's own lazy-initialized internal logger that outputs logs to the same file because most of the backend cannot access the high-level callback function because there would be circular dependencies.
- Upon the first log being sent through a callback or internal logger, the backend will initialize and configure the base logging system by itself. The backend and callback receiving logger are separate (both writing to same file), because they need different patterns.
- `HIPDNN_LOG_LEVEL` defines the logging level. Defaults to `off`.
- `HIPDNN_LOG_FILE` defines the name of the output file. Logs append to this file.
- I have added two testing suites: one for the internal backend logging and the other for callback logging.
- Added a `target_compile_definition` to some cmake files to define the `COMPONENT_NAME` for including the component in logs.
- Plugins now have a mechanism to store a callback function given by the plugin manager.

Example (running sample):
```bash
[2025-07-16 17:07:14.224] [tid 72909] [info] [hipdnn_frontend] Frontend logging initialized via callback.
[2025-07-16 17:07:14.226] [tid 72909] [info] [hipdnn_backend] API called: [hipdnnCreate] handle_ptr=0x7ffed449f3b0
[2025-07-16 17:07:14.226] [tid 72909] [info] [hipdnn_backend] Created handle: 0x1ee4a00
[2025-07-16 17:07:14.226] [tid 72909] [info] [hipdnn_backend] API success: [hipdnnCreate] created_handle=0x1ee4a00
[2025-07-16 17:07:14.226] [tid 72909] [info] [hipdnn_backend] API called: [hipdnnDestroy] handle=0x1ee4a00
[2025-07-16 17:07:14.226] [tid 72909] [info] [hipdnn_backend] API success: [hipdnnDestroy] 
```

## Checklist


README.md
```diff
@@ -30,7 +30,7 @@ There are two environment variables that control the logging in hipDNN.
 
 If the log level is not set, then the default log level is off.
 
-`HIPDNN_LOG_DIR`: this controls the directory where log files will be written.  If this is not set, log files will be written to stdout.  The backend log file will have the name `hipdnn_timestamp.log` when written to the directory.
+`HIPDNN_LOG_FILE`: this controls the path to the log file that will be written.  If this is not set, log files will be written to the current directory with the name `hipdnn_timestamp.log`.
```
adickin-amd: if someone specifies the log file, what is the behaviour expected on subsequent runs of the software?  Do we just overwrite the entire logfile?  Should we keep around a certain number of logfiles, for example, we could keep around the last 5 logs and have logic to append .1 through .5 to each log.

adickin-amd: maybe we just need to dig into what miopen does for this.

SamuelReeder: Currently it truncates the file, but that is probably better since the user would have more control.

backend/include/hipdnn_backend.h
```diff
@@ -283,6 +284,23 @@ HIPDNN_BACKEND_EXPORT void hipdnnGetLastErrorString(char* message, size_t max_si
...
+ * @brief Callback function for logging messages.
+ *
+ * This function is called by the hipDNN library to log messages. The severity level of the message
+ * is provided, along with the message itself and user-defined data.
+ * 
+ * @param [in] severity  The severity level of the message.
+ * @param [in] msg        The message to be logged.
+ * @param [in] user_data   User-defined data passed to the callback function.
+ * 
+ * @retval HIPDNN_STATUS_SUCCESS           The logging callback was successfully executed
```
adickin-amd: can probably remove the return values, your function is a void

backend/src/hipdnn_backend.cpp
```diff
@@ -273,3 +273,38 @@ HIPDNN_BACKEND_EXPORT void hipdnnGetLastErrorString(char* message, size_t max_si
         LOG_API_SUCCESS(api_name, "set_error_message={:p}", static_cast<void*>(message));
     });
 }
+
+HIPDNN_BACKEND_EXPORT void
+    hipdnnLoggingCallback_ext(hipdnnSeverity_t severity, const char* msg, void* user_data)
+{
+    (void)user_data;
```
adickin-amd: if we dont need the user_data, then I think its okay to remove it from the callback.

backend/src/plugin/plugin_core.cpp
```diff
@@ -87,5 +99,17 @@ std::string_view Plugin_base::get_last_error_string() const noexcept
     return error_str;
 }
 
+hipdnnPluginStatus_t Plugin_base::set_logging_callback(hipdnnCallback_t callback)
```
averinevg, line 86: I think that a plugin manager should set `hipdnnLoggingCallback_ext` for all plugins during initialization. This can be changed later.

README.md
```diff
@@ -27,7 +27,7 @@ There are two environment variables that control the logging in hipDNN.
...
```
averinevg: It might be worth adding an option to disable file logging.

README.md
averinevg: We can create a new file each time and include the current date and time in the file name, so that filenames are both unique and human-readable. When creating a new file, we should also check that total number of files does not exceed 5 (or whatever limit is). Additionally, we need to monitor the file size and create a new file when the size limit is exceeded.

README.md
SamuelReeder: Yes, I intend to remove the default file-logging functionality and log to stderr or stdout. As in: if user specifies file, then append logs to file or creates new file with index suffix, otherwise, log to stderr (for example).

backend/CMakeLists.txt
SamuelReeder, line 1: @mousdahl-amd's installation work will take precedence over this.

backend/src/CMakeLists.txt
```diff
@@ -31,7 +31,8 @@ target_include_directories(hipdnn_backend_private
 )
 
 target_compile_options(hipdnn_backend_private PRIVATE ${HIPDNN_WARNING_COMPILE_OPTIONS})
-target_link_libraries(hipdnn_backend_private PUBLIC hip::host hipdnn_sdk)
+target_link_libraries(hipdnn_backend_private PUBLIC hip::device hipdnn_sdk)
```
SamuelReeder: Backend should not link `hip::device`, but it's oddly needed for the sample currently. I'm hoping to see the upcoming installation changes before I spend time fixing this. In sample I link the frontend which transitively includes the backend because its `INTERFACE` linked to frontend, and I then use the callback from backend. I don't think the transitive behavior is correct either, however.

README.md
SamuelReeder: Decided to append logs to file for now. There is no handling for when a file gets too large, but [`rotating_file_sink`](https://github.com/gabime/spdlog/wiki/Sinks) looks promising if we want that.

BrianHarrisonAMD: Seems fine for now.

backend/CMakeLists.txt
```diff
@@ -4,4 +4,3 @@
 set(HIP_DNN_BACKEND_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include")
 
 add_subdirectory(src)
-add_subdirectory(tests)
```
BrianHarrisonAMD: Why was this removed?

backend/src/plugin/plugin_core.cpp
```diff
@@ -31,6 +31,18 @@ void Plugin_base::resolve_symbols()
...
+    try
+    {
+        const auto func_name_set_logging_callback = "hipdnnPluginSetLoggingCallback";
+        _func_set_logging_callback
+            = _lib.get_symbol<decltype(_func_set_logging_callback)>(func_name_set_logging_callback);
+    }
+    catch(const Hipdnn_exception&)
+    {
+        _func_set_logging_callback = nullptr;
+    }
```
BrianHarrisonAMD: Probably should log a warning, or info message if the plugin being loaded doesn't support it.

backend/tests/test_plugin.cpp
```diff
@@ -143,3 +151,19 @@ TEST(PluginManagerTest, LastErrorOnSecondLoad)
...
+TEST(PluginManagerTest, SetLoggingCallback)
+{
+    plugin::Plugin_manager_base<Plugin> plugin_manager;
+    std::vector<std::filesystem::path> plugin_paths = {"./hipdnn_test_plugin1"};
+    plugin_manager.load_plugins(plugin_paths);
+
+    // This is bad but necessary unless we overload `get_plugins` to return a non-const vector.
+    // The purpose of this test is only to check the setter functionality.
+    auto& plugins = const_cast<std::vector<Plugin>&>(plugin_manager.get_plugins());
+    ASSERT_EQ(plugins.size(), 1);
```
BrianHarrisonAMD: Could make this version private, and make the tester a friend to avoid this. 
A bit awkward to test this piece.
We would ideally want to stub this out, but the plugin loading is in the manager, and doesn't seem easy.

frontend/CMakeLists.txt
```diff
@@ -11,7 +11,11 @@ endif()
 add_subdirectory(tests)
 
 add_library(hipdnn_frontend INTERFACE)
-target_link_libraries(hipdnn_frontend INTERFACE hipdnn_sdk)
+
+# Since this is a header-only library, we set the INTERFACE property. However, this importantly assumes that logging will only be done by consumers of the frontend library.
+# If we intend to log from within the frontend library itself, one needs to manually set compile definition to initialize the logger.
+target_compile_definitions(hipdnn_frontend INTERFACE COMPONENT_NAME="hipdnn_frontend")
+target_link_libraries(hipdnn_frontend INTERFACE hipdnn_sdk hipdnn_backend)
```
BrianHarrisonAMD: What does this mean?
We will likely have logging inside the frontend library headers, but I wouldn't expect consumers to add any of their own logging.

backend/CMakeLists.txt
SamuelReeder: Merge mistake, will add back.

samples/main.cpp
```diff
@@ -33,7 +33,11 @@ static bool not_equal(const Tensor_attributes& tensor,
 
 int main()
 {
-    hipdnn::logging::initialize_logger_to_std_out("hipdnn_sample");
+#ifdef COMPONENT_NAME
+    hipdnn::logging::initialize_callback_logging(COMPONENT_NAME, hipdnnLoggingCallback_ext);
+#endif
+
+    HIPDNN_LOG_INFO("HipDNN SDK Sample: Graph Serialization and Deserialization");
```
BrianHarrisonAMD: I am a bit surprised by this.
I would expect for this to exist in the frontend headers somewhere, and a consumer app doesn't need to really know about this. 

Maybe at most, some utility function that is called to initialize frontend logging.

frontend/CMakeLists.txt
SamuelReeder: The issue is that I can only set `INTERFACE` compile definitions for frontend. However, for logging to include component name, I need to find another way to set that.

samples/main.cpp
SamuelReeder: Yes you're right, I need to modify how its done in frontend (see last comment). I was unsure if samples would use our logging scheme at first, but after seeing how MIOpen does it, probably not.

frontend/CMakeLists.txt
BrianHarrisonAMD: We probably should have some frontend logging utility / method that set's it up, and just has the component name hardcoded. 

We wouldn't need this define then.

sdk/include/hipdnn_sdk/logging/logger.hpp
```diff
@@ -3,103 +3,97 @@
...
-
-    setup_log_pattern(logging_area_name);
-}
+#ifdef ENABLE_BACKEND_LOGGING // The backend configures all logging
+inline std::string output_file;
+inline bool g_logging_initialized = false;
+inline std::mutex g_logging_init_mutex;
+inline const std::string G_LOGGING_AREA = "hipdnn_backend";
+inline std::shared_ptr<spdlog::logger> g_backend_logger;
+inline std::shared_ptr<spdlog::logger> g_callback_receiver_logger;
```
BrianHarrisonAMD: Does this need to be in the SDK?
I think we could remove the define, and migrate this to a different backend only header.

sdk/tests/main.cpp
```diff
@@ -10,8 +10,8 @@ int main(int argc, char** argv)
 {
     ::testing::InitGoogleTest(&argc, argv);
 
-    hipdnn::logging::initialize_logger_to_std_out("sdk_tests");
-    hipdnn::logging::set_log_level("info");
+    // hipdnn::logging::initialize_logger_to_std_out("sdk_tests");
+    // hipdnn::logging::set_log_level("info");
```
BrianHarrisonAMD: Can be removed?

sdk/include/hipdnn_sdk/logging/logger.hpp
SamuelReeder: Are you saying all the `#ifdef ENABLE_BACKEND_LOGGING` masked functions should be moved too? I think that's necessary to maintain separation of backend and sdk.

Edit: actually I think I can move some variables if I refactor the functions.

frontend/CMakeLists.txt
SamuelReeder: I did this and defined `COMPONENT_NAME` in `hipdnn_frontend.hpp` so everything in frontend has the definition and can easily log.

backend/src/hipdnn_backend.cpp
```diff
@@ -11,6 +11,7 @@
 #include "helpers.hpp"
 #include "hipdnn_exception.hpp"
 #include "plugin/plugin_manager.hpp"
+#include <spdlog/spdlog.h>
```
SamuelReeder, line 16: Didn't mean to push this now, but will be kept or removed when I move some things to backend.

sdk/include/hipdnn_sdk/logging/logger.hpp
```diff
@@ -121,39 +115,115 @@ inline void set_log_level(const std::string& level)
...
+        {
+            hipdnn::logging::set_log_level("off");
+        }
+
+        g_logging_initialized = true;
+
+        return output_file;
     }
-    else
+    catch(const spdlog::spdlog_ex& ex)
```
SamuelReeder, line 60: Streamlining the error handling too.

sdk/include/hipdnn_sdk/logging/logger.hpp
SamuelReeder: I migrated all the functions and variables in the sdk that were only available to the backend, to the actual backend. I think this refactor is a big improvement. The only downside is that logging from the backend has to import a different file than everywhere else. Nonetheless, you call the same macros and everything else is the same.

backend/src/logging/logging.cpp
```diff
@@ -0,0 +1,132 @@
...
+// Could refactor this to a class with a single static instance.
+// The benefit would be a destructor to cleanup logging.
+std::mutex s_logging_init_mutex;
+bool s_logging_initialized = false;
+const std::string S_BACKEND_LOGGER_NAME = "hipdnn_backend";
+const std::string S_CALLBACK_RECEIVER_LOGGER_NAME = "hipdnn_callback_receiver";
+
+} // namespace
+
+void initialize()
```
SamuelReeder, line 33: My C programming side made me tempted to return int, but is void most appropriate? Could also return hipdnnStatus, but feels a bit weird here.

BrianHarrisonAMD: Probably void, and then we throw if it ever can fail since this is in the backend.
Right now it seems fine as is.

backend/src/logging/logging.cpp
```diff
@@ -0,0 +1,136 @@
...
+        // It doesn't need to return if log_level == off, but it avoids unnecessary initialization
+        if(log_level == nullptr || std::string(log_level) == "off")
+        {
+            s_logging_initialized = true;
+            return;
+        }
+
+        if(!spdlog::thread_pool())
+        {
+            spdlog::init_thread_pool(8192, 1);
```
adickin-amd, line 55: what is the default thread pool?  Curious as to why we do this?  When is the case where the thread_pool is false?

backend/src/logging/logging.cpp
```diff
@@ -0,0 +1,136 @@
...
+            return;
+        }
+
+        const char* log_level = std::getenv("HIPDNN_LOG_LEVEL");
+        const char* log_file_path = std::getenv("HIPDNN_LOG_FILE");
+
+        // It doesn't need to return if log_level == off, but it avoids unnecessary initialization
+        if(log_level == nullptr || std::string(log_level) == "off")
+        {
+            s_logging_initialized = true;
```
adickin-amd, line 49: I wonder if we want to default to the stderr_color_sink_mt so we are in control of the logger.  Might be something to consider once we look into the deallocation problem

backend/src/logging/logging.cpp
SamuelReeder: It's 8192 queue size and 1 worker thread, just as there, but apparently it needs to be init: https://github.com/gabime/spdlog/issues/952

backend/src/logging/logging.cpp
SamuelReeder: So do the initialization for stderr even if no logs will be printed?

mousdahl-amd (approved):
LGTM

plugins/miopen_legacy_plugin/tests/main.cpp
```diff
@@ -10,8 +10,5 @@ int main(int argc, char** argv)
 {
     ::testing::InitGoogleTest(&argc, argv);
 
-    hipdnn::logging::initialize_logger_to_std_out("miopen_legacy_plugin");
```
adickin-amd, line 14: Do tests still correctly output to stdout/stderr without these?

backend/src/logging/logging.cpp
adickin-amd: i may actually be wrong on that.  I misunderstood how you route to the loggers now.  Its no longer using the spdlog::info functions I was using before.

adickin-amd (approved):

plugins/miopen_legacy_plugin/tests/main.cpp
SamuelReeder: Oh yes should add that one back. The sdk tests one has to stay removed though unless using a custom callback, since it shouldn't use backend callback.

SamuelReeder: Later on could maybe make a default fallback logger, so we can still log if we can't link backend for callback.

adickin-amd: okay, lets capture that and add it back in later.  its super useful to be able to log to stdout from tests.  id do it in a follow up story.

SamuelReeder: The rest of the tests can log to stdout, just not sdk. But it might be possible if I only link the sdk tests folder against backend. I'll see.

adickin-amd: i dont think we should link the sdk tests or plugin tests to the backend.  we can work out another solution.  Its odd to have tests depend on the backend only for logging.  Lets just remove logging from these tests for now and revisit it later if we find we miss the logs.  

If you do want to get this working, I was thinking you could just define a custom logger function in the main of each test library and have it print to the console directly.  

```

#include <gtest/gtest.h>

 void myTestLogger(hipdnnSeverity_t severity, const char* msg)
{
   // I assume the callback logger is formatting msg so we dont need to do much.
    std::cout << msg; // might need a << "\n" as well.

// alternatively you could create another shared spdlog logger and store the shared object in this file
}

int main(int argc, char** argv)
{
    ::testing::InitGoogleTest(&argc, argv);

    hipdnn::logging::initialize_callback_logging("miopen_legacy_plugin", myTestLogger);

    return RUN_ALL_TESTS();
}

```

SamuelReeder: Yeah that solution isn't bad and we still get the formatting, it just takes a bit more work to use the env vars if desired. 

Logging in sdk will generally be challenging otherwise since we have backend of course linking against sdk, but then sdk might want the callback from the backend as well. But that seems to be the cost of logs going through backend while having the formatting that only the sdk can provide.

adickin-amd: I think its okay if the tests dont respect the env variables at this time.

SamuelReeder: Removed backend linking and added a testing utility callback to use in any cases where the backend callback is currently unobtainable. In practice, the plugins will use the true callback since it is passed by plugin manager, but we do not have tests to load the legacy plugin via plugin manager yet.

adickin-amd (approved):

PR #57, opened 2025-06-25
Backed api execute plans

BrianHarrisonAMD:
## Proposed changes

- Added get_engine_config method in execution_plan_descriptor.cpp
- Added definition of hipdnnBackendExecute(hipdnnHandle_t handle,hipdnnBackendDescriptor_t execution_plan,hipdnnBackendDescriptor_t variant_pack)
- Wrote Plugin_manager::execute(Execution_plan_descriptor* execution_plan_desc, hipdnnHandle* handle, Variant_descriptor* variant_desc)
- Test added to create batch norm graph, link the tensor of the graph to variant pack and call execute method

## Checklist


CMakeLists.txt
```diff
@@ -98,6 +98,9 @@ list(PREPEND HIPDNN_WARNING_COMPILE_OPTIONS
     -Winit-self                              # Warn about variables initialized with itself
     -Wunreachable-code                       # Warn about unreachable code
     -Wno-error=unused-command-line-argument  # Disable error for unused command line arguments
+    -Wimplicit-fallthrough                   # Warn about fallthrough cases in switch statements
+    #-Wswitch-enum                            # Warn if a switch statement does not cover all enumerators
+    #-Wswitch-default                         # Warn if a switch statement does not have a default case
```
BrianHarrisonAMD: I think we should add the bottom two that are commented out, but the fall through is already set by -Wextra I believe.

backend/src/descriptors/execution_plan_descriptor.cpp
```diff
@@ -176,4 +179,39 @@ void Execution_plan_descriptor::set_engine_config(hipdnnBackendAttributeType_t a
...
+
+    if(element_count != nullptr)
+    {
+        *element_count = requested_element_count;
+    }
+
+    if(array_of_elements == nullptr)
+    {
+        return;
+    }
```
BrianHarrisonAMD: Do we need this check?

Should be a throw if it's not already checked before this point.

backend/src/plugin/plugin_manager.cpp
```diff
@@ -3,19 +3,36 @@
...
-    /* some stuff to help you find which plugins should be loaded, but for now blank */)
+void Plugin_manager::initialize(/* heuristics */)
 {
+
     // TODO : actually find and init the plugins properly
     // for now we will just use a fake plugin.
-    // # DISCUSS: how do we find the plugins?
-    _plugins.push_back(std::make_shared<Fake_plugin>());
+    // # DISCUSS: how do we find the plugins? store in json?
+    auto fake_plugin = std::make_shared<Fake_plugin>();
```
BrianHarrisonAMD: Plugin manager should have a directory that it loads plugins from during init, and possibly an env that allows you to add additional plugin load paths / dirs to search.

backend/src/plugin/plugin_manager.cpp
```diff
@@ -3,19 +3,36 @@
...
+
+std::shared_ptr<Hipdnn_plugin_base> Plugin_manager::get_plugin(int64_t engine_id)
+{
+    auto plugin_iter = _plugins.find(engine_id);
+    if(plugin_iter == _plugins.end())
+    {
+        return nullptr;
+    }
+    return plugin_iter->second;
 }
```
BrianHarrisonAMD: Probably fine for now, but I think @averinevg was going to make it so all the API for interacting with plugins happens at the plugin manager level. 

I suppose we can clean this up later, and add API forwarding to the plugin manager.

backend/src/plugin/plugin_manager.cpp
```diff
@@ -71,4 +92,85 @@ std::set<int64_t>
...
+                   HIPDNN_STATUS_BAD_PARAM,
+                   "Plugin_manager::execute failed: execution_plan_desc is not finalized");
+
+    hipdnnBackendDescriptor_t engine_config = nullptr;
+
+    plan_desc->get_attribute(HIPDNN_ATTR_EXECUTION_PLAN_ENGINE_CONFIG,
+                             HIPDNN_TYPE_BACKEND_DESCRIPTOR,
+                             1,
+                             nullptr,
+                             &engine_config);
```
BrianHarrisonAMD, line 129: Can probably just add a getter to directly get this.
Don't think we need to use the get apis internally as it involves a bunch of extra checks and casts.

BrianHarrisonAMD: Ill leave it for now since we do this in quite a few places, and it's just a style thing.

backend/src/plugin/plugin_manager.cpp
```diff
@@ -71,4 +92,85 @@ std::set<int64_t>
...
+    engine_config_desc->get_attribute(
+        HIPDNN_ATTR_ENGINECFG_ENGINE, HIPDNN_TYPE_BACKEND_DESCRIPTOR, 1, nullptr, &engine);
+
+    THROW_IF_NULL(
+        engine, HIPDNN_STATUS_BAD_PARAM, "Plugin_manager::execute failed: engine is null");
+
+    THROW_IF_NE(engine->type,
+                HIPDNN_BACKEND_ENGINE_DESCRIPTOR,
+                HIPDNN_STATUS_BAD_PARAM,
+                "Plugin_manager::execute failed: invalid engine descriptor type");
```
BrianHarrisonAMD: Do we need these checks?

Seems like expectation is this shouldn't be possible after getting it.

backend/src/plugin/plugin_manager.cpp
```diff
@@ -3,19 +3,36 @@
...
+        _plugins.insert({engine_id, fake_plugin});
+    }
+}
+
+std::shared_ptr<Hipdnn_plugin_base> Plugin_manager::get_plugin(int64_t engine_id)
+{
+    auto plugin_iter = _plugins.find(engine_id);
+    if(plugin_iter == _plugins.end())
+    {
+        return nullptr;
```
BrianHarrisonAMD: Should throw if it's not found.

backend/src/plugin/plugin_manager.cpp
```diff
@@ -71,4 +92,85 @@ std::set<int64_t>
...
+
+    int64_t engine_id = 0;
+    engine_desc->get_attribute(
+        HIPDNN_ATTR_ENGINE_GLOBAL_INDEX, HIPDNN_TYPE_INT64, 1, nullptr, &engine_id);
+
+    auto plugin = get_plugin(engine_id);
+    THROW_IF_NULL(plugin,
+                  HIPDNN_STATUS_BAD_PARAM_OUT_OF_BOUND,
+                  std::string("Plugin_manager::execute has invalid engine id: ")
+                      + std::to_string(engine_id));
```
BrianHarrisonAMD: Going to move the throw to get_plugin, and not check null.

backend/src/plugin/plugin_manager.cpp
```diff
@@ -71,4 +92,85 @@ std::set<int64_t>
...
+    hipdnnBackendDescriptor_t graph = nullptr;
+    engine_desc->get_attribute(
+        HIPDNN_ATTR_ENGINE_OPERATION_GRAPH, HIPDNN_TYPE_BACKEND_DESCRIPTOR, 1, nullptr, &graph);
+
+    THROW_IF_NULL(graph, HIPDNN_STATUS_BAD_PARAM, "Plugin_manager::execute failed: graph is null");
+
+    THROW_IF_NE(graph->type,
+                HIPDNN_BACKEND_OPERATIONGRAPH_DESCRIPTOR,
+                HIPDNN_STATUS_BAD_PARAM,
+                "Plugin_manager::execute failed: invalid graph descriptor type");
```
BrianHarrisonAMD: Shouldn't need these.

backend/src/plugin/plugin_manager.cpp
```diff
@@ -3,19 +3,36 @@
...
+
     // TODO : actually find and init the plugins properly
     // for now we will just use a fake plugin.
-    // # DISCUSS: how do we find the plugins?
-    _plugins.push_back(std::make_shared<Fake_plugin>());
+    // # DISCUSS: how do we find the plugins? store in json?
+    auto fake_plugin = std::make_shared<Fake_plugin>();
+
+    // for all the applicable engines, add the same plugin to the map
+    for(const int64_t& engine_id : fake_plugin->get_applicable_engines(nullptr))
```
BrianHarrisonAMD: We should have just a way to query all engines available for building this map.
Ill add one without a graph to get what engines this plugin has internally.

backend/src/descriptors/execution_plan_descriptor.cpp
```diff
@@ -176,4 +179,39 @@ void Execution_plan_descriptor::set_engine_config(hipdnnBackendAttributeType_t a
...
+
+    THROW_IF_LT(requested_element_count,
+                1,
+                HIPDNN_STATUS_BAD_PARAM,
+                "Execution_plan_descriptor failed to get engine config: Requested element count "
+                "too small.");
+
+    if(element_count != nullptr)
+    {
+        *element_count = requested_element_count;
```
SamuelReeder: Is this supposed to just echo back their requested count?

backend/src/plugin/plugin_manager.hpp
```diff
@@ -34,10 +34,14 @@ struct Plugin_manager
...
     // This will redirect the execute to the plugin that owns the engine selected inside the ExecutionPlan
     // Throws if invalid stuff is provided, and later is wrapped with a status + provides message
-    void execute(Execution_plan_descriptor* plan, hipdnnHandle* handle, Variant_descriptor* pack);
+    void execute(hipdnnHandle* handle,
+                 hipdnnBackendDescriptor_t execution_plan,
+                 hipdnnBackendDescriptor_t variant_pack);
 
 private:
-    std::vector<std::shared_ptr<Hipdnn_plugin_base>> _plugins;
+    std::map<int64_t, std::shared_ptr<Hipdnn_plugin_base>> _plugins;
```
BrianHarrisonAMD: Probably should have both, since we want a look-up from engineId -> plugin, and we also want to be able to query just the unique plugins.

backend/src/descriptors/execution_plan_descriptor.cpp
BrianHarrisonAMD: I think it's actually wrong. 
It should return 1, and we probably should throw if it's not requesting 1.

backend/tests/descriptors/test_execution_plan_descriptor.cpp
```diff
@@ -185,17 +186,65 @@ TEST_F(Execution_plan_descriptor_test, GetExecutionPlanDescriptorWorkspaceSize)
...
+    ASSERT_THROW_HIPDNN_STATUS(
+        _plan->get_attribute(
+            HIPDNN_ATTR_EXECUTION_PLAN_ENGINE_CONFIG, HIPDNN_TYPE_INT64, 1, &count, &dummy),
+        HIPDNN_STATUS_BAD_PARAM);
+
     ASSERT_THROW_HIPDNN_STATUS(_plan->get_attribute(HIPDNN_ATTR_EXECUTION_PLAN_ENGINE_CONFIG,
                                                     HIPDNN_TYPE_BACKEND_DESCRIPTOR,
-                                                    1,
-                                                    nullptr,
+                                                    0, // Too small (needs to be at least 1)
```
adickin-amd, line 230: uhhh, this is 0 but says it needs to be 1?

adickin-amd: oh nm, its a test

averinevg (approved):

PR #60, opened 2025-07-04
Added tensor() and tensor_like() to Graph

mousdahl-amd:
## Proposed changes
Added tensor() and tensor_like() methods to the Graph frontend api.  The cuDNN version did a lot of internal tracking for the purposes of uid collision checking and validation, which we don't appear to need, so it simplified them enough that they could become static methods.  I left them on Graph for api compatibility.

## Checklist

### DIFF
--- a/frontend/tests/test_graph.cpp
+++ b/frontend/tests/test_graph.cpp
@@ -863,7 +863,13 @@ TEST(GraphTests, TensorLikeGraphAttributes)
     EXPECT_EQ(tensor_like->get_name(), "TensorLike");
     EXPECT_NE(tensor_like->get_uid(), 100);
 
+    EXPECT_NE(tensor_like, tensor);
+
     auto tensor_like_noname = Graph::tensor_like(tensor_like);
     EXPECT_EQ(tensor_like_noname->get_name(), "");
+
+    EXPECT_EQ(tensor->get_name(), "TestTensor");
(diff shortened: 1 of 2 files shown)


frontend/tests/test_graph.cpp
```diff
@@ -823,4 +823,47 @@ TEST(GraphTests, BuildAndSerializePointwiseAndBatchnormBackwardGraph)
...
+
+    EXPECT_EQ(tensor_like->get_data_type(), DataType_t::FLOAT);
+    EXPECT_FALSE(tensor_like->get_is_virtual());
+    EXPECT_EQ(tensor_like->get_dim(), std::vector<int64_t>({1, 2, 3, 4}));
+    EXPECT_EQ(tensor_like->get_stride(), std::vector<int64_t>({5, 6, 7, 8}));
+    EXPECT_EQ(tensor_like->get_name(), "TensorLike");
+    EXPECT_NE(tensor_like->get_uid(), 100);
+
+    auto tensor_like_noname = Graph::tensor_like(tensor_like);
+    EXPECT_EQ(tensor_like_noname->get_name(), "");
```
BrianHarrisonAMD: Could add a check to ensure the original pointer != the cloned pointer, and that the original is unchanged after calling tensor_like.

BrianHarrisonAMD (approved):
Looks good to me!

SamuelReeder (approved):

PR #61, opened 2025-07-04
Implement the first 3 functions of the engine plugin api for miopen

adickin-amd:
## Proposed changes

- Add an initial structure for miopen solvers, only implemented the is_applicable check to start.  Some other functions such as get_workspace_size also exist but they currently do nothing for now. 
- When a handle is created, we also create a handle to miopen and save it on the handle.
- Add functionality I think can be shared across plugins to the sdk as header only classes.
- Added proper logging to the api entry functions to match the backend
- implement the last error string
- Add a way to track lifecycle for the new miopen container so it automatically deallocates if no handles are held for this plugin
- added tests where it makes sense. 

## Checklist


plugins/miopen_legacy_plugin/engines/solvers/miopen_batchnorm_solver.cpp
```diff
@@ -0,0 +1,22 @@
+/* Copyright © Advanced Micro Devices, Inc., or its affiliates. */
+/* SPDX-License-Identifier:  MIT */
+
+#include "miopen_batchnorm_solver.hpp"
+
+namespace miopen_legacy_plugin
+{
+
+bool Miopen_batchnorm_solver::is_applicable(const hipdnn_sdk::data_objects::GraphT& op_graph) const
```
adickin-amd, line 9: rather than make a "fake" solver I just stubbed in batchnorm as thats next.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -4,46 +4,298 @@
...
+                                                     std::string(typeid(T).name()) + " is nullptr");
+    }
+}
+
+// Keep a weak pointer to the Miopen_container thats made when we create a plugin handle.
+// The original shared_ptr is then stored on the handle so that it can be used for the lifecycle
+// of the handle.  If we create another handle, then we can use the weak pointer to get access
+// to the existing Miopen_container.  If all handles are destroyed, then this allows us to properly
+// clean up the container without having to fully unload the plugin.
+std::weak_ptr<Miopen_container> miopen_container_lifecycle_ptr;
```
adickin-amd, line 49: I have a debate with myself if this is the right thing to do.  I think we only need to instantiate the engine/solvers once and they can be shared across the handles, each handle would have its own miopen handle which would be used when interacting with miopen.  The alternative would be to create a static container which can only be cleaned up during shutdown.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -4,46 +4,298 @@
...
+hipdnnPluginStatus_t hipdnnEnginePluginCreate(hipdnnEnginePluginHandle_t* handle)
+{
+    LOG_API_ENTRY("handle_ptr={:p}", static_cast<void*>(handle));
+
+    return hipdnn_plugin::try_catch([&, api_name = __func__]() {
+        throw_if_null(handle);
+
+        miopen_legacy_plugin::Miopen_handle_factory::create_miopen_handle(handle);
+
+        if(auto miopen_container_ptr = miopen_container_lifecycle_ptr.lock())
```
adickin-amd, line 140: commenting for visibility as this is miopen container lifecycle stuff.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -4,46 +4,298 @@
...
 
 static const char* _plugin_name = "miopen_legacy_plugin";
 static const char* _plugin_version = "1.0.0";
 
-// Implementation of Plugin API
+using namespace hipdnn_plugin;
+using namespace miopen_legacy_plugin;
+
+// NOLINTNEXTLINE(modernize-avoid-c-arrays)
+thread_local char Plugin_last_error_manager::last_error[HIPDNN_MAX_ERROR_STRING_SIZE] = "";
```
adickin-amd, line 26: Note here: I have to define the last_error in my cpp as you cannot do it in the header only Plugin_last_error_manager class because it will cause duplicate symbol errors.  I debate if I should just make this class part of the plugin and not share its functionality in the sdk as its odd to have to do this.

sdk/include/hipdnn_sdk/plugin/plugin_last_error_manager.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include <hipdnn_sdk/utilities/string_util.hpp>
+
+#include <string>
+
+namespace hipdnn_plugin
+{
+// TODO replace with using the define from a common SDK header.
+static const size_t HIPDNN_MAX_ERROR_STRING_SIZE = 256;
+
+// TODO
```
adickin-amd, line 18: comment for visibility

sdk/tests/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+# Copyright © Advanced Micro Devices, Inc., or its affiliates.
+# SPDX-License-Identifier:  MIT
+
+if(HIP_DNN_SKIP_TESTS)
+    message(STATUS "Skipping MIOpen Legacy Plugin tests")
```
mousdahl-amd: Is this status correct?

plugins/miopen_legacy_plugin/engine_manager.cpp
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier:  MIT
+
+#include "engine_manager.hpp"
+#include "engines/miopen_engine.hpp"
+#include <algorithm>
+
+namespace miopen_legacy_plugin
+{
+
+Engine_manager::Engine_manager(std::set<std::unique_ptr<Engine>>& engines)
```
adickin-amd: This is a set passed as a reference in order to allow me to use dependency injection to pass engines.  This way i can test this class in isolation.  

I originally had a `add_engine` method which took in a shared_ptr to the engine and saved it to a set.  Id didnt like that way as it didnt leave us with 1 place that owned the engine in order to have simple construction/destruction logic.

adickin-amd: Another alternative could be to bring back the `add_engine` method but use a `std::reference_wrapper` to pass engines into this class.

adickin-amd: closing, i removed this.  ownership of engines gets passed to the manger and ownership of solvers gets passed to the engines.  This was way to weird.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -4,46 +4,298 @@
...
+void hipdnnPluginGetLastErrorString(const char** error_str)
+{
+    LOG_API_ENTRY("error_str_ptr={:p}", static_cast<void*>(error_str));
+
+    hipdnn_plugin::try_catch([&, api_name = __func__]() {
+        throw_if_null(error_str);
+
+        *error_str = Plugin_last_error_manager::get_last_error();
+
+        LOG_API_SUCCESS(api_name, "set_error_string={:p}", static_cast<void*>(error_str));
```
mousdahl-amd: `get_error_string` if you want it to match the others

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -4,46 +4,298 @@
...
 {
-    if(!type)
+    LOG_API_ENTRY("type_ptr={:p}", static_cast<void*>(type));
+
+    return hipdnn_plugin::try_catch([&, api_name = __func__]() {
+        throw_if_null(type);
+
+        *type = HIPDNN_PLUGIN_TYPE_ENGINE;
+
+        LOG_API_SUCCESS(api_name, "type={}", *type);
```
mousdahl-amd, line 88: `get_type` to match the other methods?

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
```diff
@@ -0,0 +1,49 @@
...
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
+size_t Miopen_engine::get_workspace_size() const
+{
+    return 1337;
```
mousdahl-amd, line 41: 😎

plugins/miopen_legacy_plugin/engine_manager.hpp
mousdahl-amd, line 1: Do you think some of these concepts like "Container, Engine, Engine_manager and Solver" might make sense in a, like plugin utils shared lib of some kind?  Or do we just re-implement for every plugin?

plugins/miopen_legacy_plugin/tests/test_engine_manager.cpp
```diff
@@ -0,0 +1,83 @@
...
+using namespace miopen_legacy_plugin;
+using ::testing::Return;
+
+TEST(Engine_managerTest, ReturnsApplicableEngineIds)
+{
+    std::set<std::unique_ptr<Engine>> engines;
+
+    auto mock_engine1 = std::make_unique<Mock_engine>();
+    EXPECT_CALL(*mock_engine1, id()).WillRepeatedly(Return(1));
+    EXPECT_CALL(*mock_engine1, is_applicable(::testing::_)).WillRepeatedly(Return(true));
```
mousdahl-amd, line 21: I've never played with gmock before.  This is neat!

plugins/miopen_legacy_plugin/engine_manager.hpp
adickin-amd: very possible we could refactor some of these concepts into the shared sdk once we finish this plugin.

sdk/tests/CMakeLists.txt
adickin-amd: nope, copy/paste error, good catch.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: hmm, i mirrored the way we did these comments in hipdnn_backend.  Maybe they didnt make sense.  Im going to modify these to just be `error_str={:p}`

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: leaving as type, modified others to no longer have get_ in the name.

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
SamuelReeder: What is the significance of this number? 🤔

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: todo, add locking

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
SamuelReeder: Ah, leet 😁

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: done

plugins/miopen_legacy_plugin/engines/engine.hpp
```diff
@@ -0,0 +1,25 @@
...
+class Engine
+{
+public:
+    virtual ~Engine() = default;
+
+    virtual int64_t id() const = 0;
+
+    virtual bool is_applicable(const hipdnnPluginConstData_t* op_graph) const = 0;
+
+    virtual size_t get_workspace_size() const = 0;
```
averinevg, line 22: The workspace size depends on the op graph

adickin-amd: absolutely, Ive stubbed things in and have been adding the arguments as I need them.  My changes behind this do have the op_graph in the list.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
averinevg: I believe we can also have cpp files in the sdk and convert it into a static library.

plugins/miopen_legacy_plugin/engine_manager.cpp
```diff
@@ -0,0 +1,31 @@
...
+{
+
+Engine_manager::Engine_manager() {}
+
+void Engine_manager::add_engine(std::unique_ptr<Engine> engine)
+{
+    _engines.insert(std::move(engine));
+}
+
+std::set<int64_t> Engine_manager::get_applicable_engine_ids(const hipdnnPluginConstData_t* op_graph)
```
averinevg, line 18: Is `std::set` necessary here?

plugins/miopen_legacy_plugin/engine_manager.hpp
```diff
@@ -0,0 +1,35 @@
...
+    //disallow copy and assignment
+    Engine_manager(const Engine_manager&) = delete;
+    Engine_manager& operator=(const Engine_manager&) = delete;
+
+    void add_engine(std::unique_ptr<Engine> engine);
+
+    std::set<int64_t> get_applicable_engine_ids(const hipdnnPluginConstData_t* op_graph);
+
+private:
+    std::set<std::unique_ptr<Engine>> _engines;
```
averinevg, line 32: What's the purpose of using `std::set` here?

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
```diff
@@ -0,0 +1,49 @@
...
+int64_t Miopen_engine::id() const
+{
+    return _id;
+}
+
+bool Miopen_engine::is_applicable(const hipdnnPluginConstData_t* op_graph) const
+{
+
+    std::unique_ptr<hipdnn_sdk::data_objects::GraphT> graph;
+    hipdnn_plugin::flatbuffer_utilities::convert_serialized_plugin_graph_to_graph(
```
averinevg, line 26: There is no need to create a new object here. You can use either `flatbuffers::GetRoot<hipdnn_sdk::data_objects::Graph>()` or just `GetGraph()` to get a pointer to the graph.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: If we do that, then we need to ship the static library as part of hipdnn which is not ideal.

plugins/miopen_legacy_plugin/engine_manager.hpp
BrianHarrisonAMD: Yea I think this could be vector, or a map from id -> engine ptr if we need a look-up eventually.

adickin-amd: yea, its a unordered_map in future PRs

adickin-amd: `std::unordered_map<int64_t, std::unique_ptr<Engine>> _engines;`

plugins/miopen_legacy_plugin/engine_manager.cpp
adickin-amd: I can make it a vector if that's preferred.

plugins/miopen_legacy_plugin/engines/solvers/solver.hpp
```diff
@@ -0,0 +1,23 @@
...
+#pragma once
+
+#include <stdint.h>
+
+#include <hipdnn_sdk/data_objects/graph_generated.h>
+
+namespace miopen_legacy_plugin
+{
+
+class Solver
```
BrianHarrisonAMD, line 13: I don't think it really matters since it's in the plugin, but I typically would call this SolverBase, or SolverInterface to make it a bit more clear.

BrianHarrisonAMD: Same applies to the engine interface.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
BrianHarrisonAMD: I think it's better to not have it in the SDK, and each plugin is responsible for handling this behavior.
It's simple enough to copy this if you want the same behavior, and it's not too much code IIRC.

BrianHarrisonAMD: Alternatively, I had a header only version of this that used locks, and a class variable for storing the state.
Doing it that way wouldn't require static init. 
We could add that to the SDK and people could use it.
last_error shouldn't be frequently set (I hope), and the use of a lock shouldn't be a perf concern.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -4,46 +4,311 @@
...
+        *num_engines = 0;
+        for(auto& engine_id : applicable_engines)
+        {
+            if(*num_engines == max_engines)
+            {
+                *num_engines = applicable_engines.size();
+                HIPDNN_LOG_WARN("Maximum number of engines reached ({}), ignoring additional "
+                                "engines, num_engines count: {}",
+                                max_engines,
+                                *num_engines);
```
BrianHarrisonAMD: Might be better to treat this as info instead of warning.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -4,46 +4,311 @@
...
+                                                        hipdnnPluginConstData_t* engine_details)
 {
-    if(!error_str)
-        return;
-    *error_str = "No error";
+    if(!handle || !op_graph || !engine_details)
+        return HIPDNN_PLUGIN_STATUS_BAD_PARAM;
+
+    return HIPDNN_PLUGIN_INTERNAL_ERROR;
+}
```
BrianHarrisonAMD: I guess this doesn't exist yet, but we would eventually want to populate this even if we had no details I imagine.

BrianHarrisonAMD: Oh no it does exist nevermind.
Yea we should be making a EngineDetails object here, and just set the Id. 

Probably fine it we do that in the follow-up PR though.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -4,46 +4,311 @@
...
+
+hipdnnPluginStatus_t hipdnnEnginePluginDestroyEngineDetails(hipdnnEnginePluginHandle_t handle,
+                                                            hipdnnPluginConstData_t* engine_details)
+{
+    if(!handle || !engine_details)
+        return HIPDNN_PLUGIN_STATUS_BAD_PARAM;
+
+    return HIPDNN_PLUGIN_INTERNAL_ERROR;
+}
```
BrianHarrisonAMD: And we should destroy it here as well I suppose, but again can be in the follow up.

plugins/miopen_legacy_plugin/tests/test_miopen_legacy_engine_plugin_api.cpp
```diff
@@ -0,0 +1,179 @@
...
+{
+    EXPECT_EQ(hipdnnEnginePluginGetEngineDetails(nullptr, 0, nullptr, nullptr),
+              HIPDNN_PLUGIN_STATUS_BAD_PARAM);
+}
+
+TEST(MiopenLegacyEnginePluginApiTest, EnginePluginDestroyEngineDetailsNull)
+{
+    EXPECT_EQ(hipdnnEnginePluginDestroyEngineDetails(nullptr, nullptr),
+              HIPDNN_PLUGIN_STATUS_BAD_PARAM);
+}
```
BrianHarrisonAMD, line 155: Should add a todo for creating and destroying these when they are implemented.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: ah its in my next set of changes.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: in the next set of changes

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
adickin-amd: I will migrate to using the ptr here and for workspace calcs.

plugins/miopen_legacy_plugin/engines/solvers/solver.hpp
adickin-amd: yea I usually do I_Solver but I think we said we didnt like that.  I can change to SolverInterface

plugins/miopen_legacy_plugin/tests/test_miopen_legacy_engine_plugin_api.cpp
adickin-amd: the tests are implemented in the next round of changes

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: ill just remove it from the plugin sdk

sdk/include/hipdnn_sdk/plugin/plugin_last_error_manager.hpp
BrianHarrisonAMD: Eh might be fine to just leave this, and add instantiation of this in the example / quick start plugin we create eventually. 
Can just have it initialize this state in the plugin.cpp, and add a comment / documentation to cover it.

sdk/tests/main.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+int main(int argc, char** argv)
+{
+    ::testing::InitGoogleTest(&argc, argv);
+
+    hipdnn::logging::initialize_logger_to_std_out("backend_tests");
+    hipdnn::logging::set_log_level("info");
+
+    return RUN_ALL_TESTS();
+}
```
BrianHarrisonAMD: Do we have main for SDK & Backend tests just for logger init?

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
BrianHarrisonAMD: LGTM for now.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
BrianHarrisonAMD: Hehe I changed my mind after seeing it. 
I think it's fine either way though.

plugins/miopen_legacy_plugin/engines/solvers/solver.hpp
BrianHarrisonAMD: Or w/e our style is. 
might be Solver_interface?

BrianHarrisonAMD (approved):
LGTM!

I think the other comments can be fixed in the follow up, or just left as is since they are minor.

sdk/tests/main.cpp
adickin-amd: yea, the alternative is just to include the gtest_main as a cmake dep.  I kinda like initing the logger to info so you can see everything in the tests as you run them.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: keeping or we have to also get rid of some of the helper try/catch functions.

sdk/include/hipdnn_sdk/plugin/plugin_last_error_manager.hpp
adickin-amd: added a better note to cover this for now.

plugins/miopen_legacy_plugin/engine_manager.cpp
adickin-amd: will make vector in next PR.

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
adickin-amd: will do as part of next PR as to not hit larger merge conflicts

plugins/miopen_legacy_plugin/engines/solvers/solver.hpp
adickin-amd: Will do in next PR to reduce merge conflicts.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: made it info

PR #62, opened 2025-07-07
Engine plugin C-API part 2

averinevg:
## Proposed changes

This PR utilizes the APIs introduced in https://github.com/ROCm/hipDNN/pull/55. It updates the corresponding logic to integrate with the newly added APIs.

## Checklist


backend/src/plugin/engine_plugin.cpp
```diff
@@ -23,44 +24,223 @@ void Engine_plugin::resolve_symbols()
...
+
+hipdnnEnginePluginExecutionContext_t
+    Engine_plugin::create_execution_context(hipdnnEnginePluginHandle_t handle,
+                                            const hipdnnPluginConstData_t* engine_config,
+                                            const hipdnnPluginConstData_t* op_graph) const
+{
+    assert(_initialized);
+    hipdnnEnginePluginExecutionContext_t exec_context;
+    auto status = _func_create_execution_context(handle, engine_config, op_graph, &exec_context);
+    if(status != HIPDNN_PLUGIN_STATUS_SUCCESS)
```
mousdahl-amd, line 207: Perhaps you could make a throw_on_fail macro or something to wrap all these calls?

backend/tests/test_engine_plugin.cpp
```diff
@@ -13,8 +14,38 @@
 
 using namespace hipdnn_backend;
 
+namespace
+{
+
+template <typename T, typename Destructor>
+class Scoped_resource
```
adickin-amd, line 21: nice way to deal with cleaning up hip ptrs.  Ill have to keep this in mind when we begin writing miopen integration tests. We may want to add something like this to the sdk as either a Util or test util.

backend/tests/plugins/engine_plugin_api_impl.cpp
```diff
@@ -0,0 +1,290 @@
...
+#include "engine_plugin_api_impl.hpp"
+#include "engine_plugin_handle.hpp"
+#include "plugin_api_impl.hpp"
+
+namespace
+{
+
+hipdnnPluginStatus_t check_handle_validity(hipdnnEnginePluginHandle_t handle)
+{
+    if(handle == nullptr)
```
mousdahl-amd, line 18: I wonder if it might be a useful pattern for plugins to keep a list of the handles they've handed out and verify that the handles that are passed in are ones that "belong" to them?  That or some sort of guardian bytes at the start of the structure to verify that it isn't some other plugin's handle.

mousdahl-amd (commented):
No show-stoppers from me!

adickin-amd (approved):
I think the updates look good.  Glad to see were using the hip::host instead of device now as well.

backend/tests/test_engine_plugin.cpp
averinevg: I'll add this to the sdk in the next PR.

backend/src/plugin/engine_plugin.cpp
averinevg: Yea, I'll change this in my next PR.

backend/tests/plugins/engine_plugin_api_impl.cpp
averinevg: Initially, that's exactly what I did - added a magic number to the handle, but later decided it was overkill.

PR #64, opened 2025-07-11
First cut of batchnorm inference working in MIOpen

adickin-amd:
## Proposed changes

- added all logic to enable us to do a batchnorm fwd inference via miopen
- Added initial GraphWrapper and EngineConfigWrapper to aid with how we interact with flatbuffer data structures.
- added wrapper for creating gpu memory so hipFree is called when the object is destructed.
- Refactored solvers into Plan_builders and Plans.  When we create an execution context that creates a Plan which is saved on the execution context.  This Plan is immutable and cannot change once created.  The execute graph C function now calls execute on the plan with device buffers and workspace.
- 

TODO in next PR as this one is quite large

- Add golden data verification for batch norm fwd inference.

## Checklist


plugins/miopen_legacy_plugin/engine_manager.cpp
```diff
@@ -1,31 +1,87 @@
 // Copyright © Advanced Micro Devices, Inc., or its affiliates.
 // SPDX-License-Identifier:  MIT
 
+#include <algorithm>
+
+#include <hipdnn_sdk/logging/logger.hpp>
```
SamuelReeder: I don't see logger being used in this file. Can remove.

plugins/miopen_legacy_plugin/engine_manager.cpp
```diff
@@ -1,31 +1,87 @@
...
+                                          int64_t engine_id,
+                                          const hipdnn_plugin::Graph_interface& op_graph) const
+{
+    auto it = _engines.find(engine_id);
+    if(it == _engines.end())
+    {
+        throw Hipdnn_plugin_exception(HIPDNN_PLUGIN_STATUS_INVALID_VALUE,
+                                      "Engine with ID " + std::to_string(engine_id)
+                                          + " not found.");
+    }
```
SamuelReeder, line 74: Maybe a helper function for finding engine and throwing would be good. Code is repeated at least twice, and another is only slightly different.

plugins/miopen_legacy_plugin/engines/solvers/miopen_batchnorm_solver.cpp
```diff
@@ -2,21 +2,187 @@
...
+                                                      estMeanDesc.tensor_desc,
+                                                      estVarianceDesc.tensor_desc,
+                                                      scaleDesc.device_buffer.ptr,
+                                                      biasDesc.device_buffer.ptr,
+                                                      estMeanDesc.device_buffer.ptr,
+                                                      estVarianceDesc.device_buffer.ptr,
+                                                      epsilon);
+
+    HIPDNN_LOG_INFO("MIOpen batchnorm forward inference status: {}",
+                    miopenGetErrorString(miopen_status));
```
SamuelReeder: I see your status todo, perhaps wrap miopen API calls in some sort of `MIOPEN_CHECK` macro:
```cpp
#define MIOPEN_CHECK(status)                                                             \
    do                                                                                   \
    {                                                                                    \
        if(status != miopenStatusSuccess)                                                \
        {                                                                                \
            std::cerr << "MIOpen Error: " << miopenGetErrorString(status) << " in file " \
                      << __FILE__ << " at line " << __LINE__ << std::endl;               \
        }                                                                                \
    } while(0)
```

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -279,22 +294,41 @@ hipdnnPluginStatus_t hipdnnEnginePluginCreateExecutionContext(
...
+                  static_cast<void*>(execution_context));
+
+    return hipdnn_plugin::try_catch([&, api_name = __func__]() {
+        throw_if_null(handle);
+        throw_if_null(execution_context);
 
-    // TODO: Destroy MIOpen execution context and free resources
-    return HIPDNN_PLUGIN_STATUS_INTERNAL_ERROR;
+        delete execution_context;
+        execution_context = nullptr;
```
SamuelReeder: Don't need this line.

SamuelReeder (commented):
I didn't look at any of the testing yet, and will revisit this PR later, but overall the high-level methodology looks pretty good to me.

plugins/miopen_legacy_plugin/engine_manager.cpp
adickin-amd: removed

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: i always like setting the ptr to null after but in this case it wont matter so I have removed.

plugins/miopen_legacy_plugin/engines/solvers/miopen_batchnorm_solver.cpp
adickin-amd: agreed, I will do that.  Theres a few things in the solver I still need to address from a code standpoint.  The struct I used to combine tensor/device needs to be extracted and the mapping function is likely a useful extraction as well.  I assume most miopen solvers will need to map tensor to the raw buffer.

plugins/miopen_legacy_plugin/engine_manager.cpp
adickin-amd: good catch, i had duplicated it 3 times

adickin-amd:
> I didn't look at any of the testing yet, and will revisit this PR later, but overall the high-level methodology looks pretty good to me.

If you or anyone else gets a chance to pull and build the miopen plugin and run it that would be awesome.  We do not have CI on the plugin yet.

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/graph_wrapper.hpp
```diff
@@ -0,0 +1,109 @@
...
+            if(verifier.VerifyBuffer<hipdnn_sdk::data_objects::Graph>())
+            {
+                _graph = flatbuffers::GetRoot<hipdnn_sdk::data_objects::Graph>(buffer);
+            }
+        }
+    }
+
+    const hipdnn_sdk::data_objects::Graph& get_graph() const override
+    {
+        return *_graph;
```
mousdahl-amd: Should we check is_valid() before deref?

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/graph_wrapper.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+    uint node_count() const override
+    {
+        return _graph ? _graph->nodes()->size() : 0;
+    }
+
+    bool has_only_supported_attributes(
+        std::set<hipdnn_sdk::data_objects::NodeAttributes> supported_attributes) const override
+    {
+        if(!_graph)
```
mousdahl-amd: null checking seems a bit inconsistent.  Sometimes we're checking against nullptr and returning, other times we're throwing runtime_error.  And we've got is_valid() we could possibly leverage.

sdk/include/hipdnn_sdk/plugin/plugin_flatbuffer_utilities.hpp
```diff
@@ -37,5 +39,49 @@ inline void convert_serialized_plugin_graph_to_graph(
...
+    }
+
+    auto engine_details = hipdnn_sdk::data_objects::UnPackEngineDetails(buffer);
+    if(engine_details == nullptr)
+    {
+        throw Hipdnn_plugin_exception(HIPDNN_PLUGIN_STATUS_INTERNAL_ERROR,
+                                      "Invalid buffer: unable to unpack the flatbuffer schema.");
+    }
+
+    engine_details_out = std::move(engine_details);
```
mousdahl-amd: Any reason the unique_ptr couldn't be the return type for these methods?

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include "hipdnn_engine_plugin_handle.hpp"
+
+#define HIP_CHECK(status)                                                                      \
+    do                                                                                         \
+    {                                                                                          \
+        if(status != hipSuccess)                                                               \
+        {                                                                                      \
+            std::cerr << "HIP Error: " << hipGetErrorString(status) << " in file " << __FILE__ \
+                      << " at line " << __LINE__ << "\n";                                      \
+            exit(EXIT_FAILURE);                                                                \
```
mousdahl-amd: Would it make more sense to fail the test here rather than just exit()?

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
```diff
@@ -19,29 +20,64 @@ int64_t Miopen_engine::id() const
...
+
+size_t Miopen_engine::get_workspace_size(const hipdnnEnginePluginHandle& handle,
+                                         const hipdnn_plugin::Graph_interface& op_graph) const
+{
+    size_t workspace_size = 0;
+    for(const auto& solver : _solvers)
+    {
+        if(solver->is_applicable(op_graph))
+        {
+            workspace_size = solver->get_workspace_size(handle, op_graph);
```
mousdahl-amd: Shouldn't we return std::max workspace size?

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/graph_wrapper.hpp
adickin-amd: ah yes I should,

sdk/include/hipdnn_sdk/plugin/plugin_flatbuffer_utilities.hpp
adickin-amd: it could be.  I duplicated these from the backend where originally the return was a status that got changed to void as we decided to use exceptions for error handling.  I need to take another look at this util as the GraphWrapper might replace the need for this in the plugins.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
adickin-amd: ah probably, this is the hipcheck stuff I want to refactor so Ill do that then.

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
adickin-amd: good catch, i forgot to do this.

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/graph_wrapper.hpp
adickin-amd: replacing all the checks with a `throw_if_not_valid()` call.

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/graph_wrapper.hpp
adickin-amd: added internal throw_if_not_valid function.

sdk/include/hipdnn_sdk/plugin/plugin_flatbuffer_utilities.hpp
adickin-amd: these are now used in the tests only.  Im going to delete this utility so that the wrappers are used.

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
adickin-amd: done

plugins/miopen_legacy_plugin/engines/engine_interface.hpp
```diff
@@ -0,0 +1,36 @@
...
+    virtual size_t get_workspace_size(const hipdnnEnginePluginHandle& handle,
+                                      const hipdnn_plugin::Graph_interface& op_graph) const
+        = 0;
+
+    virtual void execute_graph(const hipdnnEnginePluginHandle& handle,
+                               const hipdnnEnginePluginExecutionContext& execution_context,
+                               const hipdnnPluginDeviceBuffer_t* device_buffers,
+                               uint32_t num_device_buffers,
+                               void* workspace) const
+        = 0;
```
BrianHarrisonAMD: Maybe it's coming next, but I would expect that each engine gets called to provide execution context details as well, or some mechanism for caching the invoker inside the execution-context.

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
```diff
@@ -19,29 +19,68 @@ int64_t Miopen_engine::id() const
...
+    flatbuffers::FlatBufferBuilder builder;
+    auto engine_details = hipdnn_sdk::data_objects::CreateEngineDetails(builder, _id);
+    builder.Finish(engine_details);
+    auto serialized_details = builder.Release();
+
+    auto* temp_buffer = new uint8_t[serialized_details.size()];
+    std::memcpy(temp_buffer, serialized_details.data(), serialized_details.size());
+
+    details_out.ptr = temp_buffer;
+    details_out.size = serialized_details.size();
```
BrianHarrisonAMD, line 47: Can't you set this directly once it's released? 
I thought there was a way to steal ownership, and avoid the copy.

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
```diff
@@ -19,29 +19,68 @@ int64_t Miopen_engine::id() const
...
+    size_t workspace_size = 0;
+    for(const auto& solver : _solvers)
+    {
+        if(solver->is_applicable(op_graph))
+        {
+            size_t solver_workspace = solver->get_workspace_size(handle, op_graph);
+            if(solver_workspace > workspace_size)
+            {
+                workspace_size = solver_workspace;
+            }
```
BrianHarrisonAMD: nit: This can be swapped to std::max one liner without temporary.

`workspace_size = std::max(solver->get_workspace_size(handle, op_graph), workspace_size);`

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
```diff
@@ -19,29 +19,68 @@ int64_t Miopen_engine::id() const
...
+
+    for(const auto& solver : _solvers)
+    {
+        if(solver->is_applicable(execution_context.graph()))
+        {
+            solver->execute_graph(
+                handle, execution_context, device_buffers, num_device_buffers, workspace);
+        }
+    }
 }
```
BrianHarrisonAMD: Should break after executing, and I guess we should be aware that there might be overlapping applicability in the future and this would need to have some ordering logic I suppose. 

Also, we probably want to add some logging to miopen_engine for tracking which solver is selected / applicable etc. Maybe we are planning on doing that inside each solver, but could be nice to also log which solver was selected to handle the work just in case.

plugins/miopen_legacy_plugin/engines/solvers/miopen_batchnorm_solver.cpp
```diff
@@ -2,21 +2,187 @@
...
+                              strides.data());
+
+    for(uint32_t i = 0; i < num_device_buffers; i++)
+    {
+        if(tensor.uid() == device_buffers[i].uid)
+        {
+            pair.device_buffer = device_buffers[i];
+            break;
+        }
+    }
```
BrianHarrisonAMD: Probably would be good to add a utiltiy for this, since I think everyone will need it eventually to use plugin device buffers. 

Can be done later on though as a separate small task if you prefer.

plugins/miopen_legacy_plugin/engines/solvers/miopen_batchnorm_solver.cpp
```diff
@@ -2,21 +2,187 @@
...
+
+    const auto& tensor_map = op_graph.get_tensor_map();
+
+    MiOpenTensorAndDeviceBufferPair xDesc;
+    MiOpenTensorAndDeviceBufferPair yDesc;
+
+    MiOpenTensorAndDeviceBufferPair scaleDesc;
+    MiOpenTensorAndDeviceBufferPair biasDesc;
+    MiOpenTensorAndDeviceBufferPair estMeanDesc;
+    MiOpenTensorAndDeviceBufferPair estVarianceDesc;
```
BrianHarrisonAMD: Are these getting leaked?
I think we need to have a scoped destructor that cleans these up, and they should be associated with the execution context to clean-up only once that's gone.

plugins/miopen_legacy_plugin/engines/solvers/solver_interface.hpp
```diff
@@ -0,0 +1,35 @@
...
+    virtual size_t get_workspace_size(const hipdnnEnginePluginHandle& handle,
+                                      const hipdnn_plugin::Graph_interface& op_graph) const
+        = 0;
+
+    virtual void execute_graph(const hipdnnEnginePluginHandle& handle,
+                               const hipdnnEnginePluginExecutionContext& execution_context,
+                               const hipdnnPluginDeviceBuffer_t* device_buffers,
+                               uint32_t num_device_buffers,
+                               void* workspace = nullptr) const
+        = 0;
```
BrianHarrisonAMD: Probably need to pass execution context down to solver_interface as well to store state to allow for instantly re-running over and over.

plugins/miopen_legacy_plugin/hipdnn_engine_plugin_execution_context.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+        temp_buffer = new uint8_t[op_graph_ptr->size];
+        std::memcpy(temp_buffer, op_graph_ptr->ptr, op_graph_ptr->size);
+        _op_graph = hipdnnPluginConstData_t{temp_buffer, op_graph_ptr->size};
+
+        _graph_interface
+            = std::make_unique<hipdnn_plugin::Graph_wrapper>(_op_graph.ptr, _op_graph.size);
+
+        _engine_config_interface = std::make_unique<hipdnn_plugin::Engine_config_wrapper>(
+            _engine_config.ptr, _engine_config.size);
```
BrianHarrisonAMD: If we pass the execution context down to the solver level, then it can probably store the MIOpen state needed instead of copying over the graph details.

Can probably have some base invoker that can be passed to the execution context during creation, or maybe we can do something similar to the way MIOpen does it's invoker pattern with lambda's, but I think the MIOpen invoker doesn't have the need to free memory like we do here. Might need an invoker_interface base class that each solver extends and adds it's own init / execute / clean-up phase.

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/engine_details_wrapper.hpp
```diff
@@ -0,0 +1,72 @@
...
+    virtual ~Engine_details_interface() = default;
+
+    virtual const hipdnn_sdk::data_objects::EngineDetails& get_engine_details() const = 0;
+    virtual bool is_valid() const = 0;
+    virtual int64_t engine_id() const = 0;
+};
+
+class Engine_details_wrapper : public Engine_details_interface
+{
+public:
```
BrianHarrisonAMD, line 27: We probably should add a comment making it clear that this doesn't handle cleaning up the memory, and that's intentional due to the way SDK & plugin system works. Same is true for all the wrappers I think.

BrianHarrisonAMD (changes requested):

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
adickin-amd: wrapped in Gpu_memory class

plugins/miopen_legacy_plugin/engines/solvers/miopen_batchnorm_solver.cpp
adickin-amd: yea this needs to be extracted. 

https://github.com/ROCm/hipDNN/pull/64#discussion_r2201873356

plugins/miopen_legacy_plugin/engines/solvers/miopen_batchnorm_solver.cpp
adickin-amd: most likely, will ensure they get deleted when I extract that struct out.

plugins/miopen_legacy_plugin/engines/solvers/solver_interface.hpp
adickin-amd: it does pass the execution_context here.  Are you thinking I should pass it as non const so it can be modified?

BrianHarrisonAMD: I mean it should pass it down to the solver_interface non-const during creation to provide the invoker details + cached artifacts needed to run it without additional set-up. Solver should basically be creating the invoker that the ExecutionContext holds onto, and invoker stores all state needed for re-running (descriptors + ptrs).

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
adickin-amd: ah, we dont want to use Release, its ReleaseRaw to take ownership

plugins/miopen_legacy_plugin/tests/test_hipdnn_plugin_execution_context.cpp
```diff
@@ -0,0 +1,50 @@
...
+    hipdnnPluginConstData_t engine_config_ptr{engine_config.data.data(),
+                                              sizeof(engine_config.data)};
+    hipdnnPluginConstData_t op_graph_ptr{op_graph.data.data(), sizeof(op_graph.data)};
+
+    hipdnnEnginePluginExecutionContext ctx(&engine_config_ptr, &op_graph_ptr);
+
+    auto& graph = ctx.graph();
+    auto& engine = ctx.engine_config();
+    (void)graph;
+    (void)engine;
```
BrianHarrisonAMD: I guess this is testing that it doesn't throw for valid cases? 
Looked a bit surprising with no checks to me, but I guess that makes sense.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    hipdnnEnginePluginHandle_t _handle = nullptr;
+};
+
+TEST_F(Batchnorm_execute_graph_test, RunBatchnormGraph)
+{
+    std::vector<int64_t> strides = {1, 3, 224, 224}; // always in nchw
+    std::vector<int64_t> dims = {1, 3, 224, 224};
```
BrianHarrisonAMD, line 42: The strides don't seem to make sense.
Actually, surprised MIOpen doesn't complain since these would be strides for WHCN or HWCN layout hehe.

This might be showing a bug inside MIOpen.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
```diff
@@ -0,0 +1,103 @@
...
+    hipdnnEnginePluginCreateExecutionContext(
+        _handle, &engine_config, &op_graph, &execution_context);
+
+    hipdnnPluginStatus_t status
+        = hipdnnEnginePluginExecuteOpGraph(_handle,
+                                           execution_context,
+                                           nullptr,
+                                           device_buffers.data(),
+                                           static_cast<uint32_t>(device_buffers.size()));
+    EXPECT_EQ(status, HIPDNN_PLUGIN_STATUS_SUCCESS);
```
BrianHarrisonAMD, line 102: I am increasingly thinking this might be a bug inside MIOpen if it's returning success after executing this. 

I guess maybe it just treats it as NCHW layout, and uses the provided strides for accessing the elements. If it was created matching these strides, then it could work. but just be really slow performance. It could also just generate incorrect results depending on how the kernel is implemented and how it assumes the memory is laid out.

I guess either way, it's a bit troubling hehe.

BrianHarrisonAMD: I guess hard to say, I looked at it, and the applicable check doesn't seem to care about layout, but then there is some checks to change behavior depending on the layout. 

Might be fine, but just poor performance with these strange strides, or might actually break.

plugins/miopen_legacy_plugin/tests/test_hipdnn_plugin_execution_context.cpp
adickin-amd: that was the intent.  We shouldnt get throws on init.  This is going to be refactored out

adickin-amd: im actually thinking of making the wrappers throw if they fail to verify the buffer.

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/engine_details_wrapper.hpp
adickin-amd: ill do that and name the pointers they hold as shallow_xyz to indicate they are shallow and dont need to be deleted.

plugins/miopen_legacy_plugin/tests/test_hipdnn_plugin_execution_context.cpp
BrianHarrisonAMD: We want plugins to early exit if they don't understand buffers, but it shouldn't be a break. It might be a new graph that you don't understand.

plugins/miopen_legacy_plugin/tests/test_miopen_legacy_engine_plugin_api.cpp
```diff
@@ -144,36 +149,174 @@ TEST(MiopenLegacyEnginePluginApiTest, EnginePluginGetApplicableEngineIdsValid)
...
+    auto serialized_graph = builder.Release();
+    hipdnnPluginConstData_t op_graph
+        = flatbuffer_test_utils::create_valid_const_data_graph(serialized_graph);
+
+    // For now, engine_config is not used, so we can pass op_graph as a placeholder
+    hipdnnPluginConstData_t* engine_config = &op_graph;
+
+    size_t workspace_size = 0;
+    auto status
+        = hipdnnEnginePluginGetWorkspaceSize(handle, engine_config, &op_graph, &workspace_size);
```
BrianHarrisonAMD: That's surprising. 
Does it actually work?

BrianHarrisonAMD: Oh I see, we literally don't use it at all. Not that we were grabbing the graph out of it haha.

plugins/miopen_legacy_plugin/tests/test_hipdnn_plugin_execution_context.cpp
adickin-amd: here would be a throw as we cant verify the buffer is a graph at all.

BrianHarrisonAMD: Probably fine, but however it gets handled above needs to treat it as the plugin being not applicable.
I think it will fail to verify if new additions are added, and the plugins need to handle that gracefully.

BrianHarrisonAMD: Can also just do what makes sense for now, and we have future work to ensure this is behaving correctly.

BrianHarrisonAMD:
Well not too surprising, but that is a ton of work! 

Thanks for doing all this @adickin-amd !

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
adickin-amd: Maybe there is a secret feature we dont know about :P.  I wouldnt get to hung up on this anyway, im sure the results are not valid, were only testing for success at this point.

plugins/miopen_legacy_plugin/CMakeLists.txt
```diff
@@ -41,11 +41,14 @@ find_package(miopen CONFIG REQUIRED)
 
 add_library(miopen_legacy_plugin SHARED
     engines/miopen_engine.cpp
-    engines/solvers/miopen_batchnorm_solver.cpp
```
adickin-amd, line 44: todo, I need to check this cmake list.  I dont think the warnings are generating as they should be while compiling.

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
adickin-amd: still todo.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_fwd_inference_plan.cpp
```diff
@@ -0,0 +1,164 @@
...
+{
+    return *_scale_pair;
+}
+
+const Miopen_tensor& Batchnorm_fwd_inference_params::bias() const
+{
+    return *_bias_pair;
+}
+
+const std::optional<const Miopen_tensor*> Batchnorm_fwd_inference_params::est_mean() const
```
adickin-amd: thoughts on returning `std::optional<const Miopen_tensor*>` here?  as far as I read, you cant do a reference in a std::optional yet.  Alternatively I can probably return a `const std::optional<std::unique_ptr<Miopen_tensor>>&`.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_fwd_inference_plan.cpp
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+Batchnorm_fwd_inference_plan::Batchnorm_fwd_inference_plan(
+    std::unique_ptr<Batchnorm_fwd_inference_params> inference_params)
+    : _inference_params(std::move(inference_params))
+{
+}
+
+void Batchnorm_fwd_inference_plan::execute(const hipdnnEnginePluginHandle& handle,
```
adickin-amd, line 95: note: this function is not unit tested any only tested via the batch norm operations test.

.clang-tidy
```diff
@@ -26,7 +26,7 @@ HeaderFileExtensions: ['h','hpp']
 
 # This regex is a inclusion filter for the files to be checked. Its a Posix regex and
 # does not seem to support negative lookahead.  
-HeaderFilterRegex: '^(.*backend.*|.*frontend.*|.*sdk/include/hipdnn_sdk/logging.*)$'
+HeaderFilterRegex: '^(.*backend.*|.*frontend.*|.*sdk/include/hipdnn_sdk/logging.*|.*sdk/include/hipdnn_sdk/plugin.*|.*plugins/miopen_legacy_plugin.*)$'
```
mousdahl-amd: Maybe the last bit could be `.*plugins/.*_plugin.*` to future proof things a bit.  As long as we keep calling stuff "blahblah_plugin" at least.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_fwd_inference_plan.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+    // these need to be optional
+    std::unique_ptr<Miopen_tensor> _est_mean_tensor_descriptor;
+    std::unique_ptr<Miopen_tensor> _est_variance_tensor_descriptor;
+};
+
+class Batchnorm_fwd_inference_plan : public Plan_interface
+{
+public:
+    const miopenBatchNormMode_t MIOPEN_BATCHNORM_MODE = miopenBNSpatial;
```
adickin-amd: should move to cpp only.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_plan_builder.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "engines/plans/miopen_batchnorm_fwd_inference_plan.hpp"
+#include "miopen_batchnorm_plan_builder.hpp"
+
+namespace miopen_legacy_plugin
+{
+
+// We have made the intentional decision to hardcode the batchnorm mode to miopenBNSpatial
+// rather than making it configurable and adding extra complexity.
+const miopenBatchNormMode_t miopen_batchnorm_mode = miopenBNSpatial;
```
adickin-amd: remove this from the builder, its only used for execute.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_plan_builder.cpp
```diff
@@ -0,0 +1,86 @@
...
+    const hipdnnEnginePluginHandle& handle,
+    const hipdnn_plugin::Graph_interface& op_graph,
+    hipdnnEnginePluginExecutionContext& execution_context) const
+{
+    const auto& node = op_graph.get_node(0);
+
+    if(node.attributes_type()
+       == hipdnn_sdk::data_objects::NodeAttributes_BatchnormInferenceAttributes)
+    {
+        //HIPDNN_LOG_INFO("Building batchnorm fwd inference plan for node: {}", node.name());
```
adickin-amd: fix this log.  node.name() needs ->str() added.

plugins/miopen_legacy_plugin/hipdnn_engine_plugin_execution_context.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include <hipdnn_sdk/data_objects/engine_config_generated.h>
+#include <hipdnn_sdk/data_objects/graph_generated.h>
+#include <hipdnn_sdk/logging/logger.hpp>
+#include <hipdnn_sdk/plugin/flatbuffer_utilities/engine_config_wrapper.hpp>
+#include <hipdnn_sdk/plugin/flatbuffer_utilities/graph_wrapper.hpp>
+#include <hipdnn_sdk/plugin/plugin_api_data_types.h>
+
+#include "engines/plans/plan_interface.hpp"
+
+struct hipdnnEnginePluginExecutionContext
```
adickin-amd, line 17: Note: I could move the implementation into a cpp file.

plugins/miopen_legacy_plugin/hipdnn_engine_plugin_execution_context.hpp
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    void set_plan(std::unique_ptr<miopen_legacy_plugin::Plan_interface> plan)
+    {
+        _plan = std::move(plan);
+    }
+
+    virtual miopen_legacy_plugin::Plan_interface& plan() const
+    {
+        return *_plan;
```
adickin-amd, line 40: should we throw if you try to get a plan when there isnt one?

plugins/miopen_legacy_plugin/miopen_container.hpp
```diff
@@ -12,7 +12,7 @@ namespace miopen_legacy_plugin
 class Engine_manager;
 
 /*
- * Container class to manage the intantiation and ownership of all MIOpen solvers and engines.
+ * Container class to manage the intantiation and ownership of all MIOpen plan builders and engines.
```
adickin-amd, line 15: todo, fix documentation to no longer reference solvers but instead call things plans.

plugins/miopen_legacy_plugin/miopen_tensor.cpp
```diff
@@ -0,0 +1,44 @@
+// Copyright © Advanced Micro Devices, Inc., or its affiliates.
+// SPDX-License-Identifier:  MIT
+
+#include "miopen_tensor.hpp"
+#include "miopen_utils.hpp"
+
+namespace miopen_legacy_plugin
+{
+
+Miopen_tensor::Miopen_tensor(const hipdnn_sdk::data_objects::TensorAttributes& tensor)
```
adickin-amd, line 10: todo, this class isnt tested.

plugins/miopen_legacy_plugin/miopen_utils.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include "miopen_utils.hpp"
+
+namespace miopen_legacy_plugin
+{
+
+namespace miopen_utils
+{
+
+//might make more sense in plan utils
```
adickin-amd: remove comment, this is fine here.

plugins/miopen_legacy_plugin/miopen_utils.hpp
```diff
@@ -0,0 +1,46 @@
...
+        }                                                                               \
+    } while(0)
+
+namespace miopen_legacy_plugin
+{
+
+namespace miopen_utils
+{
+
+//might make more sense in plan utils
```
adickin-amd: remove comment

plugins/miopen_legacy_plugin/tests/CMakeLists.txt
```diff
@@ -34,12 +34,16 @@ if(HIP_DNN_SKIP_TESTS)
...
+    engines/plans/test_miopen_batchnorm_plan_builder.cpp
+    engines/plans/test_batchnorm_fwd_inference_plan.cpp
     main.cpp
     test_miopen_legacy_engine_plugin_api.cpp
+    test_miopen_batchnorm_operations.cpp
     test_miopen_legacy_plugin_api.cpp
     test_miopen_handle_factory.cpp
     test_engine_manager.cpp
-    test_miopen_engine.cpp
+    test_hipdnn_plugin_execution_context.cpp
```
adickin-amd, line 28: todo, check warnings as errors is working correctly here.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
```diff
@@ -0,0 +1,103 @@
...
+        if(_handle != nullptr)
+        {
+            hipdnnEnginePluginDestroy(_handle);
+        }
+    }
+
+    hipdnnEnginePluginHandle_t _handle = nullptr;
+};
+
+TEST_F(Batchnorm_execute_graph_test, RunFwdbatchnormGraph)
```
adickin-amd, line 39: todo, add the check for GPU before running this test.

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/engine_details_wrapper.hpp
adickin-amd: still todo

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
adickin-amd: this is gone, refactored

plugins/miopen_legacy_plugin/engines/solvers/miopen_batchnorm_solver.cpp
adickin-amd: refactored out

plugins/miopen_legacy_plugin/engines/solvers/miopen_batchnorm_solver.cpp
adickin-amd: refactored into Miopen_tensor class

plugins/miopen_legacy_plugin/engines/solvers/solver_interface.hpp
adickin-amd: refactored as discussed

plugins/miopen_legacy_plugin/hipdnn_engine_plugin_execution_context.hpp
adickin-amd: refactored

plugins/miopen_legacy_plugin/tests/test_hipdnn_plugin_execution_context.cpp
adickin-amd: this is no longer valid as we dont hold a graph on the execution context.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_fwd_inference_plan.cpp
mousdahl-amd: Maybe a "bool try_get_est_mean(Miopen_tensor** tensor)" style for yet another alternative?

mousdahl-amd: My vote is for the second thing, if we're just handing out references.  Change the member to std::optional.

plugins/miopen_legacy_plugin/engines/solvers/miopen_batchnorm_solver.cpp
adickin-amd: added two miopen checks.  1 that throws and another that doesnt

mousdahl-amd (commented):
I think it's looking great.  I'm not sure I'm knowledgeable enough to approve yet, though.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_fwd_inference_plan.cpp
adickin-amd: I like the try get but if we can avoid another raw ptr thats my goal.

.clang-tidy
adickin-amd: will attempt

plugins/miopen_legacy_plugin/engines/miopen_engine.cpp
adickin-amd: @BrianHarrisonAMD ReleaseRaw returns a size and offset.  I cant just change to ReleaseRaw as I think I need the offset it returns as well.  I tried just using the returned buffer and size but that doesnt unpack correctly in the test.  

```
  /// @brief Get the released pointer to the serialized buffer.
  /// @param size The size of the memory block containing
  /// the serialized `FlatBuffer`.
  /// @param offset The offset from the released pointer where the finished
  /// `FlatBuffer` starts.
  /// @return A raw pointer to the start of the memory block containing
  /// the serialized `FlatBuffer`.
  /// @remark If the allocator is owned, it gets deleted when the destructor is
  /// called..
  uint8_t *ReleaseRaw(size_t &size, size_t &offset) {
```

adickin-amd: we could add an offset field but that seems like itll get confusing.  Im looking to see if there is another API we can use instead.

adickin-amd: confirmed if I do it unpacks correctly but then the `ptr` is not valid to be freed.
```
    size_t size = 0;
    size_t offset = 0;
    auto* buffer = builder.ReleaseRaw(size, offset);

    details_out.ptr = buffer + offset;
    details_out.size = size;
```

BrianHarrisonAMD: Okay the flatbuffers are a little bit awkward since they use an offset, and we need to free the full buffer, but we need offset to unpack it.

Lets leave a todo, and create a future task to store the flatbuffer object on the handle, and associate it with the const_data *. Then when we destroy, we can just clear the associated object, and we don't need to copy memory, or free these buffers. The flatbuffer object can free it's own memory once it's no longer referenced by the handle.

adickin-amd: created a jira ticket to handle this after this PR is complete

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_fwd_inference_plan.cpp
adickin-amd: ives changed this to return the std::optional ref.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_fwd_inference_plan.cpp
```diff
@@ -0,0 +1,163 @@
...
+        _bias_pair = std::make_unique<Miopen_tensor>(*bias_tensor_attr->second);
+    }
+
+    if(attributes.mean().has_value())
+    {
+        if(auto est_mean_attr = tensor_map.find(attributes.mean().value());
+           est_mean_attr != tensor_map.end())
+        {
+            _est_mean_tensor_descriptor = std::make_unique<Miopen_tensor>(*est_mean_attr->second);
+        }
```
BrianHarrisonAMD, line 83: Should throw, or log if it's missing and the attributes expects it I think.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_fwd_inference_plan.cpp
```diff
@@ -0,0 +1,163 @@
...
+    if(attributes.inv_variance().has_value())
+    {
+        if(auto inv_variance_attr = tensor_map.find(attributes.inv_variance().value());
+           inv_variance_attr != tensor_map.end())
+        {
+            _est_variance_tensor_descriptor
+                = std::make_unique<Miopen_tensor>(*inv_variance_attr->second);
+        }
+    }
+}
```
BrianHarrisonAMD: These should throw if they aren't found, and the attributes is expecting them.

plugins/miopen_legacy_plugin/tests/CMakeLists.txt
adickin-amd: fixed

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_fwd_inference_plan.cpp
```diff
@@ -0,0 +1,163 @@
...
+        _inference_params->y().tensor_descriptor(),
+        y_buffer.ptr,
+        _inference_params->scale().tensor_descriptor(),
+        _inference_params->bias().tensor_descriptor(),
+        _inference_params->est_mean().has_value()
+            ? _inference_params->est_mean().value()->tensor_descriptor()
+            : nullptr,
+        _inference_params->est_variance().has_value()
+            ? _inference_params->est_variance().value()->tensor_descriptor()
+            : nullptr,
```
BrianHarrisonAMD, line 146: nit: could set these pointers at the same time as the buffers, but not a big deal either way.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_fwd_inference_plan.cpp
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    hipdnnPluginDeviceBuffer_t est_variance_buffer = {0, nullptr};
+    if(_inference_params->est_variance().has_value())
+    {
+        est_variance_buffer = miopen_utils::find_device_buffer(
+            _inference_params->est_variance().value()->uid(), device_buffers, num_device_buffers);
+    }
+
+    auto miopen_status = miopenBatchNormalizationForwardInference_V2(
```
BrianHarrisonAMD: Should we throw on failed status here?

.clang-tidy
adickin-amd: its now just plugins/.* so they all get added.

plugins/miopen_legacy_plugin/hipdnn_engine_plugin_execution_context.hpp
```diff
@@ -0,0 +1,46 @@
...
+    virtual ~hipdnnEnginePluginExecutionContext() = default;
+
+    bool has_valid_plan() const
+    {
+        return _plan != nullptr;
+    }
+
+    void set_plan(std::unique_ptr<miopen_legacy_plugin::Plan_interface> plan)
+    {
+
```
BrianHarrisonAMD: nit: extra newline

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_fwd_inference_plan.cpp
adickin-amd: as in, create a temporary variable for these where I create the buffer?

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -244,19 +222,42 @@ hipdnnPluginStatus_t hipdnnEnginePluginGetEngineDetails(hipdnnEnginePluginHandle
...
+    LOG_API_ENTRY("handle={:p}, engine_details={}",
+                  static_cast<void*>(handle),
+                  static_cast<void*>(engine_details));
+
+    return hipdnn_plugin::try_catch([&, api_name = __func__]() {
+        throw_if_null(handle);
+        throw_if_null(engine_details);
+        throw_if_null(engine_details->ptr);
+
+        delete[] static_cast<const uint8_t*>(engine_details->ptr);
```
BrianHarrisonAMD, line 259: It'll be nice to avoid this, and rely on RAII once we swap to keeping the memory associated with the handle.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_fwd_inference_plan.cpp
adickin-amd: sure, wrapped in THROW_ON_MIOPEN_FAILURE

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: yep, next PR.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -265,12 +266,26 @@ hipdnnPluginStatus_t
...
+        throw_if_null(handle);
+        throw_if_null(engine_config);
+        throw_if_null(op_graph);
+        throw_if_null(workspace_size);
+
+        auto& engine_manager = handle->get_engine_manager();
+
+        //todo, use the engine_config
+        Graph_wrapper op_graph_wrapper(op_graph->ptr, op_graph->size);
+        *workspace_size = engine_manager.get_workspace_size(*handle, 1, op_graph_wrapper);
```
BrianHarrisonAMD: Need to grab the engine_id from engine_config here I imagine.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
BrianHarrisonAMD: Probably many places hehe.

adickin-amd: fixed

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: fixed

BrianHarrisonAMD (approved):
LGTM, thanks!

adickin-amd:
bypassing merge requirements, CI is down and this will become a major pain to deal with constant develop merges if it lives too much longer.

PR #65, opened 2025-07-11
Added Address Sanitizer

mousdahl-amd:
## Proposed changes
- Added some matching destroy() methods as siblings to create methods.  I feel like it's a good habit if we're using opaque pointers generally, and it also makes searching for matching create / destroy pairs easier when you're trying to diagnose a leak.
- Added tests for said methods
- Added ASAN to the CMakeFiles
- I attempted to add UBSAN, but was running into compilation problems when hip was building architecture libraries behind the scenes.  I've added a story to revisit it when we go to hipRTC
- Added GPU_TARGETS, even though we might remove them later
- Cleaned up a lot of leaky tests.

## Checklist


backend/src/descriptors/descriptor_factory.cpp
```diff
@@ -72,4 +72,14 @@ void Descriptor_factory::create_graph_ext(hipdnnBackendDescriptor_t* descriptor,
     *descriptor = graph_descriptor;
 }
 
+void Descriptor_factory::destroy(hipdnnBackendDescriptor_t descriptor)
+{
+    THROW_IF_NULL(
+        descriptor, HIPDNN_STATUS_BAD_PARAM_NULL_POINTER, "hipdnnBackendDescriptor_t is null.");
+
+    HIPDNN_LOG_INFO("Destroying descriptor");
```
mousdahl-amd: Oh, that's right, I thought it might be nice to log the descriptor address so you could observe their lifetimes, but I was struggling to get pointers logged.  I'll take another swing at that now that everything else is out of the way.

backend/src/handle/handle_factory.cpp
```diff
@@ -18,4 +18,14 @@ void Handle_factory::create_handle(hipdnnHandle_t* handle)
     *handle = new hipdnnHandle();
 }
 
+void Handle_factory::destroy_handle(hipdnnHandle_t handle)
+{
+    if(handle == nullptr)
+    {
+        throw Hipdnn_exception(HIPDNN_STATUS_BAD_PARAM_NULL_POINTER, "handle is null.");
+    }
```
mousdahl-amd: Should these be logged?

CMakeLists.txt
```diff
@@ -32,9 +32,9 @@ option(HIP_DNN_BUILD_FRONTEND "Build the frontend code" ON)
 option(HIP_DNN_SKIP_TESTS "Skips building all tests" OFF)
 option(HIP_DNN_BUILD_PLUGINS "Builds the plugins as part of the main cmake" OFF)
 
-
 option(CODE_COVERAGE "Build with code coverage flags" OFF)
 option(ENABLE_CLANG_TIDY "Enable Clang-Tidy checks" ON)
+option(BUILD_ADDRESS_SANITIZER "Build with Address Sanitizer enabled" OFF)
```
averinevg, line 37: [R] I'm not insisting, but I think this way would be more informative.
```suggestion
option(ENABLE_ASAN "Build with Address Sanitizer enabled" OFF)
```

CMakeLists.txt
```diff
@@ -61,13 +61,39 @@ list(
...
+        set(GPU_TARGETS 
+            gfx90a
+            gfx942
+            gfx1100
+            gfx1151
+            gfx1201
+        )
+    endif()
+endif()
+message(STATUS "Using GPU targets: ${GPU_TARGETS}")
```
adickin-amd: Thoughts on moving this into the Sanitizers.cmake file as well?  Could put it into a function and still call it here but keep the sanitizer things all in one place.

backend/src/descriptors/descriptor_factory.cpp
adickin-amd: If you havent found the code yet you can log pointers like this: https://github.com/ROCm/hipDNN/blob/develop/backend/src/hipdnn_backend.cpp#L96

CMakeLists.txt
```diff
@@ -61,13 +61,39 @@ list(
...
 # Add a check target which will run all tests discovered by gtest_discover_tests via ctest. 
-add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure -C ${CMAKE_CFG_INTDIR})
+add_custom_target(check COMMAND ${TEST_ENVIRONMENT} ${CMAKE_CTEST_COMMAND} --output-on-failure -C ${CMAKE_CFG_INTDIR})
+
+# GPU Architecture targets
+if(NOT DEFINED GPU_TARGETS)
+    # Check if Address Sanitizer is enabled - use XNACK-enabled targets
+    if(BUILD_ADDRESS_SANITIZER)
+        set(GPU_TARGETS 
+            gfx906:xnack+    # Vega 20 (MI50, MI60)
```
averinevg: `gfx906` should be removed from the list as it is no longer supported ([system-requirements-linux](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/reference/system-requirements.html), [system-requirements-windows](https://rocm.docs.amd.com/projects/install-on-windows/en/latest/reference/system-requirements.html)).

CMakeLists.txt
```diff
@@ -61,13 +61,39 @@ list(
...
+            gfx908:xnack+    # MI100 (Arcturus)
+            gfx90a:xnack+    # MI200 series (MI210, MI250, MI250X)
+            gfx940:xnack+    # MI300A (APU)
+            gfx941:xnack+    # MI300A (APU variant)
```
averinevg: `gfx940` and `gfx941` no longer exist.

CMakeLists.txt
```diff
@@ -61,13 +61,39 @@ list(
...
 
 enable_testing() # Cmake wont discover or run tests without this line
 
 # Add a check target which will run all tests discovered by gtest_discover_tests via ctest. 
-add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure -C ${CMAKE_CFG_INTDIR})
+add_custom_target(check COMMAND ${TEST_ENVIRONMENT} ${CMAKE_CTEST_COMMAND} --output-on-failure -C ${CMAKE_CFG_INTDIR})
+
+# GPU Architecture targets
+if(NOT DEFINED GPU_TARGETS)
+    # Check if Address Sanitizer is enabled - use XNACK-enabled targets
```
averinevg: Why?

backend/src/handle/handle_factory.cpp
adickin-amd: You can add a log info for sure.  The hipdnn_backend.cpp should be logging every api entry/exit so any handle destroy calls should already be logged.

CMakeLists.txt
```diff
@@ -61,13 +61,39 @@ list(
...
+        )
+    else()
+        # Standard GPU targets without XNACK
+        set(GPU_TARGETS 
+            gfx90a
+            gfx942
+            gfx1100
+            gfx1151
+            gfx1201
+        )
```
averinevg: These are not exactly standard targets - I just have the ability to test functionality on them.

backend/tests/test_flatbuffer_utilities.cpp
```diff
@@ -89,17 +95,18 @@ TEST_P(Flatbuffer_invalid_tests, WillNotUnpackInvalidBuffer)
...
 }
 
-INSTANTIATE_TEST_SUITE_P(
-    InvalidBufferTests,
-    Flatbuffer_invalid_tests,
-    ::testing::Values(std::make_pair(static_cast<const uint8_t*>(nullptr), size_t(10)),
-                      std::make_pair(std::array<uint8_t, 10>{0}.data(), size_t(10)),
-                      []() { //Valid graph but incorrect data size
-                          auto builder = Flatbuffer_utilities_test::create_valid_graph();
-                          auto serialized_graph = builder.Release();
```
adickin-amd, line 99: ah nice find.  I can totally see how this leaks.

cmake/Sanitizers.cmake
```diff
@@ -0,0 +1,38 @@
...
+    add_compile_options(${SANITIZER_COMPILE_FLAGS})
+    add_link_options(${SANITIZER_LINK_FLAGS})
+    
+    # Add compile definition for conditional compilation
+    add_compile_definitions(ADDRESS_SANITIZER)
+
+    # Set environment variables for Address Sanitizer
+    set(TEST_ENVIRONMENT 
+        "ASAN_SYMBOLIZER_PATH=${CMAKE_SYMBOLIZER}" 
+        "HSA_XNACK=1"
```
averinevg, line 44: It is unclear why this is necessary. A comment with an explanation is probably needed.

cmake/Sanitizers.cmake
```diff
@@ -0,0 +1,38 @@
...
+        #"ASAN_OPTIONS=halt_on_error=1:abort_on_error=1"
+    )
+    
+    # Disable ROCFFT kernel cache for Address Sanitizer
+    set(ROCFFT_KERNEL_CACHE_ENABLE off)
```
averinevg: The line probably should be removed.

CMakeLists.txt
mousdahl-amd: I was debating back and forth to put it there.  I landed here because I was thinking "if I were a dev and I was wondering why the build was only targeting certain architectures, the first place to look wouldn't be the Sanitizers.cmake file".  But I guess it's easy enough to search it out.  

I can put the defaults here and the overrides in Sanitizers.cmake.

CMakeLists.txt
mousdahl-amd: This came from a conversation with Eiden.  I think they're trying to co-ordinate build options a bit amongst packages.

tests/backend/test_util.cpp
```diff
@@ -138,6 +154,14 @@ void* allocate_tensor_memory([[maybe_unused]] const int64_t* dims,
     return memory;
 }
 
+void free_tensor_memory(void* data_ptr)
```
adickin-amd, line 157: good add.  From looking at the allocate function im guessing that allocate method is going to change eventually.  Once we do that we should make a class/struct that manages the deletion of any memory thats created automatically.

CMakeLists.txt
mousdahl-amd: ASAN requires XNACK

CMakeLists.txt
mousdahl-amd: I think what I'll try then is to just use the default HIP targets (which seems to autodetect whatever is on the system).  I'll leave the ability to override using GPU_TARGETS.  That means I'll still be able to use this mechanism to specify XNACK targets for ASAN builds.

adickin-amd (approved):
approved assuming Ci passes.  Do we need to get any additional utilities installed on our docker container or the math ci's docker container?

tests/backend/test_backend_execute_api.cpp
BrianHarrisonAMD, line 1: I guess this is also fine, but I think @averinevg made a scoped resource utility we could use to create the needed descriptors inside the tests, and have them automatically get freed after using. 

The class is basically the same thing though with your changes, and I think it's fine as is.

BrianHarrisonAMD:
> approved assuming Ci passes. Do we need to get any additional utilities installed on our docker container or the math ci's docker container?

Yea would be good to coordinate/figure out how we can get this added as a CI stage since we will certainly break it if we aren't running it.

BrianHarrisonAMD (approved):
LGTM assuming CI passes.

Nice to see all the leaks caught / fixed, and good thing they were all in the tests!

PR #66, opened 2025-07-14
Engine plugin C-API part 3

averinevg:
## Proposed changes

List of changes:

- Replaced `set_last_error_string` with `Plugin_last_error_manager` from the `SDK`.
- Configured the `check` target to run `SDK` tests by adding them as a dependency.
- Added a generic function for calling plugin functions to unify and simplify the code.
- Added a utility class that manages a resource with a custom destructor.
- Removed an unused file.
- Replaced `HIPDNN_MAX_ERROR_STRING_SIZE` with `HIPDNN_PLUGIN_ERROR_STRING_MAX_LENGTH`.

## Checklist


sdk/include/hipdnn_sdk/plugin/plugin_last_error_manager.hpp
```diff
@@ -37,10 +34,12 @@ class Plugin_last_error_manager
             return status;
         }
 
-        HIPDNN_LOG_ERROR("Error occured in status:{} message:{}", status, message);
+        // TODO spdlog prevents the library from being unloaded (PluginManagerTest.LastErrorOnSecondLoad is failing).
+        // This line can be uncommented once the root cause is identified and fixed.
+        //HIPDNN_LOG_ERROR("Error occured in status:{} message:{}", status, message);
```
averinevg, line 39: ⚓

sdk/tests/CMakeLists.txt
```diff
@@ -13,11 +15,19 @@ add_executable(hipdnn_sdk_tests
...
-)
\ No newline at end of file
+)
+
+add_dependencies(check hipdnn_sdk_tests)
+gtest_discover_tests(
+    hipdnn_sdk_tests
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    DISCOVERY_MODE PRE_TEST
+)
```
averinevg, line 33: ⚓

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -23,7 +23,7 @@ using namespace hipdnn_plugin;
 using namespace miopen_legacy_plugin;
 
 // NOLINTNEXTLINE(modernize-avoid-c-arrays)
-thread_local char Plugin_last_error_manager::last_error[HIPDNN_MAX_ERROR_STRING_SIZE] = "";
+thread_local char Plugin_last_error_manager::last_error[HIPDNN_PLUGIN_ERROR_STRING_MAX_LENGTH] = "";
```
averinevg, line 26: ⚓

adickin-amd: what does the anchor mean?

sdk/include/hipdnn_sdk/plugin/plugin_last_error_manager.hpp
```diff
@@ -4,30 +4,27 @@
...
 
 #include <hipdnn_sdk/logging/logger.hpp>
-#include <hipdnn_sdk/plugin/plugin_api_data_types.h>
+#include <hipdnn_sdk/plugin/plugin_api.h>
 #include <hipdnn_sdk/plugin/plugin_data_type_helpers.hpp>
 #include <hipdnn_sdk/utilities/string_util.hpp>
 
 #include <string>
 
+// NOTE: The last_error variable must be defined in one of the plugin source files:
```
adickin-amd, line 13: thanks for updating the comments here.  I missed pushing a commit that had updated this, my apologies.

adickin-amd: ahh, I had pushed them onto the miopen_last_error_manager.hpp.

sdk/include/hipdnn_sdk/plugin/plugin_last_error_manager.hpp
```diff
@@ -37,10 +34,12 @@ class Plugin_last_error_manager
...
         }
 
-        HIPDNN_LOG_ERROR("Error occured in status:{} message:{}", status, message);
+        // TODO spdlog prevents the library from being unloaded (PluginManagerTest.LastErrorOnSecondLoad is failing).
+        // This line can be uncommented once the root cause is identified and fixed.
+        //HIPDNN_LOG_ERROR("Error occured in status:{} message:{}", status, message);
 
         hipdnn::sdk::utilities::copy_max_size_with_null_terminator(
-            last_error, message, HIPDNN_MAX_ERROR_STRING_SIZE);
+            last_error, message, sizeof(last_error));
```
mousdahl-amd, line 42: Shouldn't this still be `HIPDNN_PLUGIN_ERROR_STRING_MAX_LENGTH`?

BrianHarrisonAMD: Yea I believe it should be since it can't be longer than that.

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
BrianHarrisonAMD: From context of this PR, I think it means bugs discovered.

BrianHarrisonAMD (approved):
LGTM I think the length change should be added, but otherwise nice clean-up, changes, and fixes!

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
averinevg: I just placed it here to draw your attention to these changes.

sdk/include/hipdnn_sdk/plugin/plugin_last_error_manager.hpp
averinevg: Is there something wrong with `sizeof(last_error)`?

BrianHarrisonAMD: No I think they end up being the same thing since last_error is static init'd to that length.
Mostly just a style thing, and protects against future mistakes if someone incorrectly expands the last_error's size.

I am good either way though.

averinevg: If someone expands the last_error's size, then sizeof(last_error) will increase as well, right? Even is someone renames or replaces the constant, it will still work safely. We are using the constant in only one place (well, actually in two - but with this approach, there's probably no way around that).

BrianHarrisonAMD: I thought the size had to match what the backend was expecting from the plugin with null termination, or else it'll get cut-off, but I think it's fine like this. If someone expands last_error, then the log message that the backend grabs will get cut-off, but if they expand last_error and we use the define here, then it gets cut off any ways hehe.

PR #68, opened 2025-07-15
Updated Install Targets

mousdahl-amd:
## Proposed changes
- Moved frontend, backend, and SDK include install locations
- Verified and added backend shared lib install location
- Verified plugin install location
- Modified sample to test find_package() functionality

## Checklist


samples/CMakeLists.txt
```diff
@@ -4,11 +4,25 @@
 cmake_minimum_required(VERSION 3.25.2)
 project(hipdnn_samples VERSION 0.0.1 LANGUAGES CXX)
 set(CMAKE_CXX_STANDARD 20)
+
 find_package(hip REQUIRED)
 
+# enable_language(HIP
+
+set(CMAKE_CXX_COMPILER /opt/rocm/bin/amdclang++)
```
SamuelReeder: Is it fine to hardcode this or should we set env var `CXX=/opt/rocm/llvm/bin/clang++` like MIOpen--so its obvious it might need to be set? Especially since the path I provided differs.

Edit: I looked into it and seems `/opt/rocm/bin/amdclang++` is the better compiler option, especially for a simple sample, so might be fine.

mousdahl-amd: I'm open to any ideas here, I'm a relative neophyte with CMake, so if there's a more proper way to do it, I'd love to know about it.  I could also make a toolchain file for this.

SamuelReeder: I'm no cmake expert either, but you might just be able to remove it and set `CXX` or `CMAKE_CXX_COMPILER` manually if needed. See [Consuming the HIP API in C++ code](https://rocm.docs.amd.com/en/latest/conceptual/cmake-packages.html#consuming-the-hip-api-in-c-code) and below section. That is, assuming we aren't compiling device code.

adickin-amd: honestly that toolchain file was something I cooked up so I didnt have to specify the compiler manually.  I think if you first checked this variable, if its unset,  set it like this so we have a reasonable default.   Then if anyone specifies the compiler we dont overwrite it on them.

adickin-amd: For miopen, the build steps for cmake says you'd always manually specify CXX https://github.com/ROCm/MIOpen?tab=readme-ov-file#building-miopen-from-source

mousdahl-amd: Alright, so I can just ditch that bit entirely then.

adickin-amd: just be sure to update the samples build doc to show you should specify the CXX.

SamuelReeder: I like just checking and not overwriting it like you said. Then we don't need to worry about setting it. However, I'm fine either way.

Edit: If it works without it though, then removing it seems ideal and user can manually set if desired. So I'm fully onboard for just ditching it in that case.

mousdahl-amd: Removing it as an override in the file, but I've updated the README.md with it as an example build command.

adickin-amd (approved):
LGTM

SamuelReeder (approved):

BrianHarrisonAMD (approved):
LGTM

PR #72, opened 2025-07-18
Fix tests failing with logging env vars

SamuelReeder:
## Proposed changes

Quick fix for tests failing when logging env vars set (ALMIOPEN-99). Wasn't sure how easy this would be and it predated my other logging PR so did it separately. 

Also changed a wrong component name here since I don't want to restart CI for the other one.

## Checklist

### DIFF
--- a/backend/src/hipdnn_backend.cpp
+++ b/backend/src/hipdnn_backend.cpp
@@ -212,10 +212,19 @@ HIPDNN_BACKEND_EXPORT hipdnnStatus_t
                                   element_count,
                                   array_of_elements);
 
-        LOG_API_SUCCESS(api_name,
-                        "status={}, retrieved_element_count={}",
-                        hipdnn_backend::hipdnn_get_status_string(HIPDNN_STATUS_SUCCESS),
-                        *element_count);
+        if(element_count == nullptr)
+        {
+            LOG_API_SUCCESS(api_name,
+                            "status={}, element_count_ptr=nullptr",
+                            hipdnn_backend::hipdnn_get_status_string(HIPDNN_STATUS_SUCCESS));
+        }
+        else
+        {
+            LOG_API_SUCCESS(api_name,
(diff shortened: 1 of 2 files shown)


backend/src/hipdnn_backend.cpp
```diff
@@ -206,10 +206,19 @@ HIPDNN_BACKEND_EXPORT hipdnnStatus_t
                                   element_count,
                                   array_of_elements);
 
-        LOG_API_SUCCESS(api_name,
-                        "status={}, retrieved_element_count={}",
-                        hipdnn_backend::hipdnn_get_status_string(HIPDNN_STATUS_SUCCESS),
-                        *element_count);
+        if(element_count)
```
SamuelReeder: Or maybe I should `throw_if_null(element_count)`. Will look into whether it can be null.

Edit: it can be null.

adickin-amd: yea if you get a null element_count or null array_of_elements I think youd want to throw.

adickin-amd: this is after the get attribute call though.  So it would be possible that there is no elements and the call is still successful.

SamuelReeder: There are cases where it packs a descriptor or sets something else without setting elements count. ie.
```cpp
if(element_count != nullptr)
{
    *element_count = 1;
}

pack_descriptor(_engine, array_of_elements);
``` 
So yeah, the element count is sometimes optional and can just populate the array instead.

SamuelReeder: The opposite is also true, can request 0 elements and just get the `element_count` back. I think this is conventional.

SamuelReeder:
I need to rebase this. Hold reviews.

BrianHarrisonAMD (approved):

PR #73, opened 2025-07-21
Fix descriptor memory management

BrianHarrisonAMD:
## Proposed changes

Please describe the motivation behind the pull request, whether it enables a new feature or fixes a bug. If there are associated pull requests or issues, please link them to the pull request.

## Checklist

- [] I have added relevant documentation for the changes


backend/src/descriptors/backend_descriptor.hpp
BrianHarrisonAMD, line 1: Should move at least the Scoped_descriptor to another file.
Maybe move the private as well.

tests/backend/test_engine_heuristic_api.cpp
```diff
@@ -422,24 +429,9 @@ TEST_F(Engine_heuristic_api_tests, GetEngineConfigsRequestMoreThanAvailable)
...
-                                                1,
-                                                nullptr,
-                                                &engine),
-                      HIPDNN_STATUS_SUCCESS);
-            ASSERT_NE(engine, nullptr);
-            EXPECT_EQ(hipdnnBackendDestroyDescriptor(engine), HIPDNN_STATUS_SUCCESS);
-        }
-        EXPECT_EQ(hipdnnBackendDestroyDescriptor(configs[i]), HIPDNN_STATUS_SUCCESS);
+        EXPECT_EQ(hipdnnBackendDestroyDescriptor(config), HIPDNN_STATUS_SUCCESS);
     }
```
BrianHarrisonAMD: It's glorious.

backend/src/descriptors/engine_descriptor.hpp
```diff
@@ -45,12 +45,16 @@ class Engine_descriptor : public hipdnnBackendDescriptor
...
-                       void* array_of_elements) override;
+                       void* array_of_elements) const override;
 
     void set_attribute(hipdnnBackendAttributeName_t attribute_name,
                        hipdnnBackendAttributeType_t attribute_type,
                        int64_t element_count,
                        const void* array_of_elements) override;
+
+    // These getters throw an exception if the descriptor is not finalized.
+    std::shared_ptr<const Graph_descriptor> get_graph() const;
```
adickin-amd, line 56: wondering if renaming to get_finalized_XXXX() makes sense for these getters.  Then its clear reading any code using them that the item retrieved is indeed finalized.

backend/src/descriptors/engine_heuristic_descriptor.cpp
```diff
@@ -223,28 +219,42 @@ void Engine_heuristic_descriptor::get_engine_configs(hipdnnBackendAttributeType_
...
-                          HIPDNN_STATUS_BAD_PARAM_NULL_POINTER,
-                          "Engine_heuristic_descriptor failed to get engine config: Config "
-                          "descriptor is null.");
-
-            // TODO - Likely leaking returned descriptors unless caller gets and frees Engine_descriptor inside the engine_config_descriptor when destroying.
-            //        Need to add tracking, and cleaning up of these descriptors in a more user friendly way.
-            auto engine = new Engine_descriptor();
+            auto backend_wrapper = output_array[i];
+
+            auto config = unpack_descriptor<Engine_config_descriptor>(
```
adickin-amd, line 228: why not use your nice convert function, `auto config = backend_wrapper->as_descriptor<Engine_config_descriptor>();`?

backend/src/hipdnn_backend.cpp
```diff
@@ -206,11 +214,11 @@ HIPDNN_BACKEND_EXPORT hipdnnStatus_t
     return hipdnn_backend::try_catch([&, api_name = __func__]() {
         throw_if_invalid_descriptor(descriptor);
 
-        descriptor->get_attribute(attribute_name,
-                                  attribute_type,
-                                  requested_element_count,
-                                  element_count,
-                                  array_of_elements);
+        descriptor->private_descriptor->get_attribute(attribute_name,
```
adickin-amd, line 216: probably not worth it but you could breakout virtual functions on `hipdnnPrivateBackendDescriptor` into an interface like Backend_descriptor_interface and have that on both the private and non private descriptors.  Then you can just call the functions on the descriptor like the original implementation does.  

I can only see this being super valuable if it makes testing other things easier in the future.

backend/src/descriptors/engine_heuristic_descriptor.cpp
BrianHarrisonAMD: Because these used to throw specific status', and I wanted to keep that behavior. 

The convert function doesn't let you provide the status & error.

backend/src/hipdnn_backend.cpp
BrianHarrisonAMD: Ah, and just forwards it.

BrianHarrisonAMD: Ill see how it looks with that swap.
I think it might reduce some of the changes.

Some of the testing logic has to still change, but maybe I can skip a few extra unpacks etc. 

If we fix the descriptors to consistently require passing in a descriptor to get it out, then I think a lot of the awkwardness goes away luckily.

backend/src/descriptors/engine_heuristic_descriptor.cpp
BrianHarrisonAMD: It's basically a trade-off between more explicit errors and status fails, or easier to use.

backend/src/descriptors/engine_descriptor.hpp
BrianHarrisonAMD: I can change that, but I also think these might disappear once we integrate the PluginManager from @averinevg's changes. 

I think these are only used for the awkward finalization stuff.

BrianHarrisonAMD: Ill rename it anyways, and I can remove alter once we do integration if they are unused.

backend/src/descriptors/engine_heuristic_descriptor.cpp
BrianHarrisonAMD: I think I only used the utility method for tests because of this.

backend/src/descriptors/engine_descriptor.hpp
adickin-amd: ah, if they are possibly going away I wouldnt worry about it too much.

backend/src/descriptors/engine_heuristic_descriptor.cpp
```diff
@@ -223,28 +220,42 @@ void Engine_heuristic_descriptor::get_engine_configs(hipdnnBackendAttributeType_
...
+                backend_wrapper,
```
adickin-amd: could you just pass output_array[i] rather than making a temp?

backend/tests/descriptors/mocks/mock_descriptor.hpp
```diff
@@ -35,12 +35,29 @@ class Mock_descriptor : public hipdnnBackendDescriptor
...
 ACTION_P(SetArg4ToInt64, value) // NOLINT
 {
     *static_cast<int64_t*>(arg4) = value;
 }
 
+inline std::unique_ptr<hipdnnBackendDescriptor>
+    make_mock_descriptor_wrapper(hipdnnBackendDescriptorType_t desc_type = HIPDNN_INVALID_TYPE)
+{
+    auto mock_wrapper = std::make_unique<hipdnnBackendDescriptor>();
+    mock_wrapper->private_descriptor = std::make_shared<Mock_descriptor>(desc_type);
```
adickin-amd, line 50: yea if you did the forwarding idea i described above you could probably skip having to make the mock a private descriptor and just have a MockHipdnnBackendDescriptor like before I think.  At least in the 30 seconds I thought about this in my head it seems doable.

adickin-amd: oooh, but then it probably wouldnt work when you use it in the tests since it wont have a private_descriptor present to unpack.

BrianHarrisonAMD: Yea it doesn't. 
You can do the forwarding to avoid several places that you need to call into private_descriptor, but it fundamentally still changes a lot of things since we need to set the wrappers instead of the object directly.

backend/src/descriptors/engine_heuristic_descriptor.cpp
BrianHarrisonAMD: Yea absolutely.

I think when I first converted it I didn't realize we were passing in the already created descriptor for this API, since the other descriptors were unset. The wrapper was needed back then, but now it's not needed at all.

backend/src/descriptors/engine_descriptor.hpp
BrianHarrisonAMD: Yea ill just leave it. I prefer not having these public if we can avoid it, and I think we can once we do the plugin integration.

backend/src/descriptors/descriptor_factory.cpp
```diff
@@ -51,6 +52,9 @@ void Descriptor_factory::create(hipdnnBackendDescriptorType_t descriptor_type,
                                    + " is not supported.");
     }
 
+    *descriptor = new hipdnnBackendDescriptor();
+    (*descriptor)->private_descriptor = private_desc;
```
BrianHarrisonAMD: Probably can use pack static here.

backend/src/descriptors/backend_descriptor.hpp
BrianHarrisonAMD, line 1: Going to create an interface for the descriptor, and implement on public / private.
The public one will forward to the private one.
Ill also add a valid check virtual that throws if the type isn't set, or if the private_desc is nullptr to help reduce needing to access the private_desc. 

If I get to the point where I can hide all the private_desc access, then ill make it fully private, but not sure if that's possible yet.

backend/tests/descriptors/test_variant_pack.cpp
BrianHarrisonAMD, line 1: Whoops these need to be flipped and fixed.

backend/src/descriptors/backend_descriptor.hpp
BrianHarrisonAMD: Going to do this as a separate follow-up.

adickin-amd (approved):

PR #74, opened 2025-07-22
Move callback implementation to private library

SamuelReeder:
## Proposed changes

Having the implementation in the shared library caused linker errors sometimes. This moves the implementation so that it can be used within the private library safely.

## Checklist

### DIFF
--- a/backend/src/hipdnn_backend.cpp
+++ b/backend/src/hipdnn_backend.cpp
@@ -13,7 +13,6 @@
 #include "logging/logging.hpp"
 #include "plugin/plugin_manager.hpp"
 #include <hipdnn_sdk/logging/callback_types.h>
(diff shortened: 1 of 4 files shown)


SamuelReeder:
Not sure why CI is failing. Everything is fine locally but will look into it. I wonder if math CI was updated, since it is flagging things that should have been an issue before too.

adickin-amd:
> Not sure why CI is failing. Everything is fine locally but will look into it. I wonder if math CI was updated, since it is flagging things that should have been an issue before too.

Likely cause is CI is running clang 20 and we are still on clang 19

SamuelReeder:
I can log a ticket to update our repo to clang 20 in that case.

BrianHarrisonAMD (approved):
LGTM

PR #75, opened 2025-07-22
[ALMIOPEN-76]: Part 1:  Integrate frontend graph API with backend API

adickin-amd:
…e graph to the backend library from the frontend.

## Proposed changes

Please describe the motivation behind the pull request, whether it enables a new feature or fixes a bug. If there are associated pull requests or issues, please link them to the pull request.

## Checklist


frontend/include/hipdnn_frontend/backend/backend_wrapper.hpp
```diff
@@ -0,0 +1,153 @@
...
+        backend_create_and_deserialize_graph_ext(hipdnnBackendDescriptor_t* descriptor,
+                                                 const uint8_t* serialized_graph,
+                                                 size_t graph_byte_size)
+        = 0;
+    virtual void logging_callback_ext(hipdnnSeverity_t severity, const char* msg) = 0;
+};
+
+class Hipdnn_backend_wrapper : public Hipdnn_backend_interface
+{
+public:
```
BrianHarrisonAMD, line 13: Doesn't really matter right now, but we probably want to split this into a separate header eventually.

frontend/include/hipdnn_frontend/backend/hipdnn_backend_descriptor.hpp
```diff
@@ -0,0 +1,81 @@
...
+        : _descriptor(nullptr)
+        , _valid(false)
+    {
+    }
+
+    Hipdnn_backend_descriptor(hipdnnBackendDescriptor_t descriptor)
+        : _descriptor(descriptor)
+        , _valid(false)
+    {
+    }
```
BrianHarrisonAMD: Don't we want this to be _valid(true)?

This will leak if the intent was to make this a scoped descriptor, and if that's not the intent, then I don't think we should have this constructor hehe.

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -16,21 +18,112 @@ namespace hipdnn_frontend
 {
 namespace graph
 {
+//I wonder if I should instead make this not take in an error message and instead, grab
+//the error message from the backend and use that.
+#define RETURN_ON_FAILURE(backend_status, error_message)                   \
+    if((backend_status) != HIPDNN_STATUS_SUCCESS)                          \
+    {                                                                      \
+        return error_t(error_code_t::HIPDNN_BACKEND_ERROR, error_message); \
+    }
```
BrianHarrisonAMD: Yea I think we should have it grab the message from the backend instead here.
Could maybe have both versions if you wanted to have a custom error message instead of the backend.

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -16,21 +18,112 @@ namespace hipdnn_frontend
...
+                                                   1,
+                                                   _graph_desc->get()),
+            "Failed to set operation graph on the engine heuristic descriptor.");
+
+        RETURN_ON_FAILURE(hipdnn_backend().backend_set_attribute(_engine_heuristic_desc->get(),
+                                                                 HIPDNN_ATTR_ENGINEHEUR_MODE,
+                                                                 HIPDNN_TYPE_HEUR_MODE,
+                                                                 1,
+                                                                 mode.data()),
+                          "Failed to set mode on the engine heuristic descriptor.");
```
BrianHarrisonAMD: Hmm I think we do need to support N modes, but I think it was written to only support one for now. 
Ill add a note for the future work to support different heuristic modes.

frontend/include/hipdnn_frontend/backend/backend_wrapper.hpp
adickin-amd: ah yes, that should be done, ill split the interface and wrapper up.

frontend/include/hipdnn_frontend/backend/hipdnn_backend_descriptor.hpp
adickin-amd: ah yes, I made a mistake on that one. will fix.

frontend/include/hipdnn_frontend/graph.hpp
BrianHarrisonAMD: Actually come to think of it, I think it only needs to be one, but we need the frontend to run multiple modes, and order back the results.

frontend/include/hipdnn_frontend/graph.hpp
adickin-amd: i was thinking ill take in the error_message as I currently am and append the last error from backend

frontend/include/hipdnn_frontend/backend/hipdnn_backend_descriptor.hpp
adickin-amd: i also forgot tests for this class.

tests/backend/test_backend_execute_api.cpp
```diff
@@ -55,145 +55,145 @@ class Execution_backend_end_api_tests : public ::testing::Test
     }
 };
 
-TEST_F(Execution_backend_end_api_tests, TestBackendExecuteWithNullHandle)
-{
-    Graph graph;
-    graph.set_name("BatchnormInferenceGraph");
-    test_util::create_and_populate_batchnorm_node(graph);
+// TEST_F(Execution_backend_end_api_tests, TestBackendExecuteWithNullHandle)
```
adickin-amd: todo, fix this.

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -16,21 +18,112 @@ namespace hipdnn_frontend
...
+                          "Failed to get engine configurations from the heuristic descriptor.");
+
+        if(count == 0)
+        {
+            return {error_code_t::HIPDNN_BACKEND_ERROR,
+                    "No engine configurations retrieved from the heuristic desc."};
+        }
+
+        _engine_config_desc = std::move(engine_configs[0]);
+        engine_configs.erase(engine_configs.begin(), engine_configs.begin() + 1);
```
BrianHarrisonAMD: It's probably fine to leave it as is, but seems like we could simplify a lot by just trying to grab the first config. 
I know eventually we may want to grab all of them, and apply some filtering to determine the selection, and if we want to keep this that's fine.

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -41,8 +134,10 @@ class Graph : public INode
         return validate_subtree();
     }
 
-    error_t build_operation_graph()
+    error_t build_operation_graph(hipdnnHandle_t handle)
     {
+        std::ignore = handle;
+
```
BrianHarrisonAMD: Shouldn't need this right?

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -80,11 +175,170 @@ class Graph : public INode
...
+                          "Failed to set handle on the graph.");
+
+        RETURN_ON_FAILURE(hipdnn_backend().backend_finalize(_graph_desc->get()),
+                          "Failed to finalize backend descriptor for the graph");
+
+        return {error_code_t::OK, ""};
+    }
+
+    error_t create_execution_plans(hipdnnHandle_t handle, std::vector<HeurMode_t> const& mode)
+    {
```
BrianHarrisonAMD: mode -> modes.

Also, should we provide a default to allow calling without setting it?

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -80,11 +175,170 @@ class Graph : public INode
...
+        if(status.code != error_code_t::OK)
+        {
+            return status;
+        }
+
+        status = initialize_engine_config();
+        if(status.code != error_code_t::OK)
+        {
+            return status;
+        }
```
BrianHarrisonAMD: I thought we had a MACRO to make this type of thing less repetitive.
If we don't, then I think we should add one.

frontend/include/hipdnn_frontend/backend/hipdnn_backend_descriptor.hpp
BrianHarrisonAMD, line 1: We might want to add a utility, or MACRO that does the valid check, and then populates more details from the backend error message if it is failed.

Right now we are replacing the backend error message with a frontend error to describe what happened when it failed, and I think it would be better to have both. Especially in the future when we get these details shared with us hehe.

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -80,11 +175,170 @@ class Graph : public INode
...
+            hipdnn_backend().backend_get_attribute(_execution_plan_desc->get(),
+                                                   HIPDNN_ATTR_EXECUTION_PLAN_WORKSPACE_SIZE,
+                                                   HIPDNN_TYPE_INT64,
+                                                   1,
+                                                   nullptr,
+                                                   &workspace_size),
+            "Failed to get engine configurations from the heuristic descriptor.");
+
+        return {error_code_t::OK, ""};
+    }
```
BrianHarrisonAMD: I think you can also get this from the engine_config.

Is the flow expected that we don't request workspace size until the execution_plan is finalized? 
Actually looking above I guess it doesn't matter since they finalize at the same time hehe.

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -80,11 +175,170 @@ class Graph : public INode
...
+
+    error_t execute(hipdnnHandle_t handle,
+                    std::unordered_map<int64_t, void*>& variant_pack,
+                    void* workspace) const
+    {
+
+        std::ignore = variant_pack;
+        std::ignore = workspace;
```
BrianHarrisonAMD: Can remove I think.

frontend/include/hipdnn_frontend/types.hpp
```diff
@@ -22,6 +22,13 @@ enum class DataType_t
     BFLOAT16 = 3,
 };
 
+enum class HeurMode_t
+{
+    A,
+    B,
+    FALLBACK,
+};
```
BrianHarrisonAMD: We should have just FALLBACK for now.

frontend/tests/fake_backend/fake_hipdnn_backend.hpp
```diff
@@ -0,0 +1,11 @@
...
+Copyright © Advanced Micro Devices, Inc., or its affiliates.
+SPDX-License-Identifier:  MIT
+*/
+#pragma once
+
+#include <memory>
+
+#include "mock_hipdnn_backend.hpp"
+
+void set_mock_hipdnn_backend(std::weak_ptr<Mock_hipdnn_backend> backend);
```
BrianHarrisonAMD, line 13: Should be in a namespace, or static I guess.

(namespace preferred though)

tests/backend/test_backend_execute_api.cpp
BrianHarrisonAMD, line 1: Oh is this because it was using the Graph?

:(

frontend/tests/test_graph.cpp
```diff
@@ -872,4 +941,571 @@ TEST(GraphTests, TensorLikeGraphAttributes)
...
+                         hipdnnBackendAttributeName_t,
+                         hipdnnBackendAttributeType_t,
+                         int64_t,
+                         const void* array_of_elements) {
+            hipdnnHandle_t handle = *static_cast<const hipdnnHandle_t*>(array_of_elements);
+            EXPECT_EQ(handle, this->_handle);
+            return HIPDNN_STATUS_SUCCESS;
+        });
+
+    graph.build_operation_graph(_handle);
```
BrianHarrisonAMD: We should check the status return of this as well right?

frontend/include/hipdnn_frontend/types.hpp
BrianHarrisonAMD: We also need to ensure these values line-up with the backend heuristic values.
Might be best to have a convert function for packing these so we don't mess this up.

frontend/tests/test_graph.cpp
```diff
@@ -872,4 +941,571 @@ TEST(GraphTests, TensorLikeGraphAttributes)
...
+    const std::vector<HeurMode_t> heurModes = {HeurMode_t::A};
+    auto tensor_attributes = create_basic_batchnorm_graph(graph);
+
+    hipdnnBackendDescriptor_t graph_desc = reinterpret_cast<hipdnnBackendDescriptor_t>(0x1234);
+    EXPECT_CALL(*_mock_backend, hipdnnBackendCreateAndDeserializeGraph_ext(_, _, _))
+        .WillOnce([&graph_desc](hipdnnBackendDescriptor_t* descriptor, const uint8_t*, size_t) {
+            *descriptor = graph_desc;
+            return HIPDNN_STATUS_SUCCESS;
+        });
+    graph.build_operation_graph(_handle);
```
BrianHarrisonAMD, line 1001: I think we need to add checks to all calls in this file that the status returned is good.
It'll probably flag from the other expect calls, but we should probably have a function, or Macro wrapping any frontend call.

frontend/tests/test_graph.cpp
BrianHarrisonAMD, line 1: I'm not sure how we can do much about this, but the tests are tightly coupled to the expected API calls from the implementation, and if we change that slightly it will break these. 
I can't think of a way to really improve this, and have tests that give us extra coverage though. 

I think what we really want is to check the final results are doing what we wanted, and not so much the individual order and API calls, but I think you are a bit stuck doing it this way to set-up the mocks.

frontend/tests/test_graph.cpp
adickin-amd: ah yes, missed this one

frontend/tests/test_graph.cpp
adickin-amd: so for this one, i only care about the create_execution_plans at the end.  which also needs its status checked. The amount of expect calls here will blow up this test before any extra expect call here.

tests/backend/test_backend_execute_api.cpp
adickin-amd: yes, its a easy fix, just havent gotten to it.

frontend/include/hipdnn_frontend/backend/hipdnn_backend_descriptor.hpp
adickin-amd: will add error tracking to this

frontend/include/hipdnn_frontend/graph.hpp
adickin-amd: nope, forgot to delete

frontend/include/hipdnn_frontend/graph.hpp
adickin-amd: dont descriptors need to be finalized for me to get attributes from them?

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -80,11 +175,170 @@ class Graph : public INode
...
+    error_t get_workspace_size(int64_t& workspace_size) const
+    {
+        RETURN_ON_FAILURE(
+            hipdnn_backend().backend_get_attribute(_execution_plan_desc->get(),
+                                                   HIPDNN_ATTR_EXECUTION_PLAN_WORKSPACE_SIZE,
+                                                   HIPDNN_TYPE_INT64,
+                                                   1,
+                                                   nullptr,
+                                                   &workspace_size),
+            "Failed to get engine configurations from the heuristic descriptor.");
```
adickin-amd: fix string.

frontend/tests/fake_backend/fake_hipdnn_backend.hpp
adickin-amd: sure i can namespace.  this is a file only linked into the frontend test binary so there really isnt a chance of it clobbering another name later

frontend/tests/test_graph.cpp
adickin-amd: yea there isnt much we can do unless we write our own little framework to test this.  If this becomes something we do often we may want to explore that.

adickin-amd: also by setting up the mocks the expect calls do verify im passing the right data around to each function call.  We could reorder all the expect and on calls to the top of each test, it was just easier to write if I matched the flow 1-1 when I needed to.

BrianHarrisonAMD: I think this is the best we can do, and hopefully only needs to change as we extend graph which shouldn't be all that common.

BrianHarrisonAMD:
Looks good to me.

I think we need to fix the backend testing issues, resolve a couple of those comments, and add a bit more coverage for error case handling to get full coverage, but otherwise I think this is looking good.

Thanks @adickin-amd !

frontend/tests/test_graph.cpp
BrianHarrisonAMD: Well I think it's good to just have it check the returns generally to ensure we don't goofily break that in the future, but I agree only the last one really matters.

frontend/include/hipdnn_frontend/graph.hpp
BrianHarrisonAMD: Yea absolutely.
I meant the engine_config gets finalized first, and you can also request workspace from it. 
if it was possible to have them finalized differently, then it might make sense to use engine_config since it has to be finalized first, but since they are done at the same time this is good.

frontend/include/hipdnn_frontend/graph.hpp
adickin-amd: added comment

frontend/include/hipdnn_frontend/graph.hpp
adickin-amd: fixed, it exists in error.hpp

frontend/include/hipdnn_frontend/types.hpp
adickin-amd: What do you mean?  A function in the frontend that converts a frontend heuristic value -> backend heuristic value?  

I guess this means the mode im currently setting is the wrong one

tests/backend/test_backend_execute_api.cpp
adickin-amd: fixed now

frontend/include/hipdnn_frontend/graph.hpp
adickin-amd: left as is with comment.

frontend/include/hipdnn_frontend/backend/backend_logging_helpers.hpp
```diff
@@ -0,0 +1,63 @@
...
+// SPDX-License-Identifier:  MIT
+
+#pragma once
+
+#include <ostream>
+
+#include <hipdnn_sdk/logging/logger.hpp>
+#include <hipdnn_status.h>
+
+inline const char* to_string(hipdnnStatus_t status)
```
SamuelReeder, line 11: Should we remove this and have all these `to_string` functions in sdk? Since we have another one in `backend/src/error.hpp`.

BrianHarrisonAMD: We can't since hipdnn_status is declared in the backend.

frontend/include/hipdnn_frontend/backend/scoped_hipdnn_backend_descriptor.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+class Scoped_hipdnn_backend_descriptor
+{
+private:
+    hipdnnBackendDescriptor_t _descriptor;
+    bool _valid;
+
+    // For some reason clang isnt picking up the usage of the two variables correctly.
+    // Ive marked them as [[maybe_unused]] to avoid warnings.
+    static void log_backend_error([[maybe_unused]] const std::string& error_string,
```
SamuelReeder, line 23: If no frontend logger is initialized then logging will be a no-op and the variables are unused. The tests or sample should initialize the frontend logger though. I added a utility for this.

frontend/include/hipdnn_frontend/backend/backend_logging_helpers.hpp
BrianHarrisonAMD: Frontend can depend on SDK & Backend.
Backend can depend on SDK.
Plugin can depend on SDK.
SDK can't depend on any other hipDNN component.

SamuelReeder: Right I forgot. It does not make sense to declare status in sdk I assume?

BrianHarrisonAMD: Not really since it's part of the backend API, and you should be able to use backend with just using the C api includes. Since SDK is header only you could technically get away with it, but I rather just duplicate this small method rather than start adding that dep.

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -16,21 +18,125 @@ namespace hipdnn_frontend
...
+    {
+        int64_t available_engine_count = 0;
+        RETURN_ON_BACKEND_FAILURE(
+            hipdnn_backend().backend_get_attribute(_engine_heuristic_desc->get(),
+                                                   HIPDNN_ATTR_ENGINEHEUR_RESULTS,
+                                                   HIPDNN_TYPE_BACKEND_DESCRIPTOR,
+                                                   0,
+                                                   &available_engine_count,
+                                                   nullptr),
+            "Failed to set mode on the engine heuristic descriptor.");
```
SamuelReeder: Should this say "failed to get results"? It's get_attribute call.

SamuelReeder (approved):
LGTM but make sure CI is passing.

frontend/include/hipdnn_frontend/backend/scoped_hipdnn_backend_descriptor.hpp
adickin-amd: yea even with logging enabled an initialized correctly this was still getting flagged. I also had to change the way logging was done for the tests as the callback implementation being async causes all logs to print after all tests have run.  I will bring this up in standup to chat about it

frontend/include/hipdnn_frontend/backend/backend_logging_helpers.hpp
adickin-amd: i do need tests here.

frontend/include/hipdnn_frontend/types.hpp
BrianHarrisonAMD: Yea there is other convert functions for SDK stuff I think.

BrianHarrisonAMD (approved):
👍

PR #76, opened 2025-07-22
[ALMIOPEN-46] Batchnorm infer/train/bwds sample apps

SamuelReeder:
## Proposed changes

Provides 3 new hipDNN samples:

1. Batchnorm inference (fully functional)
2. Batchnorm training (functional, but no engine configs)
3. Batchnorm backwards (functional, but no engine configs)

Each of the samples have their own binary, and an explanation in the README.

## Checklist


samples/README.MD
```diff
@@ -1,9 +1,42 @@
...
+        ```
+        where `y` would then be propogated as input to the subsequent layer.
+*   **`bn_training`** (`batchnorm/bn_training.cpp`): Executes the forward pass of a batch normalization training graph on a 4D input tensor.
+    *   For an input `x` of shape `(N, C, H, W)`, the mean and variance are calculated over the `N`, `H`, and `W` dimensions for each of the `C` channels or mini-batches, resulting in a `batch_mean` and `batch_variance` of shape `(1, C, 1, 1)`. It then transforms the input and updates the running statistics:
+        ```
+        y = gamma * (x - batch_mean) / sqrt(batch_variance + eps) + beta
+        running_mean = (1 - momentum) * running_mean + momentum * batch_mean
+        running_variance = (1 - momentum) * running_variance + momentum * batch_variance
+        ```
+    *   The graph outputs the normalized tensor `y`, along with the mini-batch statistics (`saved_mean`, `saved_inv_variance`) required for the backward pass, and the updated population statistics (`running_mean`, `running_variance`) required for inference.
```
SamuelReeder: [The schema](https://github.com/ROCm/hipDNN/blob/develop/sdk/schemas/batchnorm_attributes.fbs#L17) says its inverse variance, but is it actually referring to inverse standard deviation?

samples/batchnorm/bn_backward.cpp
```diff
@@ -0,0 +1,138 @@
...
+    auto saved_inv_var_tensor
+        = Tensor::make_nchw_tensor<IntermediateType>(saved_inv_variance->get_dim());
+
+    auto dx_tensor = Tensor::make_nchw_tensor<InputType>(dx->get_dim());
+    auto dgamma_tensor = Tensor::make_nchw_tensor<IntermediateType>(dgamma->get_dim());
+    auto dbeta_tensor = Tensor::make_nchw_tensor<IntermediateType>(dbeta->get_dim());
+
+    dy_tensor.template fill_with_random_values<InputType>(static_cast<InputType>(0.0f),
+                                                          static_cast<InputType>(1.0f));
+    dy_tensor.memory().mark_host_modified();
```
mousdahl-amd: You shouldn't need mark_host_modified().  If the device memory's never been accessed, it's assumed that it needs to allocation and copy from the host.

samples/CMakeLists.txt
```diff
@@ -10,14 +10,35 @@ find_package(hipdnn_frontend CONFIG REQUIRED)
...
+add_executable(bn_backward
+    batchnorm/bn_backward.cpp
+)
+target_link_libraries(bn_backward PRIVATE 
+    hip::host
+    hip::device
+    hipdnn_frontend
+    hipdnn_backend
+    hipdnn_sdk
+)
```
mousdahl-amd: If we're going to have a bunch of these, it might make sense to break this into a function.  If we're just going to have a few, probably not worth the effort.

SamuelReeder: That's a good point, I'll do that.

frontend/include/hipdnn_frontend/utilities.hpp
```diff
@@ -70,5 +72,5 @@ inline int32_t initialize_frontend_logging(hipdnnCallback_t fn)
 
     return 0;
 }
-}
```
BrianHarrisonAMD, line 88: Why'd you move this out of the namespace?

samples/CMakeLists.txt
```diff
@@ -2,22 +2,25 @@
 # SPDX-License-Identifier:  MIT
 
 cmake_minimum_required(VERSION 3.25.2)
-project(hipdnn_samples VERSION 0.0.1 LANGUAGES CXX)
+project(hipdnn_samples VERSION 0.1.0 LANGUAGES CXX)
```
BrianHarrisonAMD, line 5: hehe

frontend/include/hipdnn_frontend/utilities.hpp
SamuelReeder: I thought `hipdnn_frontend::` more suitable than `hipdnn_frontend::graph::`.

samples/batchnorm/bn_backward.cpp
```diff
@@ -0,0 +1,133 @@
...
+    auto x = create_tensor({16, 16, 16, 16}, input_type);
+    x->set_uid(uid++);
+
+    auto gamma = create_tensor({1, 16, 1, 1}, intermediate_type);
+    gamma->set_uid(uid++);
+
+    auto saved_mean = create_tensor({1, 16, 1, 1}, intermediate_type);
+    saved_mean->set_uid(uid++);
+
+    auto saved_inv_variance = create_tensor({1, 16, 1, 1}, intermediate_type);
```
BrianHarrisonAMD: Might be nice to replace these with NHWC variables to help show how they relate for readability, but fine as is as well.

samples/batchnorm/bn_backward.cpp
```diff
@@ -0,0 +1,133 @@
...
+    saved_inv_variance->set_uid(uid++);
+
+    auto bn_bwd_attributes = graph::Batchnorm_backward_attributes();
+    bn_bwd_attributes.set_saved_mean_and_inv_variance(saved_mean, saved_inv_variance);
+
+    auto [dx, dgamma, dbeta] = graph->batchnorm_backward(dy, x, gamma, bn_bwd_attributes);
+
+    dx->set_output(true).set_uid(uid++);
+    dgamma->set_output(true).set_uid(uid++);
+    dbeta->set_output(true).set_uid(uid++);
```
BrianHarrisonAMD: You should be able to skip set_uid, and probably worth removing from the samples.
It will automagically manage uid's for you.

samples/utils/helpers.hpp
```diff
@@ -0,0 +1,73 @@
...
+            exit(EXIT_FAILURE);                                                                \
+        }                                                                                      \
+    } while(0)
+
+#define HIPDNN_CHECK(status)                                                             \
+    do                                                                                   \
+    {                                                                                    \
+        if(status != HIPDNN_STATUS_SUCCESS)                                              \
+        {                                                                                \
+            std::cerr << "MIOpen Error: " << hipdnnGetErrorString(status) << " in file " \
```
SamuelReeder: Oops forgot to change this.

samples/batchnorm/bn_backward.cpp
```diff
@@ -0,0 +1,133 @@
...
+    auto dy_tensor = Tensor::make_nchw_tensor<InputType>(dy->get_dim());
+    auto x_tensor = Tensor::make_nchw_tensor<InputType>(x->get_dim());
+    auto gamma_tensor = Tensor::make_nchw_tensor<IntermediateType>(gamma->get_dim());
+    auto saved_mean_tensor = Tensor::make_nchw_tensor<IntermediateType>(saved_mean->get_dim());
+    auto saved_inv_var_tensor
+        = Tensor::make_nchw_tensor<IntermediateType>(saved_inv_variance->get_dim());
+
+    auto dx_tensor = Tensor::make_nchw_tensor<InputType>(dx->get_dim());
+    auto dgamma_tensor = Tensor::make_nchw_tensor<IntermediateType>(dgamma->get_dim());
+    auto dbeta_tensor = Tensor::make_nchw_tensor<IntermediateType>(dbeta->get_dim());
```
BrianHarrisonAMD: Probably fine as is, but could also add a template option for swapping tensor layouts since it should support either.
If we don't want to add that change, then could leave a comment, or note in the readme.

samples/batchnorm/bn_backward.cpp
SamuelReeder: Oh, I didn't know NHWC would work.

samples/batchnorm/bn_backward.cpp
```diff
@@ -0,0 +1,133 @@
...
+    std::unordered_map<int64_t, void*> variant_pack;
+    variant_pack[dy->get_uid()] = dy_tensor.memory().template device_data<void>();
+    variant_pack[x->get_uid()] = x_tensor.memory().template device_data<void>();
+    variant_pack[gamma->get_uid()] = gamma_tensor.memory().template device_data<void>();
+    variant_pack[saved_mean->get_uid()] = saved_mean_tensor.memory().template device_data<void>();
+    variant_pack[saved_inv_variance->get_uid()]
+        = saved_inv_var_tensor.memory().template device_data<void>();
+    variant_pack[dx->get_uid()] = dx_tensor.memory().template device_data<void>();
+    variant_pack[dgamma->get_uid()] = dgamma_tensor.memory().template device_data<void>();
+    variant_pack[dbeta->get_uid()] = dbeta_tensor.memory().template device_data<void>();
```
BrianHarrisonAMD: I think we have a clean-up task for this already, but it would be nice to have a better way to grab these buffer pointers. If it exists, can you add a note to clean-up the samples as well?

samples/batchnorm/bn_backward.cpp
SamuelReeder: I could do all this with my NHWC ticket, since I didn't think any NHWC tensors would work yet.

samples/batchnorm/bn_backward.cpp
```diff
@@ -0,0 +1,133 @@
...
+    dx_tensor.memory().mark_device_modified();
+    auto dx_host_ptr = dx_tensor.memory().template host_data<InputType>();
+    std::cout << "First 10 dx values: ";
+    for(int i = 0; i < 10; ++i)
+    {
+        std::cout << static_cast<float>(dx_host_ptr[i]) << " ";
+    }
+
+    std::cout << "\nBatch normalization backward graph execution complete for " << input_type
+              << ".\n\n";
```
BrianHarrisonAMD: Would be nice to add an optional verification step as well just to demonstrate it.
(Might have to leave just stubs in for the implementations that we are adding shortly).

BrianHarrisonAMD (approved):
Looks good to me!

Some small suggestions on ways to improve, but I think this also is a good start as is.

samples/batchnorm/bn_backward.cpp
BrianHarrisonAMD: Whoops sorry I mean replace these dims with NCHW, and specify the values above.
That way people can easily see that the C value is what is used for the 1D tensors.

Like:
```
int64_t N = 16;
int64_t C = 16;
int64_t H = 16;
int64_t W = 16;

auto x = create_tensor{{N, C, H, W}, input_type);
...
```

samples/batchnorm/bn_backward.cpp
SamuelReeder: I mean, wouldn't hurt to template it now though!

samples/README.MD
BrianHarrisonAMD: Should be matching with BN training from MIOpen:

https://github.com/ROCm/MIOpen/blob/develop_deprecated/include/miopen/miopen.h#L2733-L2788

samples/batchnorm/bn_backward.cpp
SamuelReeder: Ah yes, my (16, 16, 16, 16) tensors with no explanations are definitely a little unclear.

samples/README.MD
BrianHarrisonAMD: But yea it is actually [inverse std](https://github.com/ROCm/MIOpen/blob/develop_deprecated/src/kernels/MIOpenBatchNormActivFwdTrainSpatial.cl#L143).

samples/batchnorm/bn_backward.cpp
mousdahl-amd: I've got a note in my Migratable_memory story to do that.  Well, at the very least, try to make it so we don't need the weird template syntax.  That creeps me out.

samples/README.MD
SamuelReeder: Ah thanks, AI said it was called inverse variance purely by convention, but a bit weird still.

mousdahl-amd (approved):

PR #77, opened 2025-07-22
Engine plugin resource manager part 1

averinevg:
## Proposed changes

List of changes:

- Added engine plugin resource manager
- The test plugins were refactored

## Checklist


backend/src/descriptors/graph_descriptor.cpp
```diff
@@ -78,6 +78,12 @@ void Graph_descriptor::set_attribute(hipdnnBackendAttributeName_t attribute_name
             std::string("Graph_descriptor::set_attribute() is not supported for attribute ")
                 + hipdnn_backend::hipdnn_get_attribute_name_string(attribute_name) + ".");
     }
+
+    if(attribute_name != HIPDNN_ATTR_OPERATIONGRAPH_HANDLE)
+    {
+        // Clear the serialized graph when the graph is modified
+        _serialized_graph.clear();
+    }
```
BrianHarrisonAMD, line 86: Is this needed?

Seems like it would throw before hitting here.
I guess maybe just a future protection thing?

averinevg: Yes, exactly

backend/src/descriptors/graph_descriptor.cpp
```diff
@@ -89,7 +95,32 @@ void Graph_descriptor::deserialize_graph(const uint8_t* serialized_graph, size_t
                   HIPDNN_STATUS_BAD_PARAM,
                   "Graph_descriptor::deserialize_graph: graph_byte_size is 0");
 
+    // TODO: Consider lazy graph deserialization
     flatbuffer_utilities::convert_serialized_graph_to_graph(
         serialized_graph, graph_byte_size, _graph);
```
BrianHarrisonAMD, line 100: I think we probably don't want to deserialize this in the backend at all.
This will fail if a new schema update is introduced that the backend wasn't compiled against, and I am not sure we need it generally. 

Might require some thinking on this for how we handle the verification check though, but I think it might be fine to just let the plugins deal with it, and if no plugin understands the graph it's a simple "no solution found" type result.

backend/src/plugin/engine_plugin_manager.cpp
```diff
@@ -0,0 +1,566 @@
...
+    {
+        try
+        {
+            plugin->destroy_handle(handle);
+        }
+        catch(const Hipdnn_exception& e)
+        {
+            HIPDNN_LOG_ERROR(e.get_message());
+        }
+    }
```
BrianHarrisonAMD: Are these the plugin handles that are used for calling execute, and wrapping access to the device?
If so, I think they need to be associated with hipDNNHandles that get created & destroyed.

For example with MIOpen, I would expect each hipDNNHandle to have a corresponding MIOpen handle created inside the plugin.

BrianHarrisonAMD: Yea okay below I can see it for SetStream.
I think this needs to be N handles per plugin, and created when hipDNN handles are created.
Can maybe lazily create the plugin handles when needed, but either way the hipDNN handle needs to map to a handle per plugin.

backend/src/plugin/engine_plugin_manager.cpp
```diff
@@ -0,0 +1,566 @@
...
+{
+    const auto& serialized_graph = graph_desc->get_serialized_graph();
+    const hipdnnPluginConstData_t serialized_graph_data{serialized_graph.data(),
+                                                        serialized_graph.size()};
+
+    std::vector<int64_t> engine_ids;
+
+    for(const auto& [handle, plugin] : _handle_to_plugin)
+    {
+        auto ids = plugin->get_applicable_engine_ids(handle, &serialized_graph_data);
```
BrianHarrisonAMD: Graph_descriptor has handle now, and we will need to use that handle for this API call.

backend/src/plugin/engine_plugin_manager.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+    void execute_op_graph(int64_t engine_id,
+                          hipdnnEnginePluginExecutionContext_t execution_context,
+                          void* workspace,
+                          const hipdnnPluginDeviceBuffer_t* device_buffers,
+                          uint32_t num_device_buffers) const;
+
+    std::shared_ptr<Root_engine_plugin_manager> _root_pm;
+    std::unordered_map<hipdnnEnginePluginHandle_t, const Engine_plugin*> _handle_to_plugin;
+    mutable std::unordered_map<int64_t, hipdnnEnginePluginHandle_t> _engine_id_to_handle;
```
BrianHarrisonAMD, line 93: Probably needs to be `std::unordered_map<int64_t, std::set<hipdnnEnginePluginHandle_t> _engine_id_to_handles`.

BrianHarrisonAMD: I think these need to be something like:
```
std::unordered_map<hipdnnHandle_t, std::unordered_map<int64_t, hipdnnEnginePluginHandle_t>> _handle_to_plugin_handles;
```

Might be able to make something a bit nicer to work with, but I think we need the key to be the hipdnnHandle_t & int64_t to get the associated plugin handle.

backend/src/plugin/engine_plugin_manager.cpp
```diff
@@ -0,0 +1,566 @@
...
+            {
+                throw Hipdnn_exception(HIPDNN_STATUS_PLUGIN_ERROR,
+                                       "Engine ID " + std::to_string(id)
+                                           + " is already associated with a different plugin");
+            }
+            _engine_id_to_handle[id] = handle;
+        }
+    }
+
+    return engine_ids;
```
BrianHarrisonAMD: Once we swap the way handle association works, I think this changes a bit. 

We will still want some check to ensure we don't get duplicat engineIds from the plugins, but I don't think we need to have the handle lookup / check anymore.

backend/src/plugin/engine_plugin_manager.cpp
BrianHarrisonAMD: I think generally all the places we are getting handle right now will need to change to getting the hipdnnHandle from the hipdnnBackendDescriptor's, and then grabbing the associated plugin handle to use.

backend/src/plugin/engine_plugin_manager.cpp
```diff
@@ -0,0 +1,566 @@
...
+    auto engine_ids = get_applicable_engine_ids(graph_desc);
+    if(std::ranges::find(engine_ids, engine_id) == engine_ids.end())
+    {
+        throw Hipdnn_exception(HIPDNN_STATUS_BAD_PARAM,
+                               "Engine ID " + std::to_string(engine_id)
+                                   + " is not in a valid range of engine IDs");
+    }
+
+    // TODO: Get engine details
+    // This will be implemented at the integration stage
```
BrianHarrisonAMD: Okay nice, my changes should make this quick to implement.

backend/src/plugin/engine_plugin_manager.cpp
```diff
@@ -0,0 +1,566 @@
...
+                                                        serialized_graph.size()};
+
+    auto handle = _engine_id_to_handle.at(engine_id);
+    auto plugin = _handle_to_plugin.at(handle);
+
+    return plugin->get_workspace_size(handle, engine_config, &serialized_graph_data);
+}
+
+// TODO: Pack engine_config
+// TODO: Get engine_id from engine_config
```
BrianHarrisonAMD: I think the engine_id is fine to be passed in. 
The hipdnnBackendDescriptor should have it available.

backend/src/plugin/engine_plugin_manager.cpp
```diff
@@ -0,0 +1,566 @@
...
+    plugin->execute_op_graph(
+        handle, execution_context, workspace, device_buffers, num_device_buffers);
+}
+
+void Engine_plugin_manager::finalize_engine(hipdnnBackendDescriptor_t desc) const
+{
+    assert(desc->type == HIPDNN_BACKEND_ENGINE_DESCRIPTOR);
+    auto engine_desc = static_cast<Engine_descriptor*>(desc);
+
+    engine_desc->finalize();
```
BrianHarrisonAMD: We can rework, or decide this when we integrate further, but currently I think it's expected the descriptor is finalized already when it gets passed in, and the plugin manager sets the additional state.

backend/src/plugin/engine_plugin_manager.hpp
```diff
@@ -0,0 +1,150 @@
...
+    Engine_execution_context_wrapper(Engine_execution_context_wrapper&& other) noexcept;
+    Engine_execution_context_wrapper& operator=(Engine_execution_context_wrapper&& other) noexcept;
+
+    hipdnnEnginePluginExecutionContext_t get() const;
+
+private:
+    std::shared_ptr<Engine_plugin_manager> _pm;
+    int64_t _engine_id;
+    hipdnnEnginePluginExecutionContext_t _execution_context;
+};
```
BrianHarrisonAMD: What are the wrappers for?

Edit: Come to think of it, I guess we need to figure out how to propagate the destruction of these things similar to how the backend descriptors are working right now. 

If I understand correctly, we need to associate these with the backend descriptors, and delete them inside the plugin as well once they are no longer referenced by anything. Actually with the changes I am doing, we can probably just store this context_wrapper inside the associated backend_descriptor, and they will both get cleaned up properly when they are no longer referenced by a user, or an associated object that's still alive. We probably just need to set these on the associated backend descriptors during the finalize call, and that should work.

backend/src/plugin/engine_plugin_manager.cpp
```diff
@@ -0,0 +1,566 @@
...
+class Root_engine_plugin_manager : public Plugin_manager_base<Engine_plugin>
+{
+};
+
+namespace
+{
+
+std::mutex plugin_mutex;
+std::vector<std::filesystem::path> override_plugin_paths;
+std::weak_ptr<Root_engine_plugin_manager> root_pm_ptr;
```
BrianHarrisonAMD: What's the intent for the weak_ptr? 
Is it just to make sure we never have more than one active at a time, but still allow for unloading for testing purposes?

mousdahl-amd: This confused me a little bit too.  Maybe drop a comment here as to the "why".  It took me a while to understand that Engine_plugin_manager wasn't derived from Plugin_manager_base<Engine_plugin>, which is what I would've assumed given the class names

mousdahl-amd (commented):
Still too new to comment as to architecture, looks great as far as I can see.

backend/src/plugin/engine_plugin_manager.hpp
```diff
@@ -0,0 +1,150 @@
...
+    size_t get_workspace_size(int64_t engine_id,
+                              const hipdnnPluginConstData_t* engine_config,
+                              Graph_descriptor* graph_desc) const;
+
+    hipdnnEnginePluginExecutionContext_t
+        create_execution_context(int64_t engine_id,
+                                 const hipdnnPluginConstData_t* engine_config,
+                                 Graph_descriptor* graph_desc) const;
+    void destroy_execution_context(int64_t engine_id,
+                                   hipdnnEnginePluginExecutionContext_t execution_context) const;
```
adickin-amd: whats the reasoning behind making these private and allowing access via friend classes rather than just having them be public?

backend/src/descriptors/graph_descriptor.cpp
BrianHarrisonAMD: Ill add something to track this separately.

backend/src/plugin/engine_plugin_manager.cpp
```diff
@@ -0,0 +1,580 @@
...
+
+std::vector<std::filesystem::path> get_default_plugin_paths()
+{
+    // This function should return the default plugin paths.
+    // For now, we return an empty vector.
+    // TODO: Implement logic to retrieve default plugin paths.
+    return {};
+}
```
BrianHarrisonAMD: Could be good for someone to take this on as a separate piece for the integration work.
We need some logic for how the default plugins will work, and then optional overrides + a way to set a specific plugin path to use for loading through _ext API.

@SamuelReeder or @bghimireamd

backend/src/plugin/engine_plugin_manager.cpp
```diff
@@ -0,0 +1,580 @@
...
+class Root_engine_plugin_manager : public Plugin_manager_base<Engine_plugin>
+{
+};
+
+namespace
+{
+
+std::mutex plugin_mutex;
+std::vector<std::filesystem::path> override_plugin_paths;
+std::weak_ptr<Root_engine_plugin_manager> root_pm_ptr;
```
BrianHarrisonAMD: I think we should try and figure out a naming change to make it more clear what is happening here.

Root_engine_plugin_manager is Engine_plugin_manager. 

Engine_plugin_manager is Plugin_handle_context?
Not sure on what a good name is, but Engine_plugin_manager owns the plugin handles, and provides access to the associated plugin operations.

backend/src/plugin/engine_plugin_manager.cpp
```diff
@@ -0,0 +1,580 @@
...
+        {
+            auto it = _engine_id_to_handle.find(id);
+            if(it != _engine_id_to_handle.end() && it->second != handle)
+            {
+                throw Hipdnn_exception(HIPDNN_STATUS_PLUGIN_ERROR,
+                                       "Engine ID " + std::to_string(id)
+                                           + " is already associated with a different plugin");
+            }
+            _engine_id_to_handle[id] = handle;
+        }
```
BrianHarrisonAMD: I think _engine_id_to_handle should be populated from the available engineIds inside a plugin.
If we don't have it already, we should add a way to request the _engineIds that exist within a plugin without providing a graph, and populate it either lazily on request, or as part of initialization. 

Right now it just partially populates it, and if its used before calling get_applicable_engine_ids it won't necessarily be correct. It also seems like we should be throwing right away if there is overlapping engineIds.

backend/src/plugin/engine_plugin_manager.cpp
BrianHarrisonAMD: I guess if we add and pull it from engine_config, then it'll ensure there isn't a mismatch through some silly bug, but can leave this for now.

backend/src/plugin/engine_plugin_manager.cpp
```diff
@@ -0,0 +1,580 @@
...
+    engine->get_attribute(
+        HIPDNN_ATTR_ENGINE_OPERATION_GRAPH, HIPDNN_TYPE_BACKEND_DESCRIPTOR, 1, nullptr, &graph);
+    auto graph_desc = static_cast<Graph_descriptor*>(graph);
+
+    // TODO: Move to the engine config descriptor
+    // Now we have only one parameter in the engine config, but we will add more parameters later.
+    flatbuffers::FlatBufferBuilder builder;
+    auto engine_config = hipdnn_sdk::data_objects::CreateEngineConfig(builder, engine_id);
+    builder.Finish(engine_config);
+    hipdnnPluginConstData_t engine_config_data{builder.GetBufferPointer(), builder.GetSize()};
```
BrianHarrisonAMD: I know we had some awkwardness with this in @adickin-amd 's plugin.
Is it expected that the plugin holds onto this, or just has it as a temporary thing? 

I think the memory gets cleaned up once this is out of scope.

backend/src/plugin/engine_plugin_manager.cpp
```diff
@@ -0,0 +1,580 @@
...
+    hipdnnBackendDescriptor_t graph;
+    heur_desc->get_attribute(
+        HIPDNN_ATTR_ENGINEHEUR_OPERATION_GRAPH, HIPDNN_TYPE_BACKEND_DESCRIPTOR, 1, nullptr, &graph);
+    assert(graph != nullptr);
+    auto graph_desc = static_cast<Graph_descriptor*>(graph);
+
+    auto engine_ids = get_applicable_engine_ids(graph_desc);
+    heur_desc->set_engine_ids(engine_ids);
+
+    // TODO: Get engine details
```
BrianHarrisonAMD: Maybe we should just leave these finalizes & execute fully blank for now, and implement them as part of integration. 

They are way different in my branch, and they need to change quite a bit anyways.

backend/tests/test_engine_plugin_manager.cpp
```diff
@@ -0,0 +1,36 @@
...
+    auto plugin_manager = plugin::Engine_plugin_manager::create();
+
+    hipStream_t stream;
+    ASSERT_EQ(hipStreamCreate(&stream), hipSuccess);
+    Scoped_resource stream_res(stream, [](hipStream_t s) { std::ignore = hipStreamDestroy(s); });
+
+    plugin_manager->set_stream(stream);
+
+    // TODO: Implement a test for executing an operation graph
+}
```
BrianHarrisonAMD: I think we need to add more tests with some mock'd implementation, or with a fake plugin.
It would be nice if we could use a mock'd object at this level, and then for the public backend tests we can use the test plugin to test E2E, but we can live with using a test plugin for these as well if it's too hard to separate out.

BrianHarrisonAMD: I think it's fine to add this test coverage as a follow-up though, and not delay this pr.

backend/src/plugin/engine_plugin_manager.cpp
averinevg: I think these functions should be moved to the descriptors.

backend/src/plugin/engine_plugin_manager.cpp
averinevg: I think all of this should be moved into descriptors. It would be more logical if these descriptors finalize themselves when `finalize()` is called.

backend/src/plugin/engine_plugin_manager.hpp
averinevg: These functions aren't meant to be public. I believe only what's necessary should be public, and everything else should be hidden.

backend/src/plugin/engine_plugin_manager.cpp
averinevg: In the current design, the plugins (Plugin_base, Engine_plugin, etc.) are dumb objects that don't hold any state. They only provide access to functions. Everything else is handled by the plugin manager. I'll add support for retrieving all engine IDs later.

backend/src/plugin/engine_plugin_manager.cpp
averinevg: I think plugins should treat `engine_config` as a temporary object.

backend/src/plugin/engine_plugin_manager.cpp
BrianHarrisonAMD: Sure can leave it as a future change.

backend/src/plugin/engine_plugin_manager.cpp
BrianHarrisonAMD: Yea before it didn't make sense, and it needed to be through the plugin manager.
If we add the plugin context (or w/e), to the hipdnnHandle, then we should be able to finalize inside the descriptors, and the finalize methods declared inside engine_plugin_manager.hpp can go away as well. 

We should just remove that extra finalize logic inside these files for now, and plan to move it into descriptors next.

backend/src/plugin/engine_plugin_manager.cpp
BrianHarrisonAMD: Yea I think you are right.
It should be read-only state that is used during the creation of execution context, but the plugin shouldn't need to hold onto it.

backend/src/plugin/engine_plugin_manager.cpp
SamuelReeder: Yeah I'll take this.

backend/src/plugin/engine_plugin_manager.cpp
SamuelReeder: If the `_engine_id_to_handle` map is only supposed to be engine IDs applicable to the current graph--should it be renamed with an "applicable" prefix or something? Especially if we have a mapping of all engine IDs later.

backend/tests/test_engine_plugin_manager.cpp
SamuelReeder, line 1: Definitely more breadth of testing to add here in addition to executing a graph with one loaded plugin. We need to cover edge cases and more realistic cases with multiple plugins with some applicable and some inapplicable engines. I and/or @bghimireamd will do this afterwards.

backend/src/plugin/engine_plugin_manager.cpp
adickin-amd: Its fine here.  neither the hipdnnPluginConstData_t  or flatbuffers::FlatBufferBuilder are kept after this function ends so no leaks here.

BrianHarrisonAMD (approved):
LGTM

averinevg:
The tests passed locally.

PR #79, opened 2025-07-23
Bg/tensor attribute value

bghimireamd:
## Proposed changes

Update frontend schema & API to handle Tensor_attributes as values

## Checklist

### DIFF
--- a/frontend/include/hipdnn_frontend/attributes/tensor_attributes.hpp
+++ b/frontend/include/hipdnn_frontend/attributes/tensor_attributes.hpp
@@ -3,15 +3,15 @@
 #pragma once
 
 #include "graph_attributes.hpp"
+#include <flatbuffers/flatbuffers.h>
 #include <hipdnn_frontend/types.hpp>
 #include <hipdnn_sdk/data_objects/graph_generated.h>
 #include <hipdnn_sdk/data_objects/tensor_attributes_generated.h>
-#include <string>
-#include <vector>
-#include <variant>
 #include <optional>
+#include <string>
 #include <type_traits>
-#include <flatbuffers/flatbuffers.h>
+#include <variant>
+#include <vector>
 
 namespace hipdnn_frontend
 {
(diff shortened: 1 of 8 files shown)


frontend/include/hipdnn_frontend/attributes/tensor_attributes.hpp
```diff
@@ -15,6 +20,41 @@ namespace graph
 class Tensor_attributes
 {
 public:
+    using ValueVariant = std::variant<std::monostate, float, uint16_t, uint8_t, int32_t>;
```
mousdahl-amd: Do we want to add a follow-up ticket to switch these to half / bfloat16 after I merge?

mousdahl-amd: No double?

BrianHarrisonAMD: Yea we should, and likely add double.

sdk/schemas/tensor_attributes.fbs
```diff
@@ -5,11 +5,35 @@ include "data_types.fbs";
...
+table IValue {
+  value: int;
+}
+
+union Tensor_Value {
+  FValue,
+  HValue,
+  UValue,
+  IValue
+}
```
BrianHarrisonAMD: Can't you have a union of primitive types?

frontend/include/hipdnn_frontend/attributes/tensor_attributes.hpp
```diff
@@ -129,8 +169,50 @@ class Tensor_attributes
     flatbuffers::Offset<hipdnn_sdk::data_objects::TensorAttributes>
         pack_attributes(flatbuffers::FlatBufferBuilder& builder) const
     {
-        return CreateTensorAttributesDirect(
-            builder, _uid, _name.c_str(), to_sdk_type(_data_type), &_stride, &_dim, _is_virtual);
+        // using FB = hipdnn_sdk::data_objects;
+
```
BrianHarrisonAMD: Can remove this I believe.

frontend/tests/test_tensor_value_attributes.cpp
BrianHarrisonAMD, line 1: We should have some tests verifying it fails to retrieve values if you pass in one type, and request another back out. 

For example, if you set a float, and request a uint8 it should fail.

sdk/schemas/tensor_attributes.fbs
```diff
@@ -5,11 +5,35 @@ include "data_types.fbs";
...
+
 table TensorAttributes {
     uid: long;
     name: string;
     data_type: DataType;
     strides: [long];
     dims: [long];
     virtual: bool;
+    value: Tensor_Value;
```
BrianHarrisonAMD, line 38: Not sure why I did this, but the naming of the other tables seems to be different than this.
We should unify the naming.

sdk/schemas/tensor_attributes.fbs
```diff
@@ -5,11 +5,35 @@ include "data_types.fbs";
...
+
+table UValue {
+  value: ubyte;
+}
+
+table IValue {
+  value: int;
+}
+
+union Tensor_Value {
```
BrianHarrisonAMD: Not sure why I did this, but the naming of the other tables seems to be different than these.
We should make the naming for these consistent.

frontend/tests/test_tensor_value_attributes.cpp
BrianHarrisonAMD: Also a test for packing, and unpacking an empty value I guess.

BrianHarrisonAMD (approved):
LGTM

sdk/schemas/tensor_attributes.fbs
BrianHarrisonAMD: I believe you cannot.

PR #80, opened 2025-07-24
[ALMIOPEN-130] - Added CPU Reference Validation

mousdahl-amd:
## Proposed changes

- Added a validator interface, and a CPU floating-point reference implementation
- Added a reference CPU batchnorm forward inference method
- Added test cases for MIOpen legacy plugin to verify batchnorm
- Added helpers for half and bfloat16
- Added a Migratable_memory class that will auto-migrate data between host and device
- Added a Test_tensor class that uses Migratable_memory

## Checklist


plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
```diff
@@ -107,5 +148,36 @@ TEST_F(Batchnorm_execute_graph_test, RunFwdbatchnormGraph)
...
+    Cpu_fp_reference_implementation<float, float, float> cpu_ref_impl;
+    cpu_ref_impl.execute(x_tensor_cpu,
+                         scale_tensor_cpu,
+                         bias_tensor_cpu,
+                         mean_tensor_cpu,
+                         variance_tensor_cpu,
+                         y_tensor_cpu,
+                         1e-3f);
+
+    Cpu_fp_reference_validation<float> cpu_ref_validation(1e-6f, 1e-6f);
```
mousdahl-amd: 1e-6f for absolute tolerances.  Is that good enough?

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
```diff
@@ -0,0 +1,89 @@
...
+        int64_t index = getIndex(tensor, bidx, cidx, row, column);
+        auto* data = tensor.memory().host_data<TUV>();
+        data[index] = value;
+    }
+
+    int64_t getIndex(const Test_tensor& tensor, int64_t bidx, int64_t cidx, int64_t row, int64_t column) const
+    {
+        const auto& strides = tensor.strides();
+        return bidx * strides[0] + cidx * strides[1] + row * strides[2] + column * strides[3];
+    }
```
mousdahl-amd: I was curious if getIndex should be moved to Test_tensor or not.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
```diff
@@ -6,16 +6,36 @@
...
-#include <hipdnn_sdk/utilities/gpu_memory.hpp>
 
 #include "hipdnn_engine_plugin_execution_context.hpp"
 #include "hipdnn_engine_plugin_handle.hpp"
 
-using namespace hipdnn::sdk::utilities;
+using namespace hipdnn_sdk::reference_test_utilities;
 
-class Batchnorm_execute_graph_test : public ::testing::Test
+struct Bn_2d_test_case
```
mousdahl-amd, line 23: Whoops, need to add half / bfloat16 test cases

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
SamuelReeder: For float I think it should be. For less precise types you probably need to scale it up.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
SamuelReeder: For float I think it should be. For less precise types you probably need to scale it up.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include <hipdnn_sdk/test_utilities/reference_implementation_interface.hpp>
+#include <vector>
+#include <numeric>
+
+namespace hipdnn_sdk {
+namespace reference_test_utilities {
+
+using namespace hipdnn_sdk::utilities;
+
+template<class T, class U, class V = U>
```
SamuelReeder: Maybe this is convention but otherwise could make these names more descriptive.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
SamuelReeder: I'm thinking it might make sense for all of `getValue`, `setValue`, and `getIndex` to be moved to test_tensor since they only interact with the tensor.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
```diff
@@ -36,55 +56,76 @@ class Batchnorm_execute_graph_test : public ::testing::Test
...
-    std::vector<int64_t> strides(4);
-    strides[3] = 1; // 1
-    strides[2] = dims[3]; // W
-    strides[1] = dims[2] * dims[3]; // H * W
-    strides[0] = dims[1] * strides[1]; // C * H * W
+template <typename T>
+hipdnnPluginDeviceBuffer_t
+    generate_random_device_buffer(Test_tensor& tensor, int uid, T min, T max, unsigned int seed = 0)
+{
+    tensor.fill_with_random_values<T>(min, max, seed);
```
SamuelReeder, line 92: Can the user easily set the seed?

sdk/include/hipdnn_sdk/test_utilities/reference_implementation_interface.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <hipdnn_sdk/data_objects/tensor_attributes_generated.h>
+
+namespace hipdnn_sdk {
+namespace reference_test_utilities {
+
+using namespace hipdnn_sdk::data_objects;
+using namespace hipdnn_sdk::utilities;
+
+template<class T, class U, class V = U>
+class Reference_implementation_interface {
```
SamuelReeder: This should be renamed to something like `Batchnorm_infer_reference_implementation_interface`.

SamuelReeder: I realize the other reference files/classes need to be renamed too. And/or put in a batchnorm folder.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
BrianHarrisonAMD: I think it's fine when it's one type, but if these have multiple different purposes it would be useful to clarify what they represent.

sdk/include/hipdnn_sdk/test_utilities/test_tensor.hpp
```diff
@@ -0,0 +1,128 @@
...
+        std::vector<int64_t> strides(n, 1);
+        if (n == 0) return strides;
+        
+        // For column-major, we start from the second dimension and multiply by the size of the previous one.
+        for (size_t i = 1; i < n; ++i) {
+            strides[i] = dims[i - 1] * strides[i - 1];
+        }
+        
+        return strides;
+    }
```
BrianHarrisonAMD: There is code in the frontend for calculating strides from a given layout ordering:
- https://github.com/ROCm/hipDNN/blob/develop/frontend/include/hipdnn_frontend/utilities.hpp
- https://github.com/ROCm/hipDNN/blob/develop/frontend/tests/test_utilities.cpp

I think it would make sense to move this to the SDK, and change this to use the same logic.
The convention is also something like NHWC, or NCHW for describing the layout ordering, since tensors are multi-dimensional. 

In the end, I think it would be good to have that stride calculation logic in the SDK, and some extra utilities added to help generate the desired layouts. I think right now only NHWC is implemented.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
```diff
@@ -0,0 +1,221 @@
...
+    explicit Migratable_memory(size_t count = 0, size_t item_size = 0) 
+        : _host_ptr(nullptr), _device_ptr(nullptr), _count(count), _item_size(item_size),
+          _total_size(count * item_size), _current_location(Location::NONE), _host_valid(false), 
+          _device_valid(false) {
+        if (count > 0) {
+            allocate_host();
+        }
+    }
+
+    ~Migratable_memory() {
```
SamuelReeder: I believe this could produce memory leak if it throws before attempting to free device mem. I'm not sure if there's implicit behavior of destructor that prevents this though.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
```diff
@@ -0,0 +1,89 @@
...
+    Cpu_fp_reference_implementation() = default;
+    ~Cpu_fp_reference_implementation() override = default;
+    
+    void execute(const Test_tensor& input,
+        const Test_tensor& scale,
+        const Test_tensor& bias,
+        const Test_tensor& estimatedMean,
+        const Test_tensor& estimatedVariance,
+        Test_tensor& output, 
+        V epsilon) override
```
BrianHarrisonAMD: Might be good to name this execute_batchnorm_inference, or something that makes it clear what algo it's running references for. 

Assuming that the reference_implementation_interface will be extended with all the other references over time.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
```diff
@@ -0,0 +1,89 @@
...
+            throw std::runtime_error("Batchnorm inference requires a 4D tensor.");
+        }
+
+        int64_t n_batches = input.dims().at(0);
+        std::vector<int64_t> channels(static_cast<size_t>(input.dims().at(1)));
+        std::iota(channels.begin(), channels.end(), 0);
+        int64_t height    = input.dims().at(2);
+        int64_t width     = input.dims().at(3);
+
+        std::for_each(channels.begin(), channels.end(), [&](int64_t cidx) {
```
BrianHarrisonAMD, line 45: Can't we make this parallel with:
```
std::for_each(
    std::execution::par,
```

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
```diff
@@ -0,0 +1,221 @@
...
+        _host_valid = false;
+        _device_valid = false;
+    }   
+
+private:
+    void throw_on_error(hipError_t err, const char* msg) {
+        if (err != hipSuccess) {
+            throw std::runtime_error(msg);
+        }
+    }
```
SamuelReeder: A utility header with the conventional `HIP_CHECK` macro may be better.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
BrianHarrisonAMD: Yea I think they should be, and probably renamed to include host in the name.

```
GetHostValue(...)
SetHostValue(...)
GetIndex(...)
```

BrianHarrisonAMD: Also, getIndex should probably be swapped to ```nidx, cidx, hidx, widx```. 

Since the dims and strides are described always in NCHW order.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_validation.hpp
```diff
@@ -0,0 +1,63 @@
...
+public:
+    
+    Cpu_fp_reference_validation() = default;
+    Cpu_fp_reference_validation(T absolute_tolerance, T relative_tolerance)
+        : _absolute_tolerance(absolute_tolerance), _relative_tolerance(relative_tolerance)
+    {
+        if (absolute_tolerance < 0 || relative_tolerance < 0) {
+            throw std::invalid_argument("Tolerances must be non-negative");
+        }
+    }
```
BrianHarrisonAMD: Can also add some constraints to ensure T is numeric type.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_validation.hpp
```diff
@@ -0,0 +1,63 @@
...
+    {
+        if (reference.count() != implementation.count()) {
+            return false;
+        }
+
+        size_t element_count = reference.count();
+        const T* ref_data = reference.host_data<T>();
+        const T* impl_data = implementation.host_data<T>();
+
+        for (size_t i = 0; i < element_count; ++i) {
```
BrianHarrisonAMD: Is this formatted?

I thought it forced newline for braces, but maybe I was misremembering.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_validation.hpp
BrianHarrisonAMD: Hehe actually the numeric limits below might actually give that to you for free.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_validation.hpp
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+        
+        return true;
+    }
+    
+private:
+    // Tolerances for comparison
+    T _absolute_tolerance = std::numeric_limits<T>::epsilon();
+    T _relative_tolerance = std::numeric_limits<T>::epsilon();
```
BrianHarrisonAMD: Might be worth making this default values for the constructor.

sdk/include/hipdnn_sdk/test_utilities/reference_implementation_interface.hpp
BrianHarrisonAMD: I think this is intended to be an interface for all reference impls, but execute should be renamed.

SamuelReeder: Oh yes, just the execute method is for bn infer. So have different execute methods for each type of reference?  Or maybe the implementation could unpack an `input` arg so the header is generalized.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
BrianHarrisonAMD: I think we don't want to throw in destructors generally.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
```diff
@@ -0,0 +1,221 @@
...
+
+    void cleanup() {
+        if (_host_ptr) {
+            throw_on_error(hipHostFree(_host_ptr), "Failed to free host memory");
+            _host_ptr = nullptr;
+        }
+        if (_device_ptr) {
+            throw_on_error(hipFree(_device_ptr), "Failed to free device memory");
+            _device_ptr = nullptr;
+        }
```
BrianHarrisonAMD: I think we need to silently ignore issues in this case.
We don't want to throw in the destructor.

Could maybe use the logging, but not sure it's worth bringing logging into the mix for this since it's not expected to be consumed outside of test apps / testing.

sdk/tests/test_utilities/test_cpu_fp_reference_implementation.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+    Cpu_fp_reference_implementation<float, float, float> ref_impl;
+
+    ref_impl.execute(input_tensor,
+                     scale_tensor,
+                     bias_tensor,
+                     mean_tensor,
+                     variance_tensor,
+                     output_tensor,
+                     1e-5f);
```
BrianHarrisonAMD: If we can, I think it would be worth adding a golden value reference(s) to compare against.

Right now I assume this is just verifying that it doesn't throw during execution?

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_validation.hpp
adickin-amd: yea I think formatting is a little off.  mine always forces the braces to newline.

sdk/include/hipdnn_sdk/test_utilities/test_tensor.hpp
```diff
@@ -0,0 +1,128 @@
...
+#include <vector>
+#include <random>
+
+namespace hipdnn_sdk {
+namespace reference_test_utilities {
+
+using namespace hipdnn_sdk::utilities;
+
+// Wraps vectors of dims/strides and Migratable_memory<T> to provide a common interface for testing
+class Test_tensor
```
adickin-amd: Does this name need to stay as Test_tensor?  This looks like it would be a great utility thats available from the SDK for anyone to utilize. For example, could this be used in all the sample code?

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
adickin-amd: i think logging is a good idea.  Since this is a utility class that could get used by anyone you dont know for sure itll only be used for test apps and testing.

sdk/tests/utilities/test_migratable_memory.cpp
```diff
@@ -0,0 +1,148 @@
...
+    Migratable_memory memory(10, sizeof(float));
+    memory.resize(20);
+
+    EXPECT_FALSE(memory.empty());
+    EXPECT_EQ(memory.count(), 20);
+    EXPECT_EQ(memory.location(), Migratable_memory::Location::HOST);
+    EXPECT_NE(memory.host_data<float>(), nullptr);
+}
+
+TEST(MigratableMemory, MigrateToDevice)
```
adickin-amd, line 98: for any tests that use the device we need to add this macro SKIP_IF_NO_DEVICES();

sdk/tests/CMakeLists.txt
```diff
@@ -34,6 +41,8 @@ target_link_libraries(hipdnn_sdk_tests
     pthread
     hip::host
     hipdnn_sdk
+    hip::host
+    hip::device
```
adickin-amd, line 45: do we need device?  im under the impression that host is enough unless youre compiling a kernel

BrianHarrisonAMD: Yea I think you are right it should be only host until we add GPU reference verification (:

sdk/include/hipdnn_sdk/test_utilities/test_tensor.hpp
SamuelReeder: Yeah I think it could be helpful. Still need to make/manage `hipdnn_frontend::graph::Tensor_attributes` pointers too though.

BrianHarrisonAMD: Seems like Tensor is a good name, and then the frontend we could add utilities that combine the two together if that's ever needed.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
```diff
@@ -0,0 +1,89 @@
...
+            double invertVar = 1.0 / static_cast<double>(sqrt(static_cast<double>(variance) + static_cast<double>(epsilon)));
+            // process the batch per channel
+            for(int row = 0; row < height; row++)
+            { // via rows
+                for(int column = 0; column < width; column++)
+                { // via columns
+                    for(int bidx = 0; bidx < n_batches; bidx++)
+                    { // via mini_batch
+                        double elemStd = static_cast<double>(getValue<T>(input,bidx, cidx, row, column)) - static_cast<double>(mean);
+                        double inhat   = elemStd * invertVar;
```
BrianHarrisonAMD: Should this be double?
Seems like it should be V based off mean and variance above. 

Seems like:
- T == input & output data-type
- U == Bias & Scale data-type
- V == accumulation data-type

Using double instead of V for the intermediate values will give us extra precision, and not necessarily match GPU kernels, but I also could be misunderstanding the purpose of TUV.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
mousdahl-amd: The original version took in a BatchnormAttributes struct, but I removed that, and didn't realize that "execute" wasn't descriptive enough!  :D

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
mousdahl-amd: That was the intent, just forgot the parameter.

sdk/tests/CMakeLists.txt
mousdahl-amd: Unfortunately including half and bfloat16 cause linker errors if you don't link against hip::device.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
mousdahl-amd: I'm curious about this too.  The implementation I pulled from implicitly converted everything to doubles.

sdk/tests/utilities/test_migratable_memory.cpp
bghimireamd: Also call to allocate_device is not necessarily required?

bghimireamd (changes requested):

sdk/tests/CMakeLists.txt
BrianHarrisonAMD: Ah, well that's fine.
We will probably want to eventually add GPU reference implementations & verification anyways.

sdk/tests/test_utilities/test_cpu_fp_reference_implementation.cpp
BrianHarrisonAMD: Thinking about this more, lets add a separate task to generate the golden data values, and set-up a design for how we want to handle golden data.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
mousdahl-amd: Yes?  I'm not sure I understand the question.

sdk/include/hipdnn_sdk/test_utilities/test_tensor.hpp
mousdahl-amd: Perhaps we could call both the sdk and frontend versions "Tensor" and they're differentiated by namespace?  The frontend one could derive from this sdk one?

BrianHarrisonAMD: Oh I love that, and the frontend one can add the support for tensor_attribute management?

.clang-tidy
```diff
@@ -26,7 +26,7 @@ HeaderFileExtensions: ['h','hpp']
 
 # This regex is a inclusion filter for the files to be checked. Its a Posix regex and
 # does not seem to support negative lookahead.  
-HeaderFilterRegex: '^(.*backend.*|.*frontend.*|.*sdk/include/hipdnn_sdk/logging.*|.*sdk/include/hipdnn_sdk/plugin.*|.*plugins/.*)$'
+HeaderFilterRegex: '^(.*backend.*|.*frontend.*|.*sdk/include/hipdnn_sdk/logging.*|.*sdk/include/hipdnn_sdk/plugin.*|.*sdk/include/hipdnn_sdk/test_utilities.*|.*sdk/include/hipdnn_sdk/utilities.*|.*sdk/tests.*|.*plugins/.*)$'
```
adickin-amd: oh how I wish this was an exclusion filter :(.  We should check if clang 20 has support for exclusions.  I wrote this when we were on clang 18.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
mousdahl-amd: Hmm, made me think though.  I'll make the default seed std::random_device{}() instead.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_validation.hpp
mousdahl-amd: Updated format checking filters

sdk/tests/test_utilities/test_cpu_fp_reference_implementation.cpp
mousdahl-amd: Task added.,

.clang-tidy
mousdahl-amd: Looks like it's been added.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
mousdahl-amd: I have no idea what to call these, I pulled them directly from the MIOpen implementation.  I'll ask copilot.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
```diff
@@ -33,58 +58,112 @@ class Batchnorm_execute_graph_test : public ::testing::Test
         }
     }
 
+    template <typename Input_type>
+    // NOLINTNEXTLINE(readability-identifier-naming)
+    void RunFwdbatchnormGraph(Bn_2d_test_case test_case,
+                              hipdnn_sdk::data_objects::DataType input_data_type,
+                              Input_type epsilon);
```
SamuelReeder: Why is this NOLINT everywhere? Is it because of the number in the test_case type?

mousdahl-amd: Nah, it's the method name.  I wanted it to match our other test method names.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
```diff
@@ -33,58 +58,112 @@ class Batchnorm_execute_graph_test : public ::testing::Test
...
+        {.n = 2, .c = 3, .h = 14, .w = 14},
+        {.n = 64, .c = 3, .h = 14, .w = 14},
+        {.n = 64, .c = 256, .h = 14, .w = 14},
+        {.n = 64, .c = 256, .h = 28, .w = 28},
+        {.n = 64, .c = 256, .h = 56, .w = 56},
+        {.n = 64, .c = 512, .h = 14, .w = 14},
+        {.n = 64, .c = 512, .h = 28, .w = 28},
+        {.n = 64, .c = 512, .h = 7, .w = 7},
+        {.n = 64, .c = 64, .h = 112, .w = 112},
+        {.n = 64, .c = 64, .h = 56, .w = 56},
```
BrianHarrisonAMD: These are decently large, does it take a long time to run, or no?

BrianHarrisonAMD: Looks like 51 MB or 25.5 MB tensors in size. 

We might want to limit these to be smaller, and have the large tests only for the full E2E runs. 
Ideally the regular tests can run in a couple of minutes for the entire suite, and we separate out the long running tests into integration E2E tests.

mousdahl-amd: Some of them take 22 seconds.  Total test time for me was 100 seconds.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
```diff
@@ -0,0 +1,285 @@
...
+    {
+        other._host_ptr = nullptr;
+        other._device_ptr = nullptr;
+        other._count = 0;
+        other._item_size = 0;
+        other._total_size = 0;
+        other._current_location = Location::NONE;
+        other._host_valid = false;
+        other._device_valid = false;
+    }
```
SamuelReeder: Why does this move constructor not copy `_item_size` and `_total_size` too?

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
```diff
@@ -33,58 +58,112 @@ class Batchnorm_execute_graph_test : public ::testing::Test
...
-    variance_buffer.uid = 6;
-    variance_buffer.ptr = gpu_mem_variance.data();
-    device_buffers.push_back(variance_buffer);
+    Tensor x_tensor = Tensor::make_nchw_tensor<Input_type>(dims);
+    device_buffers.push_back(generate_random_device_buffer(
+        x_tensor, 1, static_cast<Input_type>(0.0f), static_cast<Input_type>(1.0f), seed));
+
+    Tensor y_tensor = Tensor::make_nchw_tensor<Input_type>(dims);
+    device_buffers.push_back(
+        generate_static_device_buffer(y_tensor, 2, static_cast<Input_type>(0.0f)));
```
BrianHarrisonAMD: Should probably  make Y random since we can't expect callers to 0 out their output tensors, and we would want to see the errors.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
mousdahl-amd: Good catch.  I added those later and missed it!

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
```diff
@@ -107,5 +186,38 @@ TEST_F(Batchnorm_execute_graph_test, RunFwdbatchnormGraph)
...
+    Tensor y_tensor_cpu = Tensor::make_nchw_tensor<Input_type>(dims);
+    y_tensor_cpu.fill_with_value(static_cast<Input_type>(0.0f));
+    Tensor scale_tensor_cpu = Tensor::make_nchw_tensor<float>(derived_dims);
+    scale_tensor_cpu.fill_with_random_values(0.0f, 1.0f, seed);
+    Tensor bias_tensor_cpu = Tensor::make_nchw_tensor<float>(derived_dims);
+    bias_tensor_cpu.fill_with_random_values(0.0f, 0.1f, seed);
+    Tensor mean_tensor_cpu = Tensor::make_nchw_tensor<float>(derived_dims);
+    mean_tensor_cpu.fill_with_random_values(0.0f, 0.5f, seed);
+    Tensor variance_tensor_cpu = Tensor::make_nchw_tensor<float>(derived_dims);
+    variance_tensor_cpu.fill_with_random_values(0.1f, 0.3f, seed);
```
BrianHarrisonAMD: Oh I thought this would fill the CPU tensor with random values, and then transfer it to the GPU to ensure they match.
Is the random value generator guaranteed to be the same for GPU and CPU?

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
```diff
@@ -0,0 +1,83 @@
...
+        std::iota(channels.begin(), channels.end(), 0);
+        int64_t height = input.dims().at(2);
+        int64_t width = input.dims().at(3);
+
+        std::for_each(channels.begin(), channels.end(), [&](int64_t cidx) {
+            auto mean = estimatedMean.get_host_value<Mean_variance_data_type>(0, cidx, 0, 0);
+            auto variance
+                = estimatedVariance.get_host_value<Mean_variance_data_type>(0, cidx, 0, 0);
+            Mean_variance_data_type invert_var
+                = static_cast<Mean_variance_data_type>(1.0f) / sqrtf(variance + epsilon);
```
jerehartAMD: Would `sqrtf` here reduce the precision in the case that variance and epsilon are doubles?

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_validation.hpp
```diff
@@ -0,0 +1,71 @@
...
+        const T* ref_data = reference.host_data<T>();
+        const T* impl_data = implementation.host_data<T>();
+
+        for(size_t i = 0; i < element_count; ++i)
+        {
+
+            T ref_value = ref_data[i];
+            T impl_value = impl_data[i];
+
+            T abs_diff = std::fabs(ref_value - impl_value);
```
jerehartAMD, line 51: Same here, this might involve a cast to and from float

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
mousdahl-amd: Yeah, in both cases they use the same seed.  They both get generated on host and then the GPU one gets copied over on-demand in Migratable_memory

BrianHarrisonAMD: Okay it does do that, and uses same seed. 
I guess we are just making a separate copy for the host side verification.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_validation.hpp
jerehartAMD: Scratch this one, `sqrtf` is specific to float, but `fabs` is the right one here

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
BrianHarrisonAMD: Okay I think we can leave this for now, but once we add E2E tests we should migrate some of these, and keep this list tiny.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
mousdahl-amd: I didn't think we had double as a datatype, for some reason, but we do.  I'll add the tests and make sure.  The compiler complains pretty loudly about precision loss.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
```diff
@@ -0,0 +1,285 @@
...
+        allocate_host();
+
+        if(!_host_valid && _device_valid && (_device_ptr != nullptr))
+        {
+            throw_on_error(hipMemcpy(_host_ptr, _device_ptr, _total_size, hipMemcpyDeviceToHost),
+                           "Failed to copy from device to host");
+            _host_valid = true;
+            _current_location = Location::BOTH;
+        }
+    }
```
SamuelReeder, line 259: If two threads enter allocate_host at same time then couldn't you get race condition and malloc twice? I'm assuming we want to be thread safe. I think the same is applicable to device methods.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_validation.hpp
```diff
@@ -0,0 +1,71 @@
...
+    {
+        if(reference.count() != implementation.count())
+        {
+            return false;
+        }
+
+        size_t element_count = reference.count();
+        const T* ref_data = reference.host_data<T>();
+        const T* impl_data = implementation.host_data<T>();
+
```
bghimireamd, line 44: Additional checks to consider:
1. Check for buffers containing all zeros (could indicate uninitialized memory or incorrect kernel execution)
2. Check for infinity and NaN/inf values (could indicate numerical overflow or division by zero)

sdk/tests/test_utilities/test_cpu_fp_reference_implementation.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    Cpu_fp_reference_implementation<float, float, float> ref_impl;
+
+    ref_impl.batchnorm_fwd_inference(input_tensor,
+                                     scale_tensor,
+                                     bias_tensor,
+                                     mean_tensor,
+                                     variance_tensor,
+                                     output_tensor,
+                                     1e-5f);
```
SamuelReeder: Might be good to pull out calculator and compute a golden value to check the reference is correct, but it can probably wait.

bghimireamd: Yes, I think we will soon be adding ticket for this.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
SamuelReeder: Just mean the ability to set the seed manually so that the result is reproduceable. A repo-wide seed would be best at some point.

sdk/tests/test_utilities/test_cpu_fp_reference_implementation.cpp
mousdahl-amd: Yeah, we've got a ticket

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
mousdahl-amd: I'm not sure I understand this comment.

BrianHarrisonAMD: The y_tensor should be populated with random memory similar to the others.
Since the API caller will give us device buffers that may have pre-existing junk values, and we don't require them to clear the output buffers ahead of time.

BrianHarrisonAMD: We have had issues where our tests have cleared output values, and then we never noticed issues as a result.
By using random values for Y with a large range of values, we are making our tests match a real environment better.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_validation.hpp
```diff
@@ -0,0 +1,71 @@
...
+            if(abs_diff > _absolute_tolerance
+               && abs_diff
+                      > _relative_tolerance * std::max(std::fabs(ref_value), std::fabs(impl_value)))
+            {
+
+                return false;
+            }
+        }
+
+        return true;
```
BrianHarrisonAMD: Not sure if this is a good spot for logging, but would be nice to log these values as info for success, and warning or error for failure.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
BrianHarrisonAMD: I think you can also just implement the move operator, and then use that instead of duplicating.

```
Migratable_memory(Migratable_memory&& other) noexcept
{
    *this = std::move(other);
}
```

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
```diff
@@ -0,0 +1,285 @@
...
+        {
+            cleanup();
+            _host_ptr = other._host_ptr;
+            _device_ptr = other._device_ptr;
+            _count = other._count;
+            _item_size = other._item_size;
+            _total_size = other._total_size;
+            _current_location = other._current_location;
+            _host_valid = other._host_valid;
+            _device_valid = other._device_valid;
```
BrianHarrisonAMD, line 79: nit: I think if you made cleanup clear the member variables as well, and used std::swap you could consolidate the move operator a bit.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
BrianHarrisonAMD: I think it's probably fine to just add a comment indicating it's not thread safe.
We are only allowing 1 thread per handle, and I can't think of a situation in which we would want someone allocating and touching Tensors / migratable memory across threads.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
```diff
@@ -0,0 +1,285 @@
...
+        {
+            return;
+        }
+
+        allocate_host();
+
+        if(!_host_valid && _device_valid && (_device_ptr != nullptr))
+        {
+            throw_on_error(hipMemcpy(_host_ptr, _device_ptr, _total_size, hipMemcpyDeviceToHost),
+                           "Failed to copy from device to host");
```
BrianHarrisonAMD, line 255: hipMemcpy will require the stream actually.
Right now this will only work on the default stream, but multi-stream access will explode since it won't synchronize.

We can probably leave this for now, but need to add a story where you can specify a hipStream when constructing the Tensors to use for synchronization purposes.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
BrianHarrisonAMD: That did remind me about stream synchronization though!

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
BrianHarrisonAMD, line 1: We should add NHWC layout tests as well (right now all tensors are NCHW layout), but that can be a separate follow-up since this is already pretty big.

Edit: whoops flipped the layouts.

sdk/include/hipdnn_sdk/utilities/tensor.hpp
```diff
@@ -0,0 +1,131 @@
...
+    // Default move constructor and move assignment operator
+    // These will automatically move _memory, _dims, and _strides
+    Tensor(Tensor&&) = default;
+    Tensor& operator=(Tensor&&) = default;
+
+    template <typename T>
+    static Tensor make_nchw_tensor(const std::vector<int64_t>& dims)
+    {
+        return {dims, generate_strides(dims, {3, 2, 1, 0}), sizeof(T)};
+    }
```
BrianHarrisonAMD, line 42: Oh nice.
We should add the NHWC version as well.

BrianHarrisonAMD (approved):
Left a few minor comments, but generally LGTM!

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
mousdahl-amd: I got it working with doubles to find out that MIOpen doesn't support double yet.  Oh well, the work is done to correct this anyhow.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_validation.hpp
mousdahl-amd: I think that would make more sense in a check_buffer method.

sdk/include/hipdnn_sdk/utilities/tensor.hpp
mousdahl-amd: I'll add a story for later.

bghimireamd (approved):
generally LGTM

SamuelReeder (approved):

PR #84, opened 2025-07-27
Draft of first step of integration changes

BrianHarrisonAMD:
## Proposed changes

First version of a backend integration.
Still a bit rough, and public API tests are broken since now we are trying to use real plugins that don't exist.

### DIFF
--- a/backend/src/descriptors/backend_descriptor.hpp
+++ b/backend/src/descriptors/backend_descriptor.hpp
@@ -27,11 +27,13 @@ struct Backend_descriptor_interface
                                hipdnnBackendAttributeType_t attribute_type,
                                int64_t requested_element_count,
                                int64_t* element_count,
-                               void* array_of_elements) const = 0;
+                               void* array_of_elements) const
+        = 0;
     virtual void set_attribute(hipdnnBackendAttributeName_t attribute_name,
                                hipdnnBackendAttributeType_t attribute_type,
                                int64_t element_count,
-                               const void* array_of_elements) = 0;
+                               const void* array_of_elements)
+        = 0;
 
     virtual hipdnnBackendDescriptorType_t get_type() const = 0;
 };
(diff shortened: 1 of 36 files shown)


backend/src/descriptors/backend_descriptor.hpp
```diff
@@ -99,6 +99,15 @@ struct hipdnnBackendDescriptor : public Backend_descriptor_interface
...
 
+    // Unsafe version of as_descriptor that does not perform type checks.
+    // Use with caution, as it can lead to undefined behavior if the types do not match
+    // This is intended for internal mock usage, and not anywhere esle in the codebase.
+    template <typename Child_descriptor>
+    std::shared_ptr<Child_descriptor> as_descriptor_unsafe()
+    {
+        return std::static_pointer_cast<Child_descriptor>(_impl);
+    }
+
```
BrianHarrisonAMD: Not sure how we can handle this.
I could make a mock descriptor that extends backend_descriptor, and at least hide this as protected method to give ourself some protection.

backend/tests/descriptors/test_engine_config_descriptor.cpp
```diff
@@ -50,29 +64,36 @@ class Engine_config_descriptor_test : public ::testing::Test
...
+        EXPECT_CALL(*get_mock_engine(), is_finalized()).WillRepeatedly(Return(true));
+        EXPECT_CALL(*get_mock_engine(), get_engine_id()).WillRepeatedly(Return(1));
+        EXPECT_CALL(*get_mock_engine(), get_graph())
+            .WillRepeatedly(Return(get_mock_graph_descriptor()));
+        EXPECT_CALL(*get_mock_graph_descriptor(), get_handle())
+            .WillOnce(Return(_mock_handle.get()));
+        EXPECT_CALL(*_mock_handle, get_plugin_resource_manager())
+            .WillOnce(Return(_mock_engine_plugin_resource_manager));
+        EXPECT_CALL(*_mock_engine_plugin_resource_manager, get_workspace_size(_, _, _))
+            .WillOnce(Return(1024));
```
BrianHarrisonAMD, line 79: Basically have to mock the entire chain of what is happening in finalize now.
If we want to internalize the finalize logic to the descriptor I think this is what we need to do.

Alternatively, we could probably have some finalizer class that does it, and just mock that piece.
It would probably have to be friended with the associated descriptor, and the plugin_resource_manager.

PR #85, opened 2025-07-28
Engine config descriptor integration, and changes to support it

BrianHarrisonAMD:
## Proposed changes

Add engine config descriptor integration logic, remove stubbed integration logic, update tests to support it.

### DIFF
--- a/backend/src/descriptors/backend_descriptor.hpp
+++ b/backend/src/descriptors/backend_descriptor.hpp
@@ -99,6 +99,15 @@ struct hipdnnBackendDescriptor : public Backend_descriptor_interface
             "Failed to cast backend descriptor: Null descriptor provided.");
     }
 
+    // Unsafe version of as_descriptor that does not perform type checks.
+    // Use with caution, as it can lead to undefined behavior if the types do not match
+    // This is intended for internal mock usage, and not anywhere esle in the codebase.
+    template <typename Child_descriptor>
+    std::shared_ptr<Child_descriptor> as_descriptor_unsafe()
+    {
+        return std::static_pointer_cast<Child_descriptor>(_impl);
+    }
+
     bool is_valid();
 
     hipdnnBackendDescriptorType_t get_type() const override;
--- a/backend/src/descriptors/engine_config_descriptor.cpp
+++ b/backend/src/descriptors/engine_config_descriptor.cpp
@@ -4,8 +4,11 @@
 #include "engine_config_descriptor.hpp"
 #include "engine_descriptor.hpp"
 #include "error.hpp"
+#include "graph_descriptor.hpp"
 #include "hipdnn_backend_descriptor_type.h"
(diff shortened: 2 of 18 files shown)


backend/src/descriptors/graph_descriptor.hpp
```diff
@@ -39,7 +39,8 @@ class Graph_descriptor : public hipdnnBackendDescriptorImpl<Graph_descriptor>
 
     void deserialize_graph(const uint8_t* serialized_graph, size_t graph_byte_size);
 
-    const std::vector<uint8_t>& get_serialized_graph();
+    const std::vector<uint8_t>& get_serialized_graph() const;
```
BrianHarrisonAMD, line 42: This should be updated to match how engine_config_descriptor works, and we should find a way to prevent duplicating the flatbuffer memory (currently it copies the flatbuffer serialzied data into the std::vector).

backend/src/descriptors/engine_config_descriptor.cpp
```diff
@@ -196,4 +197,25 @@ hipdnnBackendDescriptorType_t Engine_config_descriptor::get_static_type()
...
+                      HIPDNN_STATUS_INTERNAL_ERROR,
+                      "Graph_descriptor::get_serialized_graph: graph is null");
+
+        flatbuffers::FlatBufferBuilder builder;
+        hipdnn_sdk::data_objects::EngineConfigBuilder engine_config_builder(builder);
+        engine_config_builder.add_engine_id(_engine->get_engine_id());
+        builder.Finish(engine_config_builder.Finish());
+        _engine_config_buffer = builder.Release();
+        _serialized_engine_config
+            = {.ptr = _engine_config_buffer.data(), .size = _engine_config_buffer.size()};
```
BrianHarrisonAMD, line 215: We want to use the flatbuffer objects directly, and set them as we set the other associated objects (in this case engine_id from when engine gets set). 

We should keep the serialization pattern, but find a way to not store the detached buffer, and instead store the flatbuffer object ideally.

backend/src/descriptors/backend_descriptor.hpp
```diff
@@ -99,6 +99,15 @@ struct hipdnnBackendDescriptor : public Backend_descriptor_interface
...
     }
 
+    // Unsafe version of as_descriptor that does not perform type checks.
+    // Use with caution, as it can lead to undefined behavior if the types do not match
+    // This is intended for internal mock usage, and not anywhere esle in the codebase.
+    template <typename Child_descriptor>
+    std::shared_ptr<Child_descriptor> as_descriptor_unsafe()
+    {
+        return std::static_pointer_cast<Child_descriptor>(_impl);
+    }
```
BrianHarrisonAMD, line 109: Add a todo for clean-up, since we ideally have a different way to allow this for mocks.

averinevg (approved):

PR #86, opened 2025-07-28
On going integration branch changes PR 

BrianHarrisonAMD:
PR for the integration between backend & plugins.
Several PRs that have been reviewed were merged into this branch.
Lots of changes due to swapping to a real plugin backend.


backend/src/descriptors/backend_descriptor.hpp
```diff
@@ -99,6 +99,15 @@ struct hipdnnBackendDescriptor : public Backend_descriptor_interface
...
     }
 
+    // Unsafe version of as_descriptor that does not perform type checks.
+    // Use with caution, as it can lead to undefined behavior if the types do not match
+    // This is intended for internal mock usage, and not anywhere esle in the codebase.
+    template <typename Child_descriptor>
+    std::shared_ptr<Child_descriptor> as_descriptor_unsafe()
+    {
+        return std::static_pointer_cast<Child_descriptor>(_impl);
+    }
```
BrianHarrisonAMD: Ill try to figure out a better way to handle this for the Mocks.
I don't like having this possible.

backend/src/descriptors/engine_config_descriptor.cpp
```diff
@@ -196,4 +197,25 @@ hipdnnBackendDescriptorType_t Engine_config_descriptor::get_static_type()
...
+        hipdnn_sdk::data_objects::EngineConfigBuilder engine_config_builder(builder);
+        engine_config_builder.add_engine_id(_engine->get_engine_id());
+        builder.Finish(engine_config_builder.Finish());
+        _engine_config_buffer = builder.Release();
+        _serialized_engine_config
+            = {.ptr = _engine_config_buffer.data(), .size = _engine_config_buffer.size()};
+    }
+
+    return _serialized_engine_config;
+}
```
BrianHarrisonAMD: I am going to swap this to use the flatbuffer objects, and then lazily generate the serialzied plugin data when it's needed. 

Ideally we can have a single flatbuffer object, and then just populate values on it during the creation of the descriptor, and once finalized we can cache the hipdnnPluginConstData_t without duplicating or copying any memory.

backend/src/descriptors/graph_descriptor.hpp
```diff
@@ -39,7 +39,8 @@ class Graph_descriptor : public hipdnnBackendDescriptorImpl<Graph_descriptor>
 
     void deserialize_graph(const uint8_t* serialized_graph, size_t graph_byte_size);
 
-    const std::vector<uint8_t>& get_serialized_graph();
+    const std::vector<uint8_t>& get_serialized_graph() const;
```
BrianHarrisonAMD: Swap this to match with engine_config_descriptor.
This will also ideally have a cached representation of the hipdnnPluginConstData_t once the graph is finalized, and we ideally don't need to duplicate the storage.

backend/tests/descriptors/test_engine_config_descriptor.cpp
```diff
@@ -50,29 +64,36 @@ class Engine_config_descriptor_test : public ::testing::Test
...
+        EXPECT_CALL(*get_mock_engine(), is_finalized()).WillRepeatedly(Return(true));
+        EXPECT_CALL(*get_mock_engine(), get_engine_id()).WillRepeatedly(Return(1));
+        EXPECT_CALL(*get_mock_engine(), get_graph())
+            .WillRepeatedly(Return(get_mock_graph_descriptor()));
+        EXPECT_CALL(*get_mock_graph_descriptor(), get_handle())
+            .WillOnce(Return(_mock_handle.get()));
+        EXPECT_CALL(*_mock_handle, get_plugin_resource_manager())
+            .WillOnce(Return(_mock_engine_plugin_resource_manager));
+        EXPECT_CALL(*_mock_engine_plugin_resource_manager, get_workspace_size(_, _, _))
+            .WillOnce(Return(1024));
```
BrianHarrisonAMD, line 82: We should actually add some tests here as well for grabbing the workspace after finalization. 
Before it was a post-op, but now after finalization you should be able query workspace size, and it should match what the mock is returning. 

We should also check if a negative workspace size throws as expected, and cover the error cases.

backend/src/descriptors/backend_descriptor.hpp
BrianHarrisonAMD: Friending the mocks seems to be the only decent way to handle this.

backend/tests/descriptors/test_engine_config_descriptor.cpp
BrianHarrisonAMD: Still need to do this, but can be a follow-up.

backend/src/descriptors/engine_config_descriptor.hpp
```diff
@@ -46,12 +52,12 @@ class Engine_config_descriptor : public hipdnnBackendDescriptorImpl<Engine_confi
...
-    void set_max_workspace_size(int64_t max_workspace_size);
+    static hipdnnBackendDescriptorType_t get_static_type();
 
-    // Throws an exception if the descriptor is not finalized.
-    std::shared_ptr<const Engine_descriptor> get_engine() const;
+    // Throws an exception if the descriptor is not finalized before calling these.
+    virtual std::shared_ptr<const Engine_descriptor> get_engine() const;
 
-    static hipdnnBackendDescriptorType_t get_static_type();
+    virtual hipdnnPluginConstData_t get_serialized_engine_config() const;
```
BrianHarrisonAMD, line 60: We should generally review these public accessors and see which are still needed.

I think we ended up needing them in most cases, and we can consider if a rename is worth it to make it more clear these will throw if they aren't finalized first.

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -91,6 +109,8 @@ Engine_plugin_resource_manager::Engine_plugin_resource_manager(
     {
         auto handle = plugin.create_handle();
 
+        plugin.set_logging_callback(logging::hipdnn_logging_callback);
+
```
BrianHarrisonAMD: This probably should be in a different spot since I assume this will be called multiple times if we make multiple handles?

BrianHarrisonAMD: Should really be calling it when we construct the plugins, but ill have to see if thats problematic for some reason.

SamuelReeder: I moved it to where plugins are loaded in plugin core. Can I push to this branch?

plugins/miopen_legacy_plugin/integration_tests/batchnorm_fwd_inference_integration_test.cpp
```diff
@@ -0,0 +1,340 @@
...
+INSTANTIATE_TEST_SUITE_P(RunFloatFwdBatchnormGraph,
+                         Batchnorm_forward_inference_integration_test,
+                         testing::ValuesIn(get_bn_fwd_inference_test_cases()));
+
+class Batchnorm_forward_inference_integration_test_bfloat16
+    : public Batchnorm_forward_inference_integration_test
+{
+};
+
+TEST_P(Batchnorm_forward_inference_integration_test_bfloat16, RunBfloat16FwdBatchnormGraph)
```
adickin-amd, line 316: gotta add the device skip macro to these tests.

tests/backend/test_backend_public_functions.cpp
```diff
@@ -164,4 +165,55 @@ TEST(HipDNNBackendTest, WillFailToCreateGraphIfGraphIsNull)
...
+    std::array<const char*, 2> paths = {"./valid/path.so", nullptr};
+
+    hipdnnStatus_t status = hipdnnSetEnginePluginPaths_ext(
+        paths.size(), paths.data(), HIPDNN_PLUGIN_LOADING_ABSOLUTE);
+
+    EXPECT_EQ(status, HIPDNN_STATUS_BAD_PARAM_NULL_POINTER);
+}
+
+// Uncomment this test and observe logs to see the plugin is loaded
+// TEST(HipDNNBackendTest, SetPluginPathAndCreateHandle_LoadsPlugins)
```
adickin-amd, line 201: is this something we should just have on by default?  Can we make it something thats on by default or are we waiting for some test plugins?

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -279,136 +291,6 @@ void Engine_plugin_resource_manager::execute_op_graph(
         handle, execution_context, workspace, device_buffers, num_device_buffers);
 }
 
-#if 0
```
mousdahl-amd, line 282: Should this go?

tests/backend/test_backend_public_functions.cpp
SamuelReeder: We need a API call to get the plugins, but it doesn't exist yet. I will make a ticket to complete this.

backend/src/plugin/shared_library.cpp
```diff
@@ -6,13 +6,57 @@
...
 #include "hipdnn_exception.hpp"
+#include "logging/logging.hpp"
 #include "shared_library.hpp"
 
 namespace hipdnn_backend
 {
 namespace plugin
 {
 
+std::filesystem::path Shared_library::get_current_module_directory()
```
mousdahl-amd, line 17: Let's leave a note to add this to a platform_utils.hpp in the sdk.

backend/src/plugin/shared_library.cpp
```diff
@@ -62,7 +106,7 @@ void Shared_library::load(const std::filesystem::path& library_path)
 
     auto modified_library_path = library_path;
 
-    // Check file extension
+    // Check file extension and add prefix/suffix if needed
     if(modified_library_path.has_extension())
     {
 #ifdef _WIN32
```
mousdahl-amd, line 112: We've got constexpr const char* SHARED_LIB_EXT in plugin_core.hpp.  Let's add a ticket for platform-specific cleanup.  I'd rather have a `platform_utils.hpp` header, and conditionally include `platform_utils.win32.hpp` within it (and `platform_utils.linux.hpp`)

SamuelReeder: Yeah I didn't want to put any of the platform logic in plugin_core.hpp, but had to make one exception. Adding ticket. I intend to do a follow up PR soon for some of this cleanup.

backend/tests/test_plugin.cpp
```diff
@@ -67,12 +102,166 @@ TEST(PluginManagerTest, LoadPlugins)
...
+    }
+    catch(...)
+    {
+        std::filesystem::remove_all(default_dir);
+        FAIL();
+    }
+    std::filesystem::remove_all(default_dir);
+}
+
+TEST(PluginManagerTest, LoadPluginsCombinedFileAndDirectory)
```
mousdahl-amd, line 227: Should we have a test for a malformed plugin (eg. missing an entry point)?  I'm thinking of future-proofing, like say at some future version we require that plugins have `cool_new_function()`, it would stink if things entirely blew up because they still had an old plugin.

SamuelReeder: Kind of unrelated, but should also test a plugin with no applicable engines, and tests both plugins simultaneously. Lots of testing to do for plugins still.

backend/tests/descriptors/test_engine_config_descriptor.cpp
BrianHarrisonAMD: Moving to follow-up task

backend/tests/test_plugin.cpp
BrianHarrisonAMD: Yea we need:
- Unhappy path where things fail at certain points.
- Loading and using multiple plugins in tests to ensure that behavior is correct.
- Tests checking the fail conditions for loading plugins that are malformed.
- Probably a only partially implemented plugin to ensure that it doesn't look when API mismatches (future proofing a bit)

BrianHarrisonAMD: Ill add a note to the follow-up, and we can likely split it into a few stories.

backend/src/hipdnn_backend.cpp
```diff
@@ -299,4 +283,35 @@ HIPDNN_BACKEND_EXPORT void hipdnnGetLastErrorString(char* message, size_t max_si
 HIPDNN_BACKEND_EXPORT void hipdnnLoggingCallback_ext(hipdnnSeverity_t severity, const char* msg)
 {
     hipdnn_backend::logging::hipdnn_logging_callback(severity, msg);
+}
+
+HIPDNN_BACKEND_EXPORT hipdnnStatus_t hipdnnSetEnginePluginPaths_ext(
+    size_t num_paths, const char* const* plugin_paths, hipdnnPluginLoadingMode_ext_t loading_mode)
+{
+    LOG_API_ENTRY("num_paths={}, plugin_paths_ptr={:p}, loading_mode={}",
```
jerehartAMD, line 291: :p

backend/src/plugin/plugin_core.hpp
```diff
@@ -85,53 +97,193 @@ class Plugin_manager_base
...
+        for(const auto& path : _default_plugin_paths)
+        {
+            if(path.is_relative())
+            {
+                resolved_paths.insert(base_dir / path);
+            }
+            else
+            {
+                resolved_paths.insert(path);
+            }
```
jerehartAMD, line 135: alternative:
```suggestion
            resolved_paths.insert(path.is_relative() ? base_dir / path : path);
```
Also, is path guaranteed to be relative from base_dir?

backend/src/plugin/shared_library.cpp
```diff
@@ -95,32 +139,45 @@ void Shared_library::load(const std::filesystem::path& library_path)
...
+                                   + _library_path.string());
+    }
+
+    HIPDNN_LOG_INFO(
+        "Shared_library: Attempting to load shared library from final absolute path: {}",
+        _library_path.string());
 
 #ifdef _WIN32
-    _library_handle = LoadLibraryW(modified_library_path.wstring().c_str());
+    _library_handle = LoadLibraryW(_library_path.wstring().c_str());
```
jerehartAMD, line 157: I forgot how much I hated windows for this

backend/tests/test_plugin.cpp
```diff
@@ -37,19 +54,37 @@ void dummy_callback(hipdnnSeverity_t sev, const char* msg)
...
+const std::string PLUGIN_NAME1 = "hipdnn_test_plugin1";
+const std::string PLUGIN_NAME2 = "hipdnn_test_plugin2";
+
+const std::string PLUGIN_PATH1 = "./" + PLUGIN_NAME1;
+const std::string PLUGIN_PATH2 = "./" + PLUGIN_NAME2;
+
+const std::string FULL_PLUGIN_PATH1
+    = std::string("./") + LIB_PREFIX + PLUGIN_NAME1 + SHARED_LIB_EXT;
+const std::string FULL_PLUGIN_PATH2
+    = std::string("./") + LIB_PREFIX + PLUGIN_NAME2 + SHARED_LIB_EXT;
```
jerehartAMD, line 74: Could these be `std::filesystem`s instead?

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -274,7 +274,7 @@ class Graph : public INode
                                                    HIPDNN_ATTR_EXECUTION_PLAN_ENGINE_CONFIG,
                                                    HIPDNN_TYPE_BACKEND_DESCRIPTOR,
                                                    1,
-                                                   _engine_config_desc->get()),
+                                                   &_engine_config_desc->get()),
```
jerehartAMD, line 277: Why are we taking a pointer to a pointer here? Won't this copy the pointer to the `engine_config_desc` instead?

jerehartAMD (commented):
No serious comments, just some notes and questions!

frontend/include/hipdnn_frontend/graph.hpp
adickin-amd: This is due to how the API takes in descriptors.  Its always an array_of_elements https://github.com/ROCm/hipDNN/blob/develop/backend/include/hipdnn_backend.h#L222.  So when we unpack its considered an array of engine_config_descriptors.

Here we are returning a `const hipdnnBackendDescriptor_t&` so its a pointer to a const reference of a `hipdnnBackendDescriptor*`

adickin-amd: https://github.com/ROCm/hipDNN/blob/develop/frontend/include/hipdnn_frontend/backend/scoped_hipdnn_backend_descriptor.hpp#L124
here is where get is implemented

PR #87, opened 2025-07-28
[ALMIOPEN-139] Engine plugin loading paths & mode

SamuelReeder:
## Proposed changes

Enables user-specified plugin loading that can search multiple directories for any plugins and/or load specific plugin libraries. This PR also introduces plugin loading modes so the user can specify the policy for loading plugins when they give multiple paths.

## Checklist


frontend/include/hipdnn_frontend/backend/hipdnn_backend_interface.hpp
```diff
@@ -50,6 +50,13 @@ class Hipdnn_backend_interface
                                                  size_t graph_byte_size)
         = 0;
     virtual void logging_callback_ext(hipdnnSeverity_t severity, const char* msg) = 0;
+
+    // NOLINTBEGIN
+    virtual hipdnnStatus_t set_plugin_paths_ext(int64_t num_paths,
+                                                const char* const plugin_paths[],
+                                                hipdnnPluginLoadingMode_t mode)
+        = 0;
+    // NOLINTEND
```
SamuelReeder: Needed for C-style array.

frontend/tests/test_plugin_paths.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include <array>
+#include <gtest/gtest.h>
+#include <hipdnn_backend.h>
+#include <hipdnn_frontend/backend/backend_wrapper.hpp>
+#include <hipdnn_status.h>
+
+using namespace hipdnn_frontend;
+
+TEST(HipdnnBackendWrapperTest, SetPluginPaths_Success)
```
SamuelReeder: These tests are trivial and only test the backend wrapper. I will probably remove unless we intend to test the wrapper.

backend/tests/test_plugin.cpp
```diff
@@ -67,12 +78,113 @@ TEST(PluginManagerTest, LoadPlugins)
     ASSERT_EQ(plugins[1].type(), HIPDNN_PLUGIN_TYPE_UNSPECIFIED);
 }
 
+TEST(PluginManagerTest, LoadPluginsFromDirectory)
```
SamuelReeder, line 105: Some of these tests create temporary files which is the easiest way to test this.

backend/src/hipdnn_backend.cpp
```diff
@@ -293,4 +294,34 @@ HIPDNN_BACKEND_EXPORT void hipdnnGetLastErrorString(char* message, size_t max_si
 HIPDNN_BACKEND_EXPORT void hipdnnLoggingCallback_ext(hipdnnSeverity_t severity, const char* msg)
 {
     hipdnn_backend::logging::hipdnn_logging_callback(severity, msg);
+}
+
+HIPDNN_BACKEND_EXPORT hipdnnStatus_t hipdnnSetPluginPaths_ext(
+    size_t num_paths, const char* const plugin_paths[], hipdnnPluginLoadingMode_t loading_mode)
```
mousdahl-amd: So at a high level, it's SetPluginPaths(), but internally it's only affecting Engine plugins?  Are the heuristic and other plugins handled in separate internal methods, or is the intention here that they're all managed through this function?

backend/src/plugin/plugin_core.hpp
```diff
@@ -130,8 +129,76 @@ class Plugin_manager_base
...
+            {
+                if(entry.is_regular_file())
+                {
+                    load_plugin_from_file(entry.path(), mode);
+                }
+            }
+        }
+        catch(const std::filesystem::filesystem_error& e)
+        {
+            HIPDNN_LOG_WARN(
```
mousdahl-amd, line 276: So we catch filesystem exceptions here, is it meant to re-throw, because there's a catch above this too...

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -30,32 +30,53 @@ class Engine_plugin_manager : public Plugin_manager_base<Engine_plugin>
...
-    const std::vector<std::filesystem::path>& plugin_paths)
+    const std::vector<std::filesystem::path>& plugin_paths, hipdnnPluginLoadingMode_t loading_mode)
 {
     std::lock_guard<std::mutex> lock(plugin_mutex);
 
-    // Check if the plugin paths are already saved, if so, do nothing.
-    if(!override_plugin_paths.empty())
+    THROW_IF_FALSE(pm_ptr.expired(),
+                   HIPDNN_STATUS_NOT_SUPPORTED,
+                   "hipdnnSetPluginPaths_ext cannot be called with an active handle.");
```
SamuelReeder: I think this check is enough to ensure plugins paths are set before their handle is created because `hipdnnHandle()` creates the `Engine_plugin_resource_manager` which in turn creates `Engine_plugin_manager`, so if it is expired then the handle has not been created yet or has been deleted.

backend/tests/test_engine_plugin.cpp
```diff
@@ -20,6 +20,12 @@ using Scoped_resource = hipdnn::sdk::utilities::Scoped_resource<T, Destructor>;
 
 using Test_engine_plugin_manager = plugin::Plugin_manager_base<plugin::Engine_plugin>;
 
+#if defined(_WIN32)
+constexpr const char* SHARED_LIB_EXT = ".dll";
```
mousdahl-amd: Maybe we need a platform_utils.hpp

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -30,32 +30,53 @@ class Engine_plugin_manager : public Plugin_manager_base<Engine_plugin>
...
+    if(loading_mode == HIPDNN_PLUGIN_LOADING_ABSOLUTE)
     {
-        return;
+        plugin_config.paths = plugin_paths;
+    }
+    else
+    {
+        plugin_config.paths.insert(
+            plugin_config.paths.end(), plugin_paths.begin(), plugin_paths.end());
     }
```
SamuelReeder: I need to add default path here too.

mousdahl-amd (commented):
Looks mostly good, I'm just a bit confused about the granularity of plugin paths.  Are they per plugin type (engine, heuristic, etc) or is there meant to be one big plugins folder that just has them all mixed together?

backend/src/hipdnn_backend.cpp
SamuelReeder: I think I should change this to `hipdnnSetEnginePluginPaths_ext`, and make other methods for different plugins in the future.

backend/src/plugin/plugin_core.hpp
SamuelReeder: I'll just keep the outermost filesystem exception catch.

backend/tests/test_engine_plugin.cpp
SamuelReeder: Good point, if we don't already have anything like that I can make it in sdk.

SamuelReeder:
> Looks mostly good, I'm just a bit confused about the granularity of plugin paths. Are they per plugin type (engine, heuristic, etc) or is there meant to be one big plugins folder that just has them all mixed together?

Right now everything should be for only engine plugins (I have to cleanup some naming), and then it should be the user's responsibility to manage plugin paths per type if they deviate from default plugin paths. I'll make the default engine plugin path inside a `engine/` subdirectory in `/opt/rocm/lib/hipdnn_plugins/` so the types are organized. If there is a way to detect the type of a plugin pre-load, then in that case all types could be in one directory I believe.

BrianHarrisonAMD:
> > Looks mostly good, I'm just a bit confused about the granularity of plugin paths. Are they per plugin type (engine, heuristic, etc) or is there meant to be one big plugins folder that just has them all mixed together?
> 
> Right now everything should be for only engine plugins (I have to cleanup some naming), and then it should be the user's responsibility to manage plugin paths per type if they deviate from default plugin paths. I'll make the default engine plugin path inside a `engine/` subdirectory in `/opt/rocm/lib/hipdnn_plugins/` so the types are organized. If there is a way to detect the type of a plugin pre-load, then in that case all types could be in one directory I believe.

I think we should likely add an enum, or make it part of the API call to make it explicit.

backend/src/logging/logging.hpp
```diff
@@ -8,6 +8,7 @@
 #include <spdlog/spdlog.h>
 
 #ifdef HIPDNN_BACKEND_COMPILATION
+// NOLINTBEGIN
```
BrianHarrisonAMD: Huh why?

SamuelReeder: The macros sometimes introduce too much cognitive complexity, but I'll suppress only that warning instead.

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -30,32 +30,59 @@ class Engine_plugin_manager : public Plugin_manager_base<Engine_plugin>
...
 std::weak_ptr<Engine_plugin_manager> pm_ptr;
 
 std::vector<std::filesystem::path> get_default_plugin_paths()
 {
-    // This function should return the default plugin paths.
-    // For now, we return an empty vector.
-    // TODO: Implement logic to retrieve default plugin paths.
-    return {};
+    return {"/opt/rocm/lib/hipdnn_plugins/engine/"};
 }
```
BrianHarrisonAMD: This needs to swap to find the current absolute path for the .so, and then add /hipdnn_plugins/engine/ to it. 
For example, if libhipdnnbackend.so is isntalled into /opt/rocm/lib, then the final path would be what is here, but it might be anywhere on disk.

backend/src/logging/logging.hpp
BrianHarrisonAMD: Surprised this works actually.
I would have expected it to flag in the sources using the macros, and this to only suppress around the macro, but it actually does bring this suppress across with it?

backend/include/hipdnn_backend_plugin_loading_mode.h
```diff
@@ -0,0 +1,11 @@
+// Copyright © Advanced Micro Devices, Inc., or its affiliates.
+// SPDX-License-Identifier:  MIT
+
+#pragma once
+
+typedef enum
+{
+    HIPDNN_PLUGIN_LOADING_ADDITIVE_UNIQUE = 0,
```
adickin-amd: whats the difference between additive unique and additive?  Can we have the same plugin loaded more than once?

backend/src/logging/logging.hpp
adickin-amd: do we want to be suppressing the cognitive complexity?  Unless its everywhere id rather force devs to no lint the specific pieces of code that are marked too complex.

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -30,32 +30,59 @@ class Engine_plugin_manager : public Plugin_manager_base<Engine_plugin>
...
-    if(!override_plugin_paths.empty())
+    THROW_IF_FALSE(pm_ptr.expired(),
+                   HIPDNN_STATUS_NOT_SUPPORTED,
+                   "hipdnnSetEnginePluginPaths_ext cannot be called with an active handle.");
+                   
+    plugin_config.mode = loading_mode;
+
+    if(plugin_config.mode == HIPDNN_PLUGIN_LOADING_ABSOLUTE)
     {
+        plugin_config.paths = plugin_paths;
```
adickin-amd: what happens if I have a bunch of plugins loaded and then I call this function again with a new path and LOADING_ABSOLUTE?  Do we clear the loaded plugins before moving to a new path?

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -70,10 +97,11 @@ std::shared_ptr<Engine_plugin_resource_manager> Engine_plugin_resource_manager::
...
         if(!pm)
         {
-            auto paths = override_plugin_paths.empty() ? get_default_plugin_paths()
-                                                       : override_plugin_paths;
+            const auto& paths_to_load
+                = plugin_config.paths.empty() ? get_default_plugin_paths() : plugin_config.paths;
+
             pm = std::make_shared<Engine_plugin_manager>();
-            pm->load_plugins(paths);
+            pm->load_plugins(paths_to_load, plugin_config.mode);
```
adickin-amd: if I add plugin paths after the fact, do we update the old resource managers with the new plugins?

backend/src/error.hpp
```diff
@@ -263,6 +264,22 @@ inline const char* hipdnn_get_attribute_name_string(hipdnnBackendAttributeName_t
         return "HIPDNN_ATTR_UNKNOWN";
     }
 }
+
+inline const char* hipdnn_get_plugin_loading_mode_string(hipdnnPluginLoadingMode_t mode)
```
adickin-amd: should look at adding a fmt formatter for this https://github.com/ROCm/hipDNN/blob/develop/sdk/include/hipdnn_sdk/plugin/plugin_data_type_helpers.hpp#L36

Then you can use {} in the log strings

frontend/include/hipdnn_frontend/backend/hipdnn_backend_interface.hpp
```diff
@@ -50,6 +50,13 @@ class Hipdnn_backend_interface
                                                  size_t graph_byte_size)
         = 0;
     virtual void logging_callback_ext(hipdnnSeverity_t severity, const char* msg) = 0;
+
+    // NOLINTBEGIN
+    virtual hipdnnStatus_t set_engine_plugin_paths_ext(size_t num_paths,
```
adickin-amd, line 54: whats wrong with the linting here?

frontend/tests/test_plugin_paths.cpp
adickin-amd: yea, i think ids skip testing the wrapper as we dont want integration tests here.

backend/include/hipdnn_backend.h
```diff
@@ -289,6 +290,21 @@ HIPDNN_BACKEND_EXPORT hipdnnStatus_t hipdnnBackendCreateAndDeserializeGraph_ext(
...
+ *
+ * @param[in] num_paths      The number of paths provided in the plugin_paths array.
+ * @param[in] plugin_paths   An array of path strings, each specifying either a directory to search or a plugin file.
+ *
+ * @retval HIPDNN_STATUS_SUCCESS           All specified plugins were loaded successfully.
+ * @retval HIPDNN_STATUS_BAD_PARAM_NULL_POINTER         Invalid parameter values.
+ * @retval HIPDNN_STATUS_INTERNAL_ERROR    An internal error occurred during plugin loading.
+ */
+HIPDNN_BACKEND_EXPORT hipdnnStatus_t hipdnnSetEnginePluginPaths_ext(
+    size_t num_paths, const char* const plugin_paths[], hipdnnPluginLoadingMode_t loading_mode);
```
averinevg: [R]
```suggestion
    size_t num_paths, const char* const* plugin_paths, hipdnnPluginLoadingMode_t loading_mode);
```

backend/include/hipdnn_backend_plugin_loading_mode.h
```diff
@@ -0,0 +1,11 @@
...
+    HIPDNN_PLUGIN_LOADING_ADDITIVE,
+    HIPDNN_PLUGIN_LOADING_ABSOLUTE,
+} hipdnnPluginLoadingMode_t;
```
averinevg: Consider adding a description.

backend/src/hipdnn_backend.cpp
```diff
@@ -293,4 +294,34 @@ HIPDNN_BACKEND_EXPORT void hipdnnGetLastErrorString(char* message, size_t max_si
 HIPDNN_BACKEND_EXPORT void hipdnnLoggingCallback_ext(hipdnnSeverity_t severity, const char* msg)
 {
     hipdnn_backend::logging::hipdnn_logging_callback(severity, msg);
+}
+
+HIPDNN_BACKEND_EXPORT hipdnnStatus_t hipdnnSetEnginePluginPaths_ext(
+    size_t num_paths, const char* const plugin_paths[], hipdnnPluginLoadingMode_t loading_mode)
```
averinevg: We should consider adding multilingual support. This works on Linux due to UTF-8, but unfortunately it won't work on Windows, which uses UTF-16. Worth discussing.

backend/src/plugin/plugin_core.hpp
```diff
@@ -88,38 +90,35 @@ class Plugin_manager_base
...
-                HIPDNN_LOG_INFO("Plugin info: name={}, version={}, type={}({})",
-                                name,
-                                version,
-                                type,
-                                static_cast<int>(type));
+                if(std::filesystem::is_directory(path))
+                {
+                    scan_directory_for_plugins(path, mode);
+                }
+                else if(std::filesystem::is_regular_file(path))
```
averinevg: This check should be performed inside `Shared_library`.

backend/src/plugin/plugin_core.hpp
```diff
@@ -88,38 +90,35 @@ class Plugin_manager_base
...
+                    load_plugin_from_file(path, mode);
+                }
+                else
+                {
+                    HIPDNN_LOG_WARN("Plugin path is not a file or directory, skipping: {}",
+                                    path.string());
+                }
             }
-            catch(const Hipdnn_exception& e)
+            catch(const std::filesystem::filesystem_error& e)
```
averinevg, line 182: Could you clarify under what conditions this exception might be thrown?

backend/tests/test_engine_plugin.cpp
```diff
@@ -29,10 +35,11 @@ TEST(GPU_EnginePluginTest, LoadPluginsAndExecuteOpGraph)
     Test_engine_plugin_manager plugin_manager;
 
     // Create a list of paths to plugins
-    std::vector<std::filesystem::path> plugin_paths = {"./hipdnn_test_engine_plugin1"};
+    std::vector<std::filesystem::path> plugin_paths
+        = {"./libhipdnn_test_engine_plugin1" + std::string(SHARED_LIB_EXT)};
```
averinevg: The shared object name is `libhipdnn_test_engine_plugin1.so` for Windows and `hipdnn_test_engine_plugin1.dll` for windows. `Shared_object` handles this internally.

backend/tests/test_plugin.cpp
```diff
@@ -45,11 +57,10 @@ TEST(PluginManagerTest, LoadPlugins)
     plugin::Plugin_manager_base<Plugin> plugin_manager;
 
     // Create a list of paths to plugins
-    std::vector<std::filesystem::path> plugin_paths
-        = {"./hipdnn_test_plugin1", "./hipdnn_test_plugin2"};
+    std::vector<std::filesystem::path> plugin_paths = {PLUGIN_PATH1, PLUGIN_PATH2};
```
averinevg: Ditto

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -30,32 +30,59 @@ class Engine_plugin_manager : public Plugin_manager_base<Engine_plugin>
...
+Plugin_loading_config plugin_config;
 std::weak_ptr<Engine_plugin_manager> pm_ptr;
 
 std::vector<std::filesystem::path> get_default_plugin_paths()
 {
-    // This function should return the default plugin paths.
-    // For now, we return an empty vector.
-    // TODO: Implement logic to retrieve default plugin paths.
-    return {};
+    return {"/opt/rocm/lib/hipdnn_plugins/engine/"};
```
averinevg: It might be a good idea to add a `#error` for Windows builds.

backend/src/logging/logging.hpp
SamuelReeder: Surprised me too, but I got fooled. I must have changed something else simultaneously! I think there's an option to suppress cognitive complexity for macro preprocessing so I'll look into that for now.

backend/include/hipdnn_backend_plugin_loading_mode.h
SamuelReeder: Additive loads each plugin all the paths regardless of duplicates. Additive unique does the same but only for unique plugins. So yes, with additive you can have a plugin loaded more than once. This is probably useless in practice but may have use-cases for testing I was thinking.

backend/src/plugin/engine_plugin_resource_manager.cpp
SamuelReeder: Only if the handle has been destroyed. Otherwise it will load the new plugins according to the loading mode.

backend/src/plugin/engine_plugin_resource_manager.cpp
SamuelReeder: If its done before the handle has been created. You can't load plugins on an already active handle since they are loaded when the handle is created.

frontend/include/hipdnn_frontend/backend/hipdnn_backend_interface.hpp
SamuelReeder: C-style array.

backend/src/plugin/plugin_core.hpp
SamuelReeder: I think its good practice to catch file system exceptions since they are prone to throwing.  Could be permission issues, invalid paths, etc.

Edit: in this case `is_directory` won't throw, so should wrap the lower `canonical` function if I keep it.

backend/include/hipdnn_backend_plugin_loading_mode.h
BrianHarrisonAMD: Add _ext to the enum name.

backend/include/hipdnn_backend.h
SamuelReeder: Sure, I'll change all the headers.

backend/src/error.hpp
SamuelReeder: Is it fine to add fmt formatter in the backend? Since the loading mode needs to be defined in the backend since its on the public API. The rest of the formatted plugin types are defined in the sdk.

backend/src/plugin/engine_plugin_resource_manager.cpp
adickin-amd: ah I see the check now that you explained its based on the weak pointer expiration.

backend/src/plugin/engine_plugin_resource_manager.cpp
adickin-amd: thanks for explaining.

backend/src/error.hpp
SamuelReeder: Will add todo :p

backend/tests/test_engine_plugin.cpp
SamuelReeder: Added todo :(

backend/include/hipdnn_backend.h
```diff
@@ -289,6 +290,21 @@ HIPDNN_BACKEND_EXPORT hipdnnStatus_t hipdnnBackendCreateAndDeserializeGraph_ext(
...
+ * This function specifies the set of paths to search for hipDNN plugins.
+ *
+ * @param[in] num_paths      The number of paths provided in the plugin_paths array.
+ * @param[in] plugin_paths   An array of path strings, each specifying either a directory to search or a plugin file.
+ *
+ * @retval HIPDNN_STATUS_SUCCESS           All specified plugins were loaded successfully.
+ * @retval HIPDNN_STATUS_BAD_PARAM_NULL_POINTER         Invalid parameter values.
+ * @retval HIPDNN_STATUS_INTERNAL_ERROR    An internal error occurred during plugin loading.
+ */
+HIPDNN_BACKEND_EXPORT hipdnnStatus_t hipdnnSetEnginePluginPaths_ext(
```
SamuelReeder, line 307: Maybe change to enum after.

backend/src/plugin/shared_library.cpp
```diff
@@ -6,13 +6,72 @@
...
 {
 
+// This function is for querying the address of this library
+void shared_library_anchor() {}
+
+std::filesystem::path Shared_library::get_current_module_directory()
+{
+    std::filesystem::path module_path;
+
+#ifdef _WIN32
```
SamuelReeder, line 21: I have no clue if the windows stuff works :)

SamuelReeder:
One thing I didn't make entirely clear is how relative paths are handled. For default paths they are relative to the shared library location, for custom paths I think they should be relative to the working dir (this is how sqlite does it). I'm cleaning up plugin core a little to make this less ambiguous.

backend/src/plugin/plugin_core.hpp
```diff
@@ -85,41 +97,93 @@ class Plugin_manager_base
...
-    void load_plugins(const std::vector<std::filesystem::path>& plugin_paths)
+    void load_plugins(const std::vector<std::filesystem::path>& custom_paths,
+                      hipdnnPluginLoadingMode_ext_t mode)
     {
-        // Load plugins from the specified paths
-        for(const auto& path : plugin_paths)
+        if(mode == HIPDNN_PLUGIN_LOADING_ABSOLUTE)
         {
-            // TODO Check if the plugin with the same path is already loaded
+            clear_plugins();
```
SamuelReeder: Will remove this function. Not necessary.

Edit: forgot I used it for consecutive plugin loading tests.

backend/include/hipdnn_backend.h
BrianHarrisonAMD: ?

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -23,37 +23,53 @@ namespace plugin
...
+    plugin_config.mode = loading_mode;
+
+    for(const auto& path : plugin_paths)
     {
-        return;
+        if(std::ranges::find(plugin_config.paths, path) == plugin_config.paths.end())
+        {
+            plugin_config.paths.push_back(path);
+        }
     }
```
BrianHarrisonAMD: Shouldn't this depend on loading_mode?

Seems like if it's absolute, then the existing paths should be replaced, and if it's additive it does this logic.
Ill see if this is handled elsewhere.

backend/src/plugin/plugin_core.hpp
```diff
@@ -44,7 +54,9 @@ class Plugin_base
     std::string_view version() const;
     hipdnnPluginType_t type() const;
 
-    hipdnnPluginStatus_t set_logging_callback(hipdnnCallback_t callback);
+    static hipdnnPluginType_t get_class_type();
```
BrianHarrisonAMD: Probably should be `get_plugin_type()`.

backend/src/plugin/plugin_core.hpp
```diff
@@ -85,41 +97,88 @@ class Plugin_manager_base
...
                   "Plugin must be derived from Plugin_base");
 
+protected:
+    explicit Plugin_manager_base(std::vector<std::filesystem::path> default_paths)
+        : _default_plugin_paths(std::move(default_paths))
+    {
+    }
+
+    // TODO: figure out how to ignore cognitive complexity warnings induced by logging macros
+    // NOLINTBEGIN(readability-function-cognitive-complexity)
```
BrianHarrisonAMD, line 107: Lets discuss today. I am increasingly in favor of just disabling this generally.

backend/src/plugin/engine_plugin_resource_manager.cpp
BrianHarrisonAMD: Okay I see that it is handled elsewhere with a split between default, and custom.
We probably need to consider what happens if this is called multiple times.

I.E:
- Called API with Additive path A
- Called API with Additive path B
- Called API with Absolute path C

I think we would still want to clear & replace anytime we see Absolute API call, and only push_back for additive.

backend/src/plugin/plugin_core.hpp
```diff
@@ -85,41 +97,88 @@ class Plugin_manager_base
...
-                                version,
-                                type,
-                                static_cast<int>(type));
+                auto resolved_path = std::filesystem::weakly_canonical(path);
+                load_from_resolved_path(resolved_path);
+            }
+            catch(const std::filesystem::filesystem_error& e)
+            {
+                HIPDNN_LOG_WARN(
+                    "Error resolving custom plugin path '{}': {}", path.string(), e.what());
```
BrianHarrisonAMD: We only want a warning logged here?
Should be an error I think.

backend/src/plugin/plugin_core.hpp
```diff
@@ -85,41 +97,88 @@ class Plugin_manager_base
...
 public:
     virtual ~Plugin_manager_base() = default;
 
-    void load_plugins(const std::vector<std::filesystem::path>& plugin_paths)
+    void load_plugins(const std::vector<std::filesystem::path>& custom_paths,
+                      hipdnnPluginLoadingMode_ext_t mode)
     {
-        // Load plugins from the specified paths
-        for(const auto& path : plugin_paths)
+        if(mode == HIPDNN_PLUGIN_LOADING_ABSOLUTE)
```
BrianHarrisonAMD: I think we should rework this method a bit.

In the end we have N absolute file paths that we want to load plugins from.
They could be absolute paths from a user, or a directory for defaults.
The modes should be used for discovering the filepaths we want to load, and then a final unified loop goes through the filepath collection and loads them. 

Right now we have some splits between the different modes, and I think it would be better to have basically:
```
void load_plugins()
{
std::vector<std::filesystem::path> paths;
if(additive)
{
// Scan the directory for files, swap the paths to be absolute, and push into the paths vector for loading
}

// Clean up custom paths, check that the files exist, and push into the paths vector for loading.

// iterate through the paths vector and load the final list
```

backend/include/hipdnn_backend.h
SamuelReeder: An enum for type of plugin path. ie. engine, heuristics, etc. Just wonder if that conforms to our API more.

backend/src/plugin/engine_plugin_resource_manager.cpp
SamuelReeder: Yes, this should be here.

backend/include/hipdnn_backend.h
BrianHarrisonAMD: Oh I see.

We named it SetEnginePluginPaths, so I guess we would have to add another method later hehe.

backend/src/plugin/plugin_core.hpp
```diff
@@ -130,8 +189,102 @@ class Plugin_manager_base
...
+            const auto type = plugin.type();
+
+            // For now only use engine or unspecified plugin types
+            if(type != Plugin::get_class_type() && type != HIPDNN_PLUGIN_TYPE_UNSPECIFIED)
+            {
+                throw Hipdnn_exception(HIPDNN_STATUS_PLUGIN_ERROR,
+                                       std::string("Plugin type mismatch: expected ")
+                                           + to_string(Plugin::get_class_type()) + ", got "
+                                           + to_string(type));
+            }
```
BrianHarrisonAMD: Probably should just force the type to be specified, and throw otherwise?

If we have plugins that aren't setting it, then we should update those.

backend/src/plugin/plugin_core.hpp
```diff
@@ -130,8 +189,102 @@ class Plugin_manager_base
...
+                {
+                    load_plugin_from_file(path);
+                }
+            }
+        }
+        catch(const std::filesystem::filesystem_error& e)
+        {
+            HIPDNN_LOG_WARN("Error scanning plugin directory {}: {}", dir_path.string(), e.what());
+        }
+    }
```
BrianHarrisonAMD: I think swap this to gather the filepaths, and do all the loading in a single spot.

backend/src/plugin/shared_library.cpp
BrianHarrisonAMD: That's fine for now, we will come back and enable Windows support properly.

backend/src/plugin/shared_library.cpp
```diff
@@ -6,13 +6,73 @@
...
+        if(count > 0)
+        {
+            result[static_cast<size_t>(count)] = '\0';
+            module_path = std::filesystem::path(result.data()).parent_path();
+        }
+        else
+        {
+            throw Hipdnn_exception(HIPDNN_STATUS_INTERNAL_ERROR,
+                                   "Failed to get module path using dladdr and readlink.");
+        }
```
BrianHarrisonAMD: Should we make this a define or something?
If we don't want it to hit this path in production env, then we should strip it in production env.

backend/src/plugin/shared_library.cpp
```diff
@@ -175,5 +248,10 @@ void* Shared_library::get_symbol(std::string_view symbol_name) const
     return symbol;
 }
 
+const std::filesystem::path& Shared_library::final_path() const
+{
+    return _final_path;
+}
```
BrianHarrisonAMD: Probably can call this library path, or just path.

backend/src/plugin/shared_library.cpp
```diff
@@ -6,13 +6,73 @@
...
+            throw Hipdnn_exception(HIPDNN_STATUS_INTERNAL_ERROR, "Failed to get module file name.");
+        }
+    }
+    else
+    {
+        throw Hipdnn_exception(HIPDNN_STATUS_INTERNAL_ERROR, "Failed to get module handle.");
+    }
+#elif defined(__linux__)
+    Dl_info info;
+    if(dladdr(reinterpret_cast<void const*>(&shared_library_anchor), &info) != 0
```
BrianHarrisonAMD: heh interesting.
You can actually use the get_current_module_directory (this function), instead of adding the anchor, but I guess it's fine either way.

backend/tests/test_plugin.cpp
BrianHarrisonAMD, line 1: Add a test for calling multiple times, and making sure the behavior is as expected.

(2 additives are both added)
(1 additive 1 absolute is just absolute)

backend/tests/test_shared_library.cpp
```diff
@@ -81,3 +87,39 @@ TEST(SharedLibraryTest, CallFunction)
...
+    ASSERT_NO_THROW(path = plugin::Shared_library::get_current_module_directory());
+
+    EXPECT_FALSE(path.empty());
+    EXPECT_TRUE(path.is_absolute());
+    EXPECT_TRUE(std::filesystem::is_directory(path));
+
+#ifdef __linux__
+    // Only tests that it works from a statically linked binary
+    EXPECT_TRUE(std::filesystem::exists(path / "hipdnn_backend_tests"));
+#endif
```
BrianHarrisonAMD: Why only __linux__?

Probably fine to leave this without __linux, and we will know to deal with it when we properly add Windows support.

BrianHarrisonAMD:
Looks good to me, with just a few clean-up items.

Thanks @SamuelReeder !

BrianHarrisonAMD:
Oh we should probably add a follow-up task for the public API tests.
Unless we plan on doing that immediately, and this is just a merge for integration purposes.

backend/tests/test_shared_library.cpp
SamuelReeder: I don't know if windows uses a binary or an .exe.

backend/src/plugin/plugin_core.hpp
SamuelReeder: I'm changing it to take a reference to a list and add the files there and then load all from `load_plugins`.

backend/src/plugin/plugin_core.hpp
SamuelReeder: Yeah I'll look through the test plugins. Added this since tests were failing.

backend/src/plugin/shared_library.cpp
SamuelReeder: I'm cool doing that, only did this so the purpose would be clear.

backend/src/plugin/plugin_core.hpp
SamuelReeder: I'm following this, only thing is I won't verify all custom ***file*** paths since they can be platform opaque from the user, and resolved in shared library.

SamuelReeder:
> Oh we should probably add a follow-up task for the public API tests. Unless we plan on doing that immediately, and this is just a merge for integration purposes.

I'll add ticket after. Will push all these updates soon too, was just delayed since I had meeting with Lisa :P

PR #88, opened 2025-07-29
[ALMIOPEN-129]: Fix EngineDetails buffer duplication

adickin-amd:
## Proposed changes

Store the flatbuffer engine details on the handle object and hand out pointers to the stored buffer instead of making fully copies we have to allocate and destroy. 

## Checklist


plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
```diff
@@ -266,7 +266,7 @@ hipdnnPluginStatus_t hipdnnEnginePluginDestroyEngineDetails(hipdnnEnginePluginHa
         throw_if_null(engine_details);
         throw_if_null(engine_details->ptr);
 
-        delete[] static_cast<const uint8_t*>(engine_details->ptr);
+        handle->remove_detached_buffer(engine_details->ptr);
```
BrianHarrisonAMD: I think this is the case, but are we 100% sure the same handle will always be provided to destroy as the one that was provided during create?

I am pretty sure that's guaranteed from the resource manager inside backend, but we should double check.

plugins/miopen_legacy_plugin/tests/engines/test_miopen_engine.cpp
```diff
@@ -140,14 +140,13 @@ TEST(Miopen_engineTest, IsApplicableReturnsFalseIfNoPlanBuilderApplicable)
...
+    hipdnnEnginePluginHandle dummy_handle;
 
     hipdnnPluginConstData_t result;
-    engine.get_details(result);
+    engine.get_details(dummy_handle, result);
 
     hipdnn_plugin::Engine_details_wrapper engine_details(result.ptr, result.size);
     EXPECT_EQ(engine_details.engine_id(), 1);
-
-    delete[] static_cast<const uint8_t*>(result.ptr);
```
BrianHarrisonAMD, line 150: Nice!

BrianHarrisonAMD (approved):
LGTM did we run this through ASAN to ensure memory isn't getting leaked?

plugins/miopen_legacy_plugin/miopen_legacy_plugin.cpp
adickin-amd: its totally feasible for the user to pass a different handle when doing destruction.  There is nothing stopping them from doing so.  All that would happen is a noop on the remove.  I could add warning logging to the failed to remove case so the user has some way of knowing they are doing incorrect things.  

the alternative is for me to pull these engine details buffers out into a global handler that each handle has access to.

BrianHarrisonAMD: I think we should do error logging, but I was pretty sure the design for resource manager maps engine_id -> handle, and it should be guaranteed. Obviously the API doesn't guarantee it, but I meant is it something we can guarantee for the plugin expectations.

adickin-amd: it looks like the engine_plugin_resource_manager does map this. https://github.com/ROCm/hipDNN/blob/develop/backend/src/plugin/engine_plugin_resource_manager.cpp#L197-L204

So our backend should guarantee its using it correctly

mousdahl-amd (approved):

PR #90, opened 2025-07-29
Adding ASAN builds to github pr template

mousdahl-amd:
## Proposed changes

Just a quick update to the PR template to add a requirement for an ASAN build


adickin-amd (approved):
Approved but can we add a small bit of docs for asan?  Maybe just a short blub in the Building.md file to state the best current cmake flags to use for asan

BrianHarrisonAMD (approved):
Yea if you can add a sample command to the docs as well that would be nice, but I think this can go in as is.

SamuelReeder (approved):
I second the above comments.

mousdahl-amd:
The sample cmake command went in on the cpu validation PR.  It should be in buildling.md.

SamuelReeder:
> The sample cmake command went in on the cpu validation PR. It should be in buildling.md.

Ah, the integration branch diverged so I didn't see it.

PR #91, opened 2025-07-29
default miopen plugin build on.

adickin-amd:
## Proposed changes

make plugins build on the CI by default.  They were turned off until the CI image had the miopen devel package installed. This should now be working.

This PR MUST pass all 3 CI checks before merging. 

## Checklist


SamuelReeder (approved):
Does it install the plugins as part of the main install target too?

adickin-amd:
> Does it install the plugins as part of the main install target too?

There is code that will install the plugin .so. As part of the current work I expect it might be revised to a new install location of our choosing.

BrianHarrisonAMD (approved):
Makes sense to me, but seems like there is some build fixes needed based off CI status :P.

mousdahl-amd (approved):

adickin-amd:
> Makes sense to me, but seems like there is some build fixes needed based off CI status :P.

Yep, this was the first test to see if it was functioning.

PR #92, opened 2025-07-29
[ALMIOPEN-135]: Check target now uses gtest intead of ctest by default

adickin-amd:
## Proposed changes

Ctest runner is quite slow when running all the tests so swap over to gtest.  To make it so we dont need to manually update the test list I added a simple mechanism that accumulates the test targets before creating the check target at the end of configure.  The check target can now be used for code coverage.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,39 +62,7 @@ include(cmake/Dependencies.cmake)
 include(cmake/ClangCheck.cmake)
 include(cmake/ClangTidy.cmake)
 include(cmake/Sanitizers.cmake)
-
-hipdnn_add_dependency(GTest v1.16.0)
-
-enable_testing() # Cmake wont discover or run tests without this line
-
-
-set(CHECK_COMMAND_GLOBAL "" CACHE INTERNAL "Accumulated check commands" FORCE)
-set(CHECK_DEPENDS_GLOBAL "" CACHE INTERNAL "Accumulated check depends" FORCE)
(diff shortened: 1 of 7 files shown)


BrianHarrisonAMD (approved):
LGTM assuming we ran check, code_cov, and none of the CI has issues with it.

Nice!

CMakeLists.txt
```diff
@@ -67,8 +67,34 @@ hipdnn_add_dependency(GTest v1.16.0)
 
 enable_testing() # Cmake wont discover or run tests without this line
 
+
+set(CHECK_COMMAND_GLOBAL "" CACHE INTERNAL "Accumulated check commands" FORCE)
+set(CHECK_DEPENDS_GLOBAL "" CACHE INTERNAL "Accumulated check depends" FORCE)
+function(append_test_to_check_target TARGET WORKING_DIR)
```
mousdahl-amd: Would it be possible to move these two functions to something like cmake/Tests.cmake?  Not a huge deal if not.

mousdahl-amd (approved):
Minor nitpick

CMakeLists.txt
adickin-amd: good idea, will do.  I might even take it a bit further and bundle this new function and the gtest_discover_tests into a function so they both get done all in one.

adickin-amd: refactored and extracted some more of the test setup logic into a shared helper function

mousdahl-amd (approved):

PR #93, opened 2025-07-29
Engine plugin resource manager: engine descriptor integration

averinevg:
## Proposed changes

Add engine descriptor integration logic, remove stubbed integration logic, update tests to support it.

### DIFF
--- a/backend/src/CMakeLists.txt
+++ b/backend/src/CMakeLists.txt
@@ -37,6 +37,9 @@ target_include_directories(hipdnn_backend_private
 
 target_compile_options(hipdnn_backend_private PRIVATE ${HIPDNN_WARNING_COMPILE_OPTIONS})
 target_link_libraries(hipdnn_backend_private PUBLIC hip::host hipdnn_sdk)
+if(NOT WIN32)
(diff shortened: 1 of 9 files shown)


backend/src/CMakeLists.txt
```diff
@@ -37,6 +37,9 @@ target_include_directories(hipdnn_backend_private
 
 target_compile_options(hipdnn_backend_private PRIVATE ${HIPDNN_WARNING_COMPILE_OPTIONS})
 target_link_libraries(hipdnn_backend_private PUBLIC hip::host hipdnn_sdk)
+if(NOT WIN32)
+target_link_libraries(hipdnn_backend_private PRIVATE dl)
```
BrianHarrisonAMD, line 41: Was this just missing before?

Surprised this was needed for your changes.

BrianHarrisonAMD (approved):
LGTM assuming the CI passes.

backend/src/CMakeLists.txt
averinevg: I usually always check if the project builds before making any changes. This time it didn't and threw a linker error. I should've added this earlier.

PR #94, opened 2025-07-29
Integration clean up changes

BrianHarrisonAMD:
## Proposed changes

- Standardize flatbuffer creation and access across engine_config & graph.
- Fix unsafe casting to use friend & utility to prevent having unsafe accessor usable generally.

### DIFF
--- a/backend/src/descriptors/backend_descriptor.hpp
+++ b/backend/src/descriptors/backend_descriptor.hpp
@@ -31,6 +31,9 @@ struct Backend_descriptor_interface
 //NOLINTBEGIN(readability-identifier-naming)
 namespace hipdnn_backend
 {
+
+class Mock_descriptor_utility;
+
 template <typename T>
 class hipdnnBackendDescriptorImpl : public Backend_descriptor_interface
 {
@@ -97,15 +100,6 @@ struct hipdnnBackendDescriptor : public Backend_descriptor_interface
             this,
             HIPDNN_STATUS_INTERNAL_ERROR,
             "Failed to cast backend descriptor: Null descriptor provided.");
-    }
-
-    // Unsafe version of as_descriptor that does not perform type checks.
-    // Use with caution, as it can lead to undefined behavior if the types do not match
-    // This is intended for internal mock usage, and not anywhere esle in the codebase.
-    template <typename Child_descriptor>
-    std::shared_ptr<Child_descriptor> as_descriptor_unsafe()
-    {
-        return std::static_pointer_cast<Child_descriptor>(_impl);
     }
 
     bool is_valid();
@@ -177,5 +171,8 @@ struct hipdnnBackendDescriptor : public Backend_descriptor_interface
 
 private:
     std::shared_ptr<Backend_descriptor_interface> _impl;
+
+    // Give access to mock classes for testing purposes only.
+    friend class hipdnn_backend::Mock_descriptor_utility;
 };
 //NOLINTEND(readability-identifier-naming)
--- a/backend/src/descriptors/graph_descriptor.cpp
+++ b/backend/src/descriptors/graph_descriptor.cpp
@@ -108,8 +108,7 @@ hipdnnPluginConstData_t Graph_descriptor::get_serialized_graph() const
         _graph_serialized_buffer = builder.Release();
     }
 
-    return {.ptr = _graph_serialized_buffer.data(),
-            .size = _graph_serialized_buffer.size()};
+    return {.ptr = _graph_serialized_buffer.data(), .size = _graph_serialized_buffer.size()};
 }
 
 hipdnnBackendDescriptorType_t Graph_descriptor::get_static_type()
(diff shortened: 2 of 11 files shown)


backend/src/descriptors/graph_descriptor.cpp
```diff
@@ -90,17 +89,14 @@ void Graph_descriptor::deserialize_graph(const uint8_t* serialized_graph, size_t
                   HIPDNN_STATUS_BAD_PARAM,
                   "Graph_descriptor::deserialize_graph: graph_byte_size is 0");
 
-    // TODO: Consider lazy graph deserialization
+    // TODO: Consider skipping validation entirely, or maybe add an API option to skip it for schema extension cases.
     flatbuffer_utilities::convert_serialized_graph_to_graph(
         serialized_graph, graph_byte_size, _graph);
```
BrianHarrisonAMD, line 94: Add a story for doing this.

backend/src/descriptors/backend_descriptor.hpp
```diff
@@ -177,5 +171,8 @@ struct hipdnnBackendDescriptor : public Backend_descriptor_interface
 
 private:
     std::shared_ptr<Backend_descriptor_interface> _impl;
+
+    // Give access to mock classes for testing purposes only.
+    friend class hipdnn_backend::Mock_descriptor_utility;
```
mousdahl-amd, line 176: TIL that the linker doesn't complain if your friends are imaginary and live in another binary.

mousdahl-amd (approved):
LGTM

backend/src/descriptors/graph_descriptor.cpp
BrianHarrisonAMD: Done

backend/src/descriptors/backend_descriptor.hpp
BrianHarrisonAMD: Even imaginary friends can be helpful (:

averinevg: Not sure this should be here. Is it possible to avoid such tight coupling?

backend/src/descriptors/graph_descriptor.cpp
```diff
@@ -90,17 +89,14 @@ void Graph_descriptor::deserialize_graph(const uint8_t* serialized_graph, size_t
                   HIPDNN_STATUS_BAD_PARAM,
                   "Graph_descriptor::deserialize_graph: graph_byte_size is 0");
 
-    // TODO: Consider lazy graph deserialization
+    // TODO: Consider skipping validation entirely, or maybe add an API option to skip it for schema extension cases.
```
averinevg, line 92: FlatBuffers enables the schema to evolve over time while still maintaining forwards and backwards compatibility with old flatbuffers.
https://flatbuffers.dev/evolution/

BrianHarrisonAMD: Yes I know it will support that, but the problem is that it can't validate a new schema it doesn't understand.
It can validate a new object that only had old values, but if someone wanted to add new nodes it will prevent a new plugin from seeing it since it'll fail validation in the backend.

backend/src/descriptors/backend_descriptor.hpp
BrianHarrisonAMD: We need it so that the mock objects can do a special cast without type safety checks.
This is to remove the unsafe cast option.
Mock's are pretending to be the real objects, and thus they fail the static assert checks put in place to ensure the unpacking logic is sane.

I prefer this way of handling it, since I don't want to have an unsafe version available generally, but if you can think of an alternative to achieve what we need for tests I would be interested.

backend/src/descriptors/graph_descriptor.cpp
averinevg: I think it's worth checking how this works; otherwise, how can one actually use such 'compatibility'?

BrianHarrisonAMD: Yea we do need to do that, but I think it works the way I would want:
- If you are provided something you should understand, then it deserializes and verifies properly.
- If you are provided something you can't understand (it's too new), then it'll fail to verify since you don't know what it is.

Similarly, if an old system generates an old config, then a new system can deserialize and use it (assuming the new field is optional, or has a default).

It allows you to keep handling the objects you do understand regardless of what version they were created with, but would fail in cases where you can't possibly handle it since it has info you don't know how to handle.

BrianHarrisonAMD: Some of this might also be due to the fact that we are using unions for our node types, and you can't unpack a union for a new type you don't understand.

BrianHarrisonAMD: Yea I can see it in the generated file [here](https://github.com/ROCm/hipDNN/blob/develop/sdk/include/hipdnn_sdk/data_objects/graph_generated.h#L643-L676). 

It'll fail to verify new union types it doesn't understand, and the current implementation will fail to verify in the backend if you were to extend the schema, but not rebuild the backend.so. 

That's generally fine, but we will want to add a path that allows you to skip the verification check for cases where someone wants to add a schema extension, and a plugin that supports it. The rest of the system will just fail to unpack it, and should gracefully say it can't support it.

PR #96, opened 2025-07-30
Convert FakePlugin to actual test plugin

mousdahl-amd:
## Proposed changes
- Removed some FakePlugin code from the backend, and the defunct Plugin_manager class that was using it.
- Created a test_good_plugin that generates a real plugin for testing
- This will not fully work until integrated with @averinevg and @SamuelReeder so there will be a follow-on PR to verify this plugin and the tests all work as designed.  I'll run the checks + ASAN at that point.

## Checklist


plugins/test_plugins/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+    target_include_directories(${target_name} 
+        PUBLIC 
+            ${CMAKE_CURRENT_SOURCE_DIR}
+    )
+
+    target_link_libraries(${target_name} 
+        PUBLIC 
+            hipdnn_sdk
+            hip::device
+            hip::host
```
SamuelReeder, line 34: Do we need these even though the test plugin source code does not use them?

plugins/test_plugins/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+    target_compile_options(${target_name} 
+        PRIVATE 
+            ${HIPDNN_WARNING_COMPILE_OPTIONS}
+    )
+
+    add_dependencies(check ${target_name})
+endfunction()
+
+# test_good_plugin is a well-behaved plugin that returns a happy path
+add_test_plugin(test_good_plugin test_good_plugin.cpp)
```
mousdahl-amd, line 46: We should be able to add additional test plugins with one line here.

SamuelReeder, line 46: Is the install target going to be deferred to plugin loading integration? (follow-up PR)

plugins/test_plugins/CMakeLists.txt
mousdahl-amd: I threw a little device buffer copy kernel in for fun.  If it's not necessary, I can try to remove it, but we still need hipStream_t, and I'm not sure if that'll work okay without linking.

plugins/test_plugins/CMakeLists.txt
mousdahl-amd: I'm not going to install these, they're explicitly for testing.  I'll use your new plugin path stuff in the tests.

mousdahl-amd: Or, at least, I wasn't *planning* on installing them.  Is there a reason I should?

plugins/test_plugins/CMakeLists.txt
SamuelReeder: Huh, I didn't know you could use those without including hip_runtime_api.h.

plugins/test_plugins/test_good_plugin.cpp
```diff
@@ -0,0 +1,421 @@
...
+{
+    const auto tid = blockIdx.x * blockDim.x + threadIdx.x;
+    if(tid < size)
+    {
+        output[tid] = input[tid];
+    }
+}
+
+// Run the kernel
+void run_engine(const uint32_t* input, uint32_t* output, uint32_t size)
```
SamuelReeder: Nit: I assume in a real plugin we would want to pass the stream here too.

SamuelReeder (approved):
LGTM

plugins/test_plugins/CMakeLists.txt
SamuelReeder: I have logic for finding the default plugins installed relative to the backend.so, so only if we want to test that further. It would be installed in ./hipdnn_plugins/engines/ or maybe (./hipdnn_plugins/testing/) relative to backend.

plugins/test_plugins/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+
+    target_include_directories(${target_name} 
+        PUBLIC 
+            ${CMAKE_CURRENT_SOURCE_DIR}
+    )
+
+    target_link_libraries(${target_name} 
+        PUBLIC 
+            hipdnn_sdk
+            hip::device
```
averinevg, line 33: `hip::device` can be removed

mousdahl-amd: Tried it, it won't compile the kernel in test_good_plugin.cpp without it.

averinevg: Weird. For the plugin from which this kernel was taken, this is not required.

PR #99, opened 2025-07-31
Make tensor IDs in the flatbuffer schemas strongly typed

jerehartAMD:
## Proposed changes

Updates the schema to use a strongly typed TensorID class as a wrapper around "long" instead of just using longs. The goal of this is to make it clear which values are supposed to be tensorIDs and which are supposed to be 64 bit integers, as well as provide extra type safety

This change has fairly widespread changes, there are a couple that may not be ideal. I'm doing a self-review to highlight some I noticed. I want to get some opinions before I continue down this path

## Checklist

### DIFF
--- a/frontend/include/hipdnn_frontend/attributes/batchnorm_attributes.hpp
+++ b/frontend/include/hipdnn_frontend/attributes/batchnorm_attributes.hpp
@@ -36,6 +36,7 @@ public:
         next_running_variance = 4
     };
 
+    // TODO: This can just be an array, right?
     std::unordered_map<input_names, std::shared_ptr<Tensor_attributes>> inputs;
     std::unordered_map<output_names, std::shared_ptr<Tensor_attributes>> outputs;
     std::vector<std::shared_ptr<Tensor_attributes>> peer_stats;
@@ -159,12 +160,12 @@ public:
     flatbuffers::Offset<hipdnn_sdk::data_objects::BatchnormAttributes>
         pack_attributes(flatbuffers::FlatBufferBuilder& builder) const
     {
-        auto peer_stats_vector = std::vector<int64_t>{};
+        auto peer_stats_vector = std::vector<hipdnn_sdk::data_objects::TensorID>{};
         for(const auto& peer_stat : peer_stats)
         {
             if(peer_stat)
             {
-                peer_stats_vector.emplace_back(peer_stat->get_uid());
+                peer_stats_vector.emplace_back(peer_stat->get_id());
             }
         }
 
@@ -178,24 +179,19 @@ public:
 
         return hipdnn_sdk::data_objects::CreateBatchnormAttributesDirect(
             builder,
-            get_x()->get_uid(),
-            get_scale()->get_uid(),
-            get_bias()->get_uid(),
-            get_epsilon()->get_uid(),
+            &get_x()->get_id(),
+            &get_scale()->get_id(),
+            &get_bias()->get_id(),
+            &get_epsilon()->get_id(),
             &peer_stats_vector,
-            prev_running_mean ? flatbuffers::Optional<int64_t>(prev_running_mean->get_uid())
-                              : flatbuffers::nullopt,
-            prev_running_variance ? flatbuffers::Optional<int64_t>(prev_running_variance->get_uid())
-                                  : flatbuffers::nullopt,
-            momentum ? flatbuffers::Optional<int64_t>(momentum->get_uid()) : flatbuffers::nullopt,
-            get_y()->get_uid(),
-            mean ? flatbuffers::Optional<int64_t>(mean->get_uid()) : flatbuffers::nullopt,
-            inv_variance ? flatbuffers::Optional<int64_t>(inv_variance->get_uid())
-                         : flatbuffers::nullopt,
-            next_running_mean ? flatbuffers::Optional<int64_t>(next_running_mean->get_uid())
-                              : flatbuffers::nullopt,
-            next_running_variance ? flatbuffers::Optional<int64_t>(next_running_variance->get_uid())
-                                  : flatbuffers::nullopt);
+            &prev_running_mean->get_id(),
+            &prev_running_variance->get_id(),
+            &momentum->get_id(),
+            &get_y()->get_id(),
+            &mean->get_id(),
+            &inv_variance->get_id(),
+            &next_running_mean->get_id(),
+            &next_running_variance->get_id());
     }
 
 private:
--- a/frontend/include/hipdnn_frontend/attributes/tensor_attributes.hpp
+++ b/frontend/include/hipdnn_frontend/attributes/tensor_attributes.hpp
@@ -58,7 +58,12 @@ public:
 
     int64_t get_uid() const
     {
-        return _uid;
+        return _id.value();
+    }
+
+    const hipdnn_sdk::data_objects::TensorID& get_id() const
+    {
+        return _id;
     }
 
     const std::string& get_name() const
@@ -98,13 +103,13 @@ public:
 
     bool has_uid() const
     {
-        return _uid_set;
+        return _id_set;
     }
 
     Tensor_attributes& set_uid(int64_t uid)
     {
-        _uid = uid;
-        _uid_set = true;
+        _id = uid;
+        _id_set = true;
         return *this;
     }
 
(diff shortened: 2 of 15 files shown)


frontend/include/hipdnn_frontend/attributes/tensor_attributes.hpp
```diff
@@ -224,8 +229,8 @@ class Tensor_attributes
     }
 
 private:
-    int64_t _uid = 0;
-    bool _uid_set = false;
+    hipdnn_sdk::data_objects::TensorID _id;
```
jerehartAMD, line 232: This puts a flatbuffer generated datatype inside our classes. It's private, but there's an accessor for it. Do we want to expose this? I'm leaning no?

The reason I included it in the class is so that we can pass a pointer to the id when we generate the flatbuffers. However, we could make our own strongly typed class not tied to `hipdnn_sdk::data_objects::TensorID` and reinterpret_cast the pointer. Since they're trivially copyable this isn't UB

sdk/include/hipdnn_sdk/test_utilities/flatbuffer_graph_test_utils.hpp
```diff
@@ -40,24 +45,24 @@ inline flatbuffers::FlatBufferBuilder
         tensor_attributes;
 
     tensor_attributes.push_back(hipdnn_sdk::data_objects::CreateTensorAttributesDirect(
-        builder, 1, "x", hipdnn_sdk::data_objects::DataType_FLOAT, &strides, &dims));
+        builder, make_tensor_id(1).get(), "x", hipdnn_sdk::data_objects::DataType_FLOAT, &strides, &dims));
```
jerehartAMD, line 48: This is the clunkiest part. We need to pass in a struct for tensors now, and this API requires structs to be passed by pointer. Since `CreateXDirect` functions immediately use the value pointed to, we can create a unique ptr which will go out of scope after the `CreateXDirect` function is completed. It's definitely not as clean as it was before, but this is functional.

sdk/include/hipdnn_sdk/data_objects/batchnorm_attributes_generated.h
```diff
@@ -25,19 +27,23 @@ bool operator!=(const BatchnormAttributesT &lhs, const BatchnormAttributesT &rhs
...
+  std::unique_ptr<hipdnn_sdk::data_objects::TensorID> epsilon{};
+  std::vector<hipdnn_sdk::data_objects::TensorID> peer_stats{};
+  std::unique_ptr<hipdnn_sdk::data_objects::TensorID> prev_running_mean{};
+  std::unique_ptr<hipdnn_sdk::data_objects::TensorID> prev_running_variance{};
+  std::unique_ptr<hipdnn_sdk::data_objects::TensorID> momentum{};
+  std::unique_ptr<hipdnn_sdk::data_objects::TensorID> y{};
+  std::unique_ptr<hipdnn_sdk::data_objects::TensorID> mean{};
+  std::unique_ptr<hipdnn_sdk::data_objects::TensorID> inv_variance{};
+  std::unique_ptr<hipdnn_sdk::data_objects::TensorID> next_running_mean{};
+  std::unique_ptr<hipdnn_sdk::data_objects::TensorID> next_running_variance{};
```
jerehartAMD, line 42: As a result of this, classes like this also become a lot more complicated, and they are now full of unique_ptrs instead of scalar values. I have a thread that I'm going to go down tomorrow though, it looks like if you use enums instead of structs we might be able to solve this issue, but then you must cast it to and from the enum, which feels like it makes us lose most of the benefit

edit: And is also probably UB

jerehartAMD:
I'm currently leaning towards not making things strongly typed on the schema side (and renaming the schema fields to make them clear), but one happy medium we could have is make the TensorIDs strongly typed on the C++ side. Won't get us all the benefits of having it strongly typed all the way through, but I think it'd matter more there anyways

sdk/schemas/batchnorm_attributes.fbs
```diff
@@ -1,20 +1,22 @@
...
-    next_running_mean: long = null;
-    next_running_variance: long = null;
+    x: TensorID;
+    scale: TensorID;
+    bias: TensorID;
+    epsilon: TensorID;
+    peer_stats: [TensorID];
+    prev_running_mean: TensorID;
+    prev_running_variance: TensorID;
+    momentum: TensorID;
```
BrianHarrisonAMD, line 16: These still need to be optional I believe (`long = null`).
How does that work now?

sdk/schemas/batchnorm_attributes.fbs
```diff
@@ -1,20 +1,22 @@
...
+    y: TensorID;
+    mean: TensorID;
+    inv_variance: TensorID;
+    next_running_mean: TensorID;
+    next_running_variance: TensorID;
```
BrianHarrisonAMD, line 21: Same with these.

sdk/include/hipdnn_sdk/test_utilities/flatbuffer_graph_test_utils.hpp
BrianHarrisonAMD: Ah, that's quite a bit of extra packing effort just to work around the clunkiness :(.

frontend/include/hipdnn_frontend/attributes/batchnorm_attributes.hpp
```diff
@@ -178,24 +179,19 @@ class Batchnorm_attributes : public AttributesCRTP<Batchnorm_attributes>
...
-            next_running_variance ? flatbuffers::Optional<int64_t>(next_running_variance->get_uid())
-                                  : flatbuffers::nullopt);
+            &prev_running_mean->get_id(),
+            &prev_running_variance->get_id(),
+            &momentum->get_id(),
+            &get_y()->get_id(),
+            &mean->get_id(),
+            &inv_variance->get_id(),
+            &next_running_mean->get_id(),
+            &next_running_variance->get_id());
```
BrianHarrisonAMD, line 194: We still need some way to flag optional, and since it's a public API that allows setting uid's manually, we can't really use a specific value to flag nullopt. 

I think we need to add the optional logic back.

BrianHarrisonAMD:
Assuming we can still add optional logic similar to before, the actual changes in the frontend code seem fine to me.
I am not super excited about how the packing has to work with unique_ptr's, and that's a bit unfortunate. 

I guess id be interested to see what other people think.
I remember going down this route and bailing since I got to about this same point.
I would probably lean towards renaming the fields to make it more clear, and living with long's instead of structs, but if someone really prefers the strong types I am alright with it.

sdk/include/hipdnn_sdk/test_utilities/flatbuffer_graph_test_utils.hpp
mousdahl-amd: Does marking it as `required` in the schema have any impact?

sdk/schemas/batchnorm_attributes.fbs
jerehartAMD: From my understanding `long = null` is essentially a signal to turn the long into an std::optional<long>. All non scalar fields are automatically nullable and default to null, so we don't need them here. However, we may need `(required)` for the others when we use structs (it remains an nullable value, but will throw if you don't set it)

frontend/include/hipdnn_frontend/attributes/batchnorm_attributes.hpp
jerehartAMD: I got my pointer chain a little mixed up here, for some reason I was assuming the get_id would be the nullptr, but it must be the shared_ptr itself, this will absolutely be a nullptr dereference

sdk/include/hipdnn_sdk/test_utilities/flatbuffer_graph_test_utils.hpp
jerehartAMD: Unfortunately not on the layout or access pattern, `required` will make it throw though if it's not set

BrianHarrisonAMD: I guess if we keep this strong-type version, then yea we need to flip all these to required.
When they were just long's the `long=null` was how you make it optional, and otherwise it has to be set to a value.

jerehartAMD:
This work may be revisited, but for now we're renaming these fields

PR #101, opened 2025-07-31
Engine plugin resource manager: execution plan descriptor integration

averinevg:
## Proposed changes

Add execution plan descriptor integration logic, update tests to support it.


SamuelReeder (approved):
LGTM

BrianHarrisonAMD (approved):

PR #103, opened 2025-07-31
Change names of tensor id fields in schemas to indicate their type

jerehartAMD:
## Proposed changes

To make the intended values of the tensor id fields clear, this PR renames them to include "\_tid" at the end. I suspect this might be too brief, but adding \_tensor\_id to every field caused its name to get lost amongst the "typing". There might be a happy medium somewhere inbetween, and changing the suffix is trivial

## Checklist


sdk/schemas/batchnorm_attributes.fbs
```diff
@@ -4,17 +4,17 @@
...
-    peer_stats: [long];
-    prev_running_mean: long = null;
-    prev_running_variance: long = null;
-    momentum: long = null;
-    y: long;
-    mean: long = null;
-    inv_variance: long = null;
-    next_running_mean: long = null;
-    next_running_variance: long = null;
+    x_tid: long;
```
adickin-amd: This is all looking good to me.  My only thoughts are we might want to make sure these two names match.  Either rename the struct to have tid or rename the flatbuffers to be uid, https://github.com/ROCm/hipDNN/blob/develop/sdk/include/hipdnn_sdk/plugin/plugin_api_data_types.h#L70-L74

jerehartAMD: I can definitely see a case for that, especially given that uid is an established term, and tid requires some explanation. The only drawback I see is that it doesn't indicate that x is a tensor

adickin-amd: true, can we go more verbose, x_tensor_uid?

jerehartAMD: It's definitely an option, I'll make a poll

docs/Building.md
```diff
@@ -10,10 +10,9 @@ hipDNN.  At the time of writing this readme, the following bare minimum system d
     - hip-devel
 - Python3
 - Cmake 3.25.2 or later
-- Ninja 1.12.1 or later
-    - Ninja is the default build system for hipDNN.
-    - If you dont set the environment variable `CMAKE_GENERATOR` to `Ninja`, cmake will default to using make.
-- 
+- Ninja 1.12.1 or later (Optional)
+    - Ninja is the preferred build system for hipDNN
+    - Cmake will default to using make unless you set CMAKE_GENERATOR to Ninja. This environment variable is already set in the development docker containers
```
BrianHarrisonAMD, line 15: 👍

BrianHarrisonAMD (approved):
LGTM assuming the CI passes.
Merge whenever you are happy with the names.

PR #110, opened 2025-08-06
Backend platform utilities

mousdahl-amd:
## Proposed changes

- Created a backed platform_utils header and platform-specific implementation cpp files

## Checklist


backend/src/platform_utils.hpp
```diff
@@ -0,0 +1,41 @@
+// Copyright © Advanced Micro Devices, Inc., or its affiliates.
+// SPDX-License-Identifier:  MIT
+
+#pragma once
+
+#include <filesystem>
+#include <string>
+
+namespace hipdnn_backend::platform_utils
+{
```
BrianHarrisonAMD, line 10: I am fine with this style, but I think in other places we separated it out.
Surprised tidy doesn't force it one way or another.

Might want to double check this file is getting picked up.

backend/tests/test_plugin.cpp
```diff
@@ -54,24 +55,16 @@ void dummy_callback(hipdnnSeverity_t sev, const char* msg)
...
+const std::string PLUGIN_PATH2 = std::filesystem::path(".") /= PLUGIN_NAME2;
 
-const std::string FULL_PLUGIN_PATH1
-    = std::string("./") + LIB_PREFIX + PLUGIN_NAME1 + SHARED_LIB_EXT;
-const std::string FULL_PLUGIN_PATH2
-    = std::string("./") + LIB_PREFIX + PLUGIN_NAME2 + SHARED_LIB_EXT;
+const std::string FULL_PLUGIN_PATH1 = std::filesystem::path(".")
+    /= platform_utils::get_library_name(PLUGIN_NAME1.c_str());
+const std::string FULL_PLUGIN_PATH2 = std::filesystem::path(".")
+    /= platform_utils::get_library_name(PLUGIN_NAME2.c_str());
```
BrianHarrisonAMD: Can we make this std::fileystem::path's?

I think they get converted later anyways I believe.

BrianHarrisonAMD (approved):
LGTM

Minor comments that you can decide on.

backend/src/platform_utils.hpp
mousdahl-amd: I'll double check that tidy is hitting this for sure, but I've seen a mixture among the files.  I've got a namespace cleanup story in the backlog, so maybe we can pick something and normalize on it.

backend/src/platform_utils.windows.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+#ifdef _WIN32
+
+#include "hipdnn_exception.hpp"
+#include <windows.h>
+
+namespace hipdnn_backend::platform_utils
+{
+
+inline std::filesystem::path get_current_module_directory()
```
SamuelReeder: I think this should be non-inline, like the header.

Edit: or... is this needed for windows?

mousdahl-amd: Nope, just a mistake!  I'm going to be doing most of the Windows cleanup in the next story.

backend/src/platform_utils.hpp
mousdahl-amd: Tidy is hitting the file, the only check that cares about namespaces is `modernize-concat-nested-namespaces` which is explicitly turned off.

SamuelReeder (approved):
LGTM

PR #111, opened 2025-08-07
Forcing include order for the validators when using half_utils.hpp or bfloat16_utils.hpp

mousdahl-amd:
## Proposed changes
- If using the validators in the context of an AMD HIP application, it will ensure that the validators include the half and bfloat16 types so that the custom versions of abs() and max() can be used, rather than falling back to the float versions because the headers weren't included in the correct order, thereby causing implicit cast warnings.

## Checklist

### DIFF
--- a/sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
+++ b/sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
@@ -7,6 +7,12 @@
 #include <numeric>
 #include <vector>
 
+#if defined(__HIP_PLATFORM_AMD__)
+// Need these for the half and bfloat16 types
+#include <hipdnn_sdk/utilities/half_utils.hpp>
+#include <hipdnn_sdk/utilities/hip_bfloat16_utils.hpp>
+#endif
+
 namespace hipdnn_sdk
 {
 namespace reference_test_utilities
--- a/plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
+++ b/plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
@@ -4,15 +4,14 @@
 #include <gtest/gtest.h>
 #include <numeric>
 
-#include <hipdnn_sdk/utilities/half_utils.hpp>
-#include <hipdnn_sdk/utilities/hip_bfloat16_utils.hpp>
-
 #include <hipdnn_sdk/plugin/engine_plugin_api.h>
 #include <hipdnn_sdk/plugin/plugin_api_data_types.h>
 #include <hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp>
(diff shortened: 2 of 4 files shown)


BrianHarrisonAMD (approved):
LGTM.

Did we just need the includes added to the reference implementation to fix the order?

mousdahl-amd:
Yeah, it just guarantees that we've got the correct std::fabs if we're using half or bfloat16.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
```diff
@@ -7,6 +7,12 @@
 #include <numeric>
 #include <vector>
 
+#if defined(__HIP_PLATFORM_AMD__)
```
SamuelReeder, line 10: To be sure I understand: in calling code I can stop linking hip::device and use this compile definition instead? And would this be good in other headers too? Like [`hipdnn_frontend/types.hpp`](https://github.com/ROCm/hipDNN/blob/develop/frontend/include/hipdnn_frontend/types.hpp#L47) where if I recall correctly I had to link hip::device to use that header.

SamuelReeder: Actually I think we might still need to link hip::device anyways for the special datatypes. Guess it's not a big deal.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
```diff
@@ -7,6 +7,12 @@
...
+// Need these for the half and bfloat16 types
+#include <hipdnn_sdk/utilities/half_utils.hpp>
```
adickin-amd, line 12: is every reference implementation going to end up in this file?  I could see the reference implementations getting split up by operation once we have enough of them and this file gets large.  Will each new file just need to know these need to bet at the top?

mousdahl-amd: I think that was the original intent, but then you said those words in that order, and now I'm sad.  You're right, this could get pretty big.  A split by operation makes sense to me.

BrianHarrisonAMD: Yea we probably should split them by namespace and OP over time. 
I was hoping we could keep all the definitions together, but this is header only, and can't split across multiple cpp's.

We might need to have some design where you can specify op providers, and then a reference implementation container that gives you access to each reference implementation provider. Especially since the fusion piece will be generic and across ops, we probably need to have some parent that contains all the definitions in one place, but the providers just implement their specific piece.

mousdahl-amd: This sounds like a "tackle it during fusion" sort of effort.  I'll leave it as is for now.

PR #113, opened 2025-08-07
[ALMIOPEN-164] API plugin loading tests and check targets

SamuelReeder:
## Proposed changes

Added plugin loading tests at the backend API level, mainly to check that plugins are loaded relative to .so correctly and that custom plugins are handled correctly in these cases.

In addition, this adds new `unit-check` and `integration-check` targets to improve quality of life.

## Checklist


cmake/Tests.cmake
```diff
@@ -28,6 +68,44 @@ add_custom_target(
...
+    message(STATUS "Created integration-check target")
+else()
+    add_custom_target(
+        integration-check
+        COMMAND ${CMAKE_COMMAND} -E echo "No integration tests found"
+        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+        VERBATIM)
+    message(STATUS "Created empty integration-check target (no integration tests found)")
+endif()
+endfunction()
```
BrianHarrisonAMD: ninja check still does everything, and now we have ninja unit-check, and ninja integration-check?

SamuelReeder: Yes

backend/src/hipdnn_backend.cpp
```diff
@@ -294,7 +294,10 @@ HIPDNN_BACKEND_EXPORT hipdnnStatus_t hipdnnSetEnginePluginPaths_ext(
                   hipdnn_backend::hipdnn_get_plugin_loading_mode_string(loading_mode));
 
     return hipdnn_backend::try_catch([&, api_name = __func__] {
-        throw_if_null(plugin_paths);
+        if(num_paths > 0)
+        {
+            throw_if_null(plugin_paths);
+        }
```
BrianHarrisonAMD, line 300: I guess the desire is to use this to unload?

backend/src/hipdnn_backend.cpp
```diff
@@ -307,11 +310,33 @@ HIPDNN_BACKEND_EXPORT hipdnnStatus_t hipdnnSetEnginePluginPaths_ext(
...
+                  static_cast<void*>(num_plugin_paths),
+                  static_cast<const void*>(plugin_paths),
+                  static_cast<void*>(max_string_size));
+
+    return hipdnn_backend::try_catch([&, api_name = __func__] {
+        throw_if_null(num_plugin_paths);
+        throw_if_null(max_string_size);
+
+        plugin::Engine_plugin_resource_manager::get_loaded_plugin_files(
+            num_plugin_paths, plugin_paths, max_string_size);
```
BrianHarrisonAMD: Should we have a similar mechanism as descriptors for getting the count & max size?

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -73,6 +74,54 @@ std::set<std::filesystem::path> Engine_plugin_resource_manager::get_plugin_paths
...
+                                                             char** plugin_paths,
+                                                             size_t* max_string_size)
+{
+    auto pm = pm_ptr.lock();
+    if(!pm)
+    {
+        *num_plugins = 0;
+        *max_string_size = 0;
+        return;
+    }
```
BrianHarrisonAMD: This is a little awkward since you have to create a handle for this to work.
I think either the API should require a handle, or we should be able to grab the paths of which plugins we want to load, and return that.

SamuelReeder: I think I'll give it the handle, since it's more useful to get the plugins that are loaded when handle is created.

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -73,6 +74,54 @@ std::set<std::filesystem::path> Engine_plugin_resource_manager::get_plugin_paths
...
+    {
+        if(plugin_paths[i] == nullptr)
+        {
+            throw Hipdnn_exception(HIPDNN_STATUS_BAD_PARAM, "A plugin path string buffer is null.");
+        }
+        hipdnn::sdk::utilities::copy_max_size_with_null_terminator(
+            plugin_paths[i], paths_vec[i].c_str(), *max_string_size);
+    }
+    *num_plugins = paths_vec.size();
+}
```
BrianHarrisonAMD: We should be able to allocate a char buffer that is the length needed, and then have some format we communicate through documentation.

Basically, we should get a char** passed in, and then we malloc a buffer big enough to store all the paths, and return it back to them with a separator between items. Not even sure we need a count in this case. 

We would need to also indicate it's the callers responsibility to free the returned memory, but this is also mostly just for debugging / testing purposes, and I think that's fine.

Or we could do it like this, but then we need to have some limit on the path I guess.

tests/backend/test_util.cpp
```diff
@@ -299,4 +299,46 @@ void extract_tensor_info_from_graph(
...
+    {
+        return {};
+    }
+
+    std::vector<std::vector<char>> path_buffers(num_plugins, std::vector<char>(max_path_length));
+    std::vector<char*> plugin_paths_c(num_plugins);
+    for(size_t i = 0; i < num_plugins; ++i)
+    {
+        plugin_paths_c[i] = path_buffers[i].data();
+    }
```
BrianHarrisonAMD: Hmm okay this does exist, I must have missed how it can get this far.

backend/src/hipdnn_backend.cpp
BrianHarrisonAMD: I guess we do looking at how it's used.

backend/src/plugin/engine_plugin_resource_manager.cpp
SamuelReeder: Ok that sounds good, I was just hesitant to make user have to free obscure memory.

BrianHarrisonAMD: I didn't realize you could get the count / max, I thought it was throwing on null paths as well, but it wasn't.

This is probably fine since there is a way to know how big to make the strings.

backend/src/plugin/engine_plugin_resource_manager.cpp
BrianHarrisonAMD: Yea agreed.

BrianHarrisonAMD:
For the check changes, please make sure all tests are still accounted for.

I.E in develop it reports X test cases, and in the branch it's still X + newly added tests, and the list is the same.

I imagine you already did, but just want to make sure we explicitly checked that it's still giving the same list since this is used for CI, and if it silently hid tests it could be hard to track down.

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -73,6 +74,54 @@ std::set<std::filesystem::path> Engine_plugin_resource_manager::get_plugin_paths
...
+    for(size_t i = 0; i < paths_vec.size(); ++i)
+    {
+        if(plugin_paths[i] == nullptr)
+        {
+            throw Hipdnn_exception(HIPDNN_STATUS_BAD_PARAM, "A plugin path string buffer is null.");
+        }
+        hipdnn::sdk::utilities::copy_max_size_with_null_terminator(
+            plugin_paths[i], paths_vec[i].c_str(), *max_string_size);
+    }
+    *num_plugins = paths_vec.size();
```
BrianHarrisonAMD: No reason to set this since it's guaranteed above.

tests/backend/test_backend_public_functions.cpp
```diff
@@ -184,7 +187,7 @@ TEST(HipDNNBackendTest, SetPluginPathsExt_FailsOnNullPointer)
     EXPECT_EQ(status, HIPDNN_STATUS_BAD_PARAM_NULL_POINTER);
 
     status = hipdnnSetEnginePluginPaths_ext(0, nullptr, HIPDNN_PLUGIN_LOADING_ABSOLUTE);
-    EXPECT_EQ(status, HIPDNN_STATUS_BAD_PARAM_NULL_POINTER);
+    EXPECT_EQ(status, HIPDNN_STATUS_SUCCESS);
```
BrianHarrisonAMD, line 197: This test should be renamed now, and also include verifying that no plugins are loaded after doing this and creating a handle.

tests/backend/test_backend_public_functions.cpp
```diff
@@ -197,23 +200,76 @@ TEST(HipDNNBackendTest, SetPluginPathsExt_FailsOnNullStringInList)
...
-//     // Can also test by manually changing default path
-//     const std::array<const char*, 1> paths = {"../../tests/test_plugins/"};
+class Hipdnn_backend_default_plugin_test : public ::testing::Test
+{
+protected:
+    void SetUp() override
+    {
+        hipdnnSetEnginePluginPaths_ext(0, nullptr, HIPDNN_PLUGIN_LOADING_ADDITIVE);
+        const std::filesystem::path dest_dir = "../../backend/src/hipdnn_plugins/engines/";
+        const std::filesystem::path source_file = "../test_plugins/libtest_good_plugin.so";
```
BrianHarrisonAMD: Platform specific issues here as well.
Can hopefully use the platform_utils that @mousdahl-amd is adding to fix this.

tests/backend/test_backend_public_functions.cpp
```diff
@@ -197,23 +200,76 @@ TEST(HipDNNBackendTest, SetPluginPathsExt_FailsOnNullStringInList)
...
+
+        std::filesystem::create_directories(dest_dir);
+        std::filesystem::copy(source_file,
+                              dest_dir / "libtest_good_default_plugin.so",
+                              std::filesystem::copy_options::overwrite_existing);
```
BrianHarrisonAMD: Ah okay I see what you were mentioning with the paths.
I think we probably need to build a plugin that goes into the default directory location for testing this.
This is pretty fragile.

BrianHarrisonAMD: Maybe a separate default plugin executable can be made that builds into the plugin directory, and it's not the good plugin.

BrianHarrisonAMD:
Generally looks fine, but the plugin files and paths need some changes.
We likely need to integrate the platform utils to deal with that aspect, and then rethink how the default plugin loading should work.

Additionally, I think we should start being more careful to split out the PRs.
I know I am guilty of this as well, but we should start being more strict.

SamuelReeder:
> Generally looks fine, but the plugin files and paths need some changes. We likely need to integrate the platform utils to deal with that aspect, and then rethink how the default plugin loading should work.
> 
> Additionally, I think we should start being more careful to split out the PRs. I know I am guilty of this as well, but we should start being more strict.

I agree with the splitting; I'm definitely guilty. 

I will revise all the above issues when we merge platform utils PR.

plugins/miopen_legacy_plugin/integration_tests/batchnorm_fwd_inference_integration_test.cpp
```diff
@@ -161,19 +161,24 @@ class Batchnorm_forward_inference_integration_test
...
+            = std::make_shared<Tensor_attributes>(std::move(inv_variance_attr));
 
-        auto scale_attr = make_tensor_attributes("scale", intermediate_data_type, graph_tensor_bundle.scale_tensor);
+        auto scale_attr = make_tensor_attributes(
+            "scale", intermediate_data_type, graph_tensor_bundle.scale_tensor);
         scale_attr.set_uid(uid++);
         auto scale_tensor_attr = std::make_shared<Tensor_attributes>(std::move(scale_attr));
 
-        auto bias_attr = make_tensor_attributes("bias", intermediate_data_type, graph_tensor_bundle.bias_tensor);
+        auto bias_attr = make_tensor_attributes(
```
SamuelReeder: Not sure why this is being formatted now.

tests/backend/test_backend_public_functions.cpp
```diff
@@ -1,10 +1,16 @@
 // Copyright © Advanced Micro Devices, Inc., or its affiliates.
 // SPDX-License-Identifier:  MIT
 
+#include "../../backend/src/platform_utils.hpp"
+#include "../test_plugins/test_plugin_constants.hpp"
 #include "hipdnn_backend.h"
+#include "test_util.hpp"
 #include <array>
+#include <filesystem>
```
SamuelReeder: Will remove this include.

tests/backend/test_util.cpp
```diff
@@ -287,4 +287,47 @@ void extract_tensor_info_from_graph(
...
+
+    status = hipdnnGetLoadedEnginePluginPaths_ext(
+        handle, &num_plugins, plugin_paths_c.data(), &max_path_length);
+    if(status != HIPDNN_STATUS_SUCCESS)
+    {
+        return {};
+    }
```
BrianHarrisonAMD: Should status != success be a throw?

I assume we shouldn't be calling this helper for failure cases, and non-success should be treated differently than empty.

tests/backend/test_util.cpp
```diff
@@ -287,4 +287,47 @@ void extract_tensor_info_from_graph(
...
+}
+
+bool is_plugin_loaded(const std::vector<std::string>& loaded_plugins,
+                      const std::string& plugin_name)
+{
+    return std::ranges::any_of(
+        loaded_plugins.begin(), loaded_plugins.end(), [&](const std::string& path) {
+            return path.find(plugin_name) != std::string::npos;
+        });
+}
```
BrianHarrisonAMD: Windows and Linux have different rules for case handling in paths.
I guess this is fine, but we might want to make the windows  version agnostic to case.

BrianHarrisonAMD: Actually, can't we use std::filesystem and avoid this generally?

BrianHarrisonAMD:
```
bool are_paths_same(const fs::path& path1, const fs::path& path2) {
    try {
        // canonical() requires the file to exist
        return fs::canonical(path1) == fs::canonical(path2);
    } catch (const fs::filesystem_error& e) {
        // Handle case where file doesn't exist
        return false;
    }
}
```

BrianHarrisonAMD (approved):
Some minor comments, but LGTM.

tests/backend/test_backend_public_functions.cpp
```diff
@@ -197,23 +212,104 @@ TEST(HipDNNBackendTest, SetPluginPathsExt_FailsOnNullStringInList)
...
+    fs::path expected_plugin_path = fs::path("../../backend/src/hipdnn_plugins/engines")
+                                    / get_library_name("test_good_default_plugin");
+
+    EXPECT_TRUE(test_util::is_plugin_loaded(loaded_plugins, expected_plugin_path.string()));
+    EXPECT_EQ(hipdnnDestroy(handle), HIPDNN_STATUS_SUCCESS);
+}
+
+TEST(HipDNNBackendTest, GetLoadedPluginPaths_AdditiveLoadsBothDefaultAndCustom)
+{
+    using namespace hipdnn_tests::plugin_constants;
```
adickin-amd: this is fine but any reason you didnt just put these at the top of the test file instead of per test?

SamuelReeder: There were no global namespaces so I thought it was cleaner, but it's pointless to keep duplicating so I can change.

adickin-amd (approved):

tests/backend/test_backend_public_functions.cpp
adickin-amd: up to you, i find in test .cpps you can just pollute without any worry.  totally up to you, was just curious to know if there was a good reason to follow your pattern.

backend/src/plugin/engine_plugin_resource_manager.hpp
```diff
@@ -48,6 +48,9 @@ class Engine_plugin_resource_manager
     static void set_plugin_paths(const std::vector<std::filesystem::path>& plugin_paths,
                                  hipdnnPluginLoadingMode_ext_t loading_mode);
     static std::set<std::filesystem::path> get_plugin_paths();
+    void get_loaded_plugin_files(size_t* num_plugins,
+                                 char** plugin_paths,
+                                 size_t* max_string_len) const;
```
jerehartAMD: I think you should move this to a different section of the header file, it's in the section of "MT-safe static functions" and I think it's neither

tests/backend/test_backend_public_functions.cpp
```diff
@@ -169,22 +173,33 @@ TEST(HipDNNBackendTest, WillFailToCreateGraphIfGraphIsNull)
 
 TEST(HipDNNBackendTest, SetPluginPathsExt_Success)
 {
-    std::array<const char*, 3> paths = {"../test_plugins/test_good_plugin", "./", "../directory/"};
+    using namespace hipdnn_tests::plugin_constants;
+    std::array<const char*, 3> paths = {PLUGIN_DIR.c_str(), "./", "../directory/"};
```
jerehartAMD: I don't think this is window's portable, `std::filesystem::path::c_str()` returns a native string which will be a wchar_t pointer on windows. I don't know if there's a better way, but `.string().c_str()` will get you the right one

SamuelReeder: Ah, didn't know that. Will change.

jerehartAMD (approved):
Everything else looks good to me!

PR #114, opened 2025-08-07
Add frontend integration tests

BrianHarrisonAMD:
## Proposed changes

- Add additional E2E integration tests for the frontend API.
- Added cline rules similar to copilot.
- Added two additional plugins.
    - One fails on execute.
    - One has no applicable engines.
- Fixed the frontend graph execute not capturing cases where execute failed (found in testing).
- Flipped existing tests to specify the exact plugin to load since it was picking up the new plugins as well.

## Checklist


.gitignore
BrianHarrisonAMD, line 1: Can revert and use latest updates.

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -347,8 +347,10 @@ class Graph : public INode
         RETURN_ON_BACKEND_FAILURE(hipdnn_backend().backend_finalize(variant_pack_desc->get()),
                                   "Failed to finalize variant pack descriptor");
 
-        hipdnn_backend().backend_execute(
-            handle, _execution_plan_desc->get(), variant_pack_desc->get());
+        RETURN_ON_BACKEND_FAILURE(hipdnn_backend().backend_execute(handle,
+                                                                   _execution_plan_desc->get(),
+                                                                   variant_pack_desc->get()),
+                                  "Execute failed.");
```
BrianHarrisonAMD, line 353: Bug fix.

tests/backend/CMakeLists.txt
```diff
@@ -24,7 +24,6 @@ target_link_libraries(public_hipdnn_backend_tests
     pthread 
     hipdnn_sdk
     hipdnn_backend
-    hipdnn_frontend
```
BrianHarrisonAMD, line 27: This was incorrect before I believe.

tests/frontend/frontend_integration_test.cpp
```diff
@@ -0,0 +1,362 @@
...
+// Test case structure for parameterized tests
+struct Integration_test_case
+{
+    const char* plugin_name;
+    std::string description;
+    std::string graph_name;
+    FailurePoint expected_failure;
+    bool use_manual_uids;
+
+    // operator<< for debugging - prints test case details when test fails
```
BrianHarrisonAMD: Clean-up vibe

tests/backend/test_backend_execute_api.cpp
```diff
@@ -24,7 +24,7 @@ class Execution_backend_end_api_tests : public ::testing::Test
 
     void SetUp() override
     {
-        const std::array<const char*, 1> paths = {"../test_plugins/"};
+        const std::array<const char*, 1> paths = {"../test_plugins/libtest_good_plugin.so"};
```
BrianHarrisonAMD: Need to think about this.
It works now due to our directory layout, but I think we eventually want to have an installable test artifact(s) that can be ran.

I suppose if we install the test artifacts with this expected structure, then that would also work.

SamuelReeder: Yeah I'm not a fan either, but our build folder layout and install layout are vastly different so its kind of hacky. I'll see if I can change default install locations in plugin loading tests PR maybe.

tests/CMakeLists.txt
```diff
@@ -15,5 +15,6 @@ set(HIPDNN_TEST_INCLUDE_DIRS
 
 if(HIP_DNN_BUILD_BACKEND)
     add_subdirectory(backend)
+    add_subdirectory(frontend)
     add_subdirectory(test_plugins)
 endif()
```
jerehartAMD: No newline at the end of this file

tests/backend/test_backend_execute_api.cpp
jerehartAMD: Wouldn't hurt to have the base testing dir be shared somewhere so we could change them all if we decide to handle it differently

tests/frontend/frontend_integration_test.cpp
```diff
@@ -0,0 +1,353 @@
...
+            break;
+        case FailurePoint::CREATE_EXECUTION_PLAN:
+            os << "CREATE_EXECUTION_PLAN";
+            break;
+        case FailurePoint::EXECUTE:
+            os << "EXECUTE";
+            break;
+        default:
+            os << "UNKNOWN";
+            break;
```
jerehartAMD, line 59: How do we all feel about Wswitch-enum? I especially like it for cases like this where it's easy to expand an enum but forget to update all the switches you need to

tests/frontend/frontend_integration_test.cpp
```diff
@@ -0,0 +1,353 @@
...
+#include <hipdnn_sdk/utilities/migratable_memory.hpp>
+#include <hipdnn_sdk/utilities/tensor.hpp>
+
+using namespace hipdnn_frontend;
+using namespace hipdnn_frontend::graph;
+using namespace hipdnn_sdk::utilities;
+
+const char* const TEST_GOOD_PLUGIN = "libtest_good_plugin.so";
+const char* const TEST_EXECUTE_FAILS_PLUGIN = "libtest_execute_fails_plugin.so";
+const char* const TEST_NO_APPLICABLE_ENGINES_PLUGIN = "libtest_no_applicable_engines_plugin.so";
```
SamuelReeder: Maybe give non-fully qualified names for windows compatibility? e.g. 
`const char* const TEST_GOOD_PLUGIN = "test_good_plugin";`

tests/backend/test_engine_api.cpp
```diff
@@ -15,7 +15,7 @@ class Engine_api_tests : public ::testing::Test
 
     void SetUp() override
     {
-        const std::array<const char*, 1> paths = {"../test_plugins/"};
+        const std::array<const char*, 1> paths = {"../test_plugins/libtest_good_plugin.so"};
```
SamuelReeder: Same for these tests, since now they will definitely fail on windows, and before they would only *probably* fail on windows.

tests/test_plugins/test_execute_fails_plugin.cpp
```diff
@@ -0,0 +1,379 @@
...
+            if(test_container_ptr != nullptr)
+            {
+                (*handle)->test_container = test_container_ptr;
+            }
+            else
+            {
+                (*handle)->test_container = std::make_shared<Test_container>();
+                test_container_lifecycle_ptr = (*handle)->test_container;
+            }
+        }
```
jerehartAMD: Just to make sure I'm understanding our intended behavior:
As long as we have at least one plugin, there'll only be one test_container_ptr. If all plugins go out of scope, the test_container_ptr will be destroyed, and we create a new one when we create a new plugin?

tests/frontend/frontend_integration_test.cpp
```diff
@@ -0,0 +1,353 @@
...
+    void run_test()
+    {
+        const auto& test_case = GetParam();
+
+        // Setup environment with specified plugin
+        _handle = setup_test_environment_with_plugin(test_case.plugin_name);
+
+        // Create tensor bundle
+        std::vector<int64_t> dims = {2, 3, 14, 14}; // n=2, c=3, h=14, w=14
+        Simple_batchnorm_2d_tensor_bundle<float, float> tensor_bundle(dims);
```
SamuelReeder, line 290: Is it worth running the test with half and bf16 too?

tests/test_plugins/test_no_applicable_engines_plugin.cpp
```diff
@@ -0,0 +1,361 @@
...
+void throw_if_null(T* value)
+{
+    if(value == nullptr)
+    {
+        throw hipdnn_plugin::Hipdnn_plugin_exception(HIPDNN_PLUGIN_STATUS_BAD_PARAM,
+                                                     std::string(typeid(T).name()) + " is nullptr");
+    }
+}
+
+class Test_container
```
jerehartAMD: There's a decent amount of duplication in these two files, I wonder if we can reduce some of that?

jerehartAMD (approved):
LGTM! Mostly just small notes

tests/frontend/frontend_integration_test.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+        tensors.scale = std::make_shared<Tensor_attributes>();
+        if(use_manual_uids)
+        {
+            tensors.scale->set_uid(uid++);
+        }
+        tensors.scale->set_name("scale")
+            .set_data_type(DataType_t::FLOAT)
+            .set_dim(tensor_bundle.scale_tensor.dims())
+            .set_stride(tensor_bundle.scale_tensor.strides());
```
SamuelReeder: I think it would actually be pretty easy to put this logic in a function with name and tensor params and avoid this repetition.

tests/test_plugins/test_execute_fails_plugin.cpp
jerehartAMD: Unless we're planning on checking the value of test_container_lifecycle_ptr in our tests, maybe we could just have it be a shared_ptr and create it once? It would make this code a lot more concise

tests/frontend/frontend_integration_test.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+        // Create tensor bundle
+        std::vector<int64_t> dims = {2, 3, 14, 14}; // n=2, c=3, h=14, w=14
+        Simple_batchnorm_2d_tensor_bundle<float, float> tensor_bundle(dims);
+
+        // Create graph and tensors using the unified function
+        auto [graph, tensors] = create_batchnorm_test_graph_with_uids(
+            test_case.graph_name, tensor_bundle, test_case.use_manual_uids);
+
+        run_graph_pipeline(graph, _handle, tensors, tensor_bundle, test_case.expected_failure);
```
SamuelReeder, line 296: Could also consider adding cpu verification, maybe not intended for these tests though.

tests/test_plugins/test_no_applicable_engines_plugin.cpp
SamuelReeder: Yeah, I wonder if its possible to create a common header with all the generic functionality and override methods where applicable in these. And virtual methods for things that each plugin needs to implement, like name. If the C-api messes it up, could just have the C-api always call those class methods and then put all the logic in the class.

tests/frontend/frontend_integration_test.cpp
BrianHarrisonAMD: I don't think it really matters since it's not real.

tests/frontend/frontend_integration_test.cpp
BrianHarrisonAMD: HmmmMmmm

BrianHarrisonAMD: Ill do a pass to make these more likely to work, but it's known we need to do some Windows work, and I don't like adding untested stuff.

tests/frontend/frontend_integration_test.cpp
SamuelReeder: Well since its a fake plugin definitely not intended for these tests, so saw many batchnorm names that I forgot it wasn't actually doing batchnorm lol

tests/test_plugins/test_execute_fails_plugin.cpp
BrianHarrisonAMD: Yea I think this was ripped from the existing one, which also ripped from the real one.
In this case, it probably can be a shared_ptr.

tests/frontend/frontend_integration_test.cpp
BrianHarrisonAMD: We tried to turn it on, and since we have incomplete impl for all the backend enums, it results in a bunch of files needing to disable that check manually.

tests/backend/test_backend_execute_api.cpp
BrianHarrisonAMD: Ill make some utilities that can be shared for this, and consolidate it all in one spot.

tests/test_plugins/test_no_applicable_engines_plugin.cpp
BrianHarrisonAMD: Yea we could absolutely make some generic plugin testing stub to reduce these.

BrianHarrisonAMD: Although, a lot of it ends up needing to exist due to C api.
Ill see if there is anything useful to keep.
I think I can just reduce these plugin files generally.

BrianHarrisonAMD: Going to see if I can make a class that I can extend with specific implementation and share the generic structure.

tests/test_plugins/test_execute_fails_plugin.cpp
BrianHarrisonAMD: Actually it's not really needed, and I just purged it generally.

SamuelReeder (approved):
LGTM, the refactoring is nice!

tests/test_plugins/test_plugin_common.hpp
```diff
@@ -0,0 +1,510 @@
...
+
+    // Virtual methods to be overridden by derived classes
+    virtual const char* get_plugin_name() const = 0;
+    virtual const char* get_plugin_version() const = 0;
+    virtual int64_t get_engine_id() const = 0;
+    virtual uint32_t get_num_applicable_engines() const = 0;
+    virtual bool supports_engine_operations() const
+    {
+        return get_num_applicable_engines() > 0;
+    }
```
jerehartAMD, line 44: I don't think I've seen this pattern before haha. This class is a mostly virtual base class, and also a class that logs function calls to a registered derivation of that base class. It feels like these could be two separate classes to make each a little more specialized (one as the base class, and one for the static singleton interface). That way you could just have a shared or unique pointer to the base class if you don't want the logging, or register it with the test logger if you want everything. There might be extra benefits to having it as one though that I'm not seeing. At the end of the day, I think this should do the job though?

jerehartAMD (commented):
Love all of that deleted code. I think if you decided you want to split the class I mentioned into two it'd be relatively easy, but I'm cool with it either way!

tests/test_plugins/test_plugin_common.hpp
BrianHarrisonAMD: Yea since it's just a testing stub, I am not too worried about it.

We could have a separate logging interface, but ultimately I think this should get us what we need and keep it all in one spot.

tests/frontend/frontend_integration_test.cpp
BrianHarrisonAMD: Yea absolutely.
I thought we added this, but I guess not.
Ill add it.

sdk/include/hipdnn_sdk/utilities/platform_path_utils.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <string>
+
+namespace hipdnn_sdk
+{
+namespace utilities
+{
+
+#ifdef _WIN32
+
+typedef HMODULE Plugin_lib_handle;
```
mousdahl-amd: I'm not 100% sure, but Plugin_lib_handle could be moved back to the backend.

mousdahl-amd (approved):
I think Plugin_lib_handle could go back to the backend platfor_utils, but other than that I think it looks good.

tests/test_plugins/test_plugin_constants.hpp
```diff
@@ -0,0 +1,57 @@
...
+    inline const char* test_good_plugin_name = TEST_GOOD_PLUGIN_NAME;
+    inline const char* test_execute_fails_plugin_name = TEST_EXECUTE_FAILS_PLUGIN_NAME;
+    inline const char* test_no_applicable_engines_plugin_name = TEST_NO_APPLICABLE_ENGINES_PLUGIN_NAME;
+
+    // Compose full plugin path with existence checking
+    inline std::string get_plugin_path(const char* plugin_name)
+    {
+        namespace fs = std::filesystem;
+        
+        fs::path plugin_file = PLUGIN_DIR / hipdnn_sdk::utilities::get_library_name(plugin_name);
```
SamuelReeder, line 27: Might need a way to use a different `PLUGIN_DIR` for tests outside the `build/tests/...` folder, but this looks good for now.

sdk/include/hipdnn_sdk/utilities/platform_path_utils.hpp
SamuelReeder: Yeah I don't think it would be used elsewhere, maybe its just nicely organized here though.

SamuelReeder (approved):
Still looks good! New additions are useful too.

sdk/include/hipdnn_sdk/utilities/platform_path_utils.hpp
BrianHarrisonAMD: Yea good point let me check.

PR #115, opened 2025-08-08
Add conv fwd

BrianHarrisonAMD:
## Proposed changes

Add conv fwd to the SDK & Frontend of hipDNN

## Checklist


frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -562,6 +564,35 @@ class Graph : public INode
         return out_0;
     }
 
+    std::shared_ptr<Tensor_attributes> conv_fprop(const std::shared_ptr<Tensor_attributes>& x,
+                                                  const std::shared_ptr<Tensor_attributes>& w,
+                                                  Conv_fprop_attributes attributes)
+    {
+        if(attributes.name.empty())
```
mousdahl-amd, line 571: Do name collisions matter?

frontend/tests/test_convolution_fprop_node.cpp
```diff
@@ -0,0 +1,329 @@
...
+    conv_attributes.set_stride({1, 1});
+    conv_attributes.set_dilation({1, 1});
+
+    Graph_attributes graph_attributes;
+    ConvolutionNode node(std::move(conv_attributes), graph_attributes);
+
+    std::unordered_set<int64_t> used_ids;
+    node.gather_hipdnn_tensor_ids(used_ids);
+
+    EXPECT_TRUE(used_ids.find(1) != used_ids.end());
```
mousdahl-amd: I'm surprised clang-tidy didn't nail you for not using .contains()

mousdahl-amd (approved):
Couple vague questions, looks good.

frontend/tests/test_convolution_fprop_node.cpp
BrianHarrisonAMD: Ruh-oh, now I am worried tidy isn't running on these tests hehe.

frontend/include/hipdnn_frontend/graph.hpp
BrianHarrisonAMD: No they shouldn't.
Names are just for logging and tracking purposes really.

frontend/tests/test_convolution_fprop_node.cpp
BrianHarrisonAMD: Hmm I did a rename, and I think the test file got duplicated.

BrianHarrisonAMD: Okay what happened was the test file got split awkwardly when I had a DC.
There was two files, and this one wasn't actually included.
The other one was the wrong name, but had the updated tests that were correct.

I removed this one, and renamed the other to be the proper name.

frontend/tests/test_conv_fprop_node.cpp
```diff
@@ -0,0 +1,667 @@
...
+    conv_attributes.set_post_padding({0, 1, 1});
+    conv_attributes.set_stride({1, 1, 1});
+    conv_attributes.set_dilation({1, 1, 1});
+
+    Graph_attributes graph_attributes;
+    ConvolutionNode node(std::move(conv_attributes), graph_attributes);
+
+    auto error = node.infer_properties_node();
+    EXPECT_EQ(error.code, error_code_t::OK) << error.err_msg;
+    ;
```
BrianHarrisonAMD: huh

frontend/include/hipdnn_frontend/graph.hpp
BrianHarrisonAMD: This might change in the future when we add serialization, but I would hope not.

frontend/include/hipdnn_frontend/attributes/convolution_fwd_attributes.hpp
```diff
@@ -0,0 +1,185 @@
...
+
+    std::unordered_map<input_names, std::shared_ptr<Tensor_attributes>> inputs;
+    std::unordered_map<output_names, std::shared_ptr<Tensor_attributes>> outputs;
+
+    // Convolution parameters
+    std::vector<int64_t> pre_padding;
+    std::vector<int64_t> post_padding;
+    std::vector<int64_t> stride;
+    std::vector<int64_t> dilation;
+    ConvolutionMode_t conv_mode = ConvolutionMode_t::CROSS_CORRELATION;
```
SamuelReeder, line 40: Should there be a group_count parameter?

SamuelReeder: Realized it might be implicit by channels of input tensor vs filter tensor.

SamuelReeder: I'm thinking a test for creating a grouped infer node in `frontend/tests/test_conv_fprop_node.cpp` would be good.

frontend/tests/test_conv_fprop_node.cpp
SamuelReeder, line 1: Not critical, but test cases for asymmetric padding and dilation and/or strides > 1 could be useful. Maybe a case for zero padding too.

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
```diff
@@ -0,0 +1,229 @@
...
+        if(attributes.get_dilation().empty())
+        {
+            return {error_code_t::ATTRIBUTE_NOT_SET,
+                    "ConvolutionNode missing dilation for pre-validation"};
+        }
+
+        return {};
+    }
+
+    error_t infer_properties_node() override
```
SamuelReeder, line 211: Assuming this does support grouping implicitly, I think it should validate that input channels dim is divisible by filter channels dim.

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
```diff
@@ -0,0 +1,229 @@
...
+public:
+    Conv_fprop_attributes attributes;
+
+    ConvolutionNode(Conv_fprop_attributes&& conv_attrs, const Graph_attributes& graph_attrs)
+        : NodeCRTP(graph_attrs)
+        , attributes(std::move(conv_attrs))
+    {
+    }
+
+    error_t pre_validate_node() const override
```
SamuelReeder, line 26: Some validation is probably the responsibility of the plugin, but things that don't make sense like tensors with mismatched number of dimensions might be possible to check here and throw a specific error.

frontend/tests/test_conv_fprop_node.cpp
```diff
@@ -0,0 +1,713 @@
...
+    conv_attributes.set_post_padding({1, 1});
+    conv_attributes.set_stride({1, 1});
+    conv_attributes.set_dilation({1, 1});
+
+    Graph_attributes graph_attributes;
+    ConvolutionNode node(std::move(conv_attributes), graph_attributes);
+
+    auto error = node.pre_validate_node();
+    EXPECT_EQ(error.code, error_code_t::OK) << error.err_msg;
+    ;
```
BrianHarrisonAMD: Extra ;

frontend/include/hipdnn_frontend/graph.hpp
```diff
@@ -562,6 +564,35 @@ class Graph : public INode
         return out_0;
     }
 
+    std::shared_ptr<Tensor_attributes> conv_fprop(const std::shared_ptr<Tensor_attributes>& x,
+                                                  const std::shared_ptr<Tensor_attributes>& w,
+                                                  Conv_fprop_attributes attributes)
```
jerehartAMD: I think there's some (albiet small) potential benefits to passing by value because you can use moves to avoid a copy and a reference count increase.

frontend/include/hipdnn_frontend/attributes/convolution_fwd_attributes.hpp
```diff
@@ -0,0 +1,185 @@
...
+        return get_input(input_names::W);
+    }
+
+    std::shared_ptr<Tensor_attributes> get_y() const
+    {
+        return get_output(output_names::Y);
+    }
+
+    // Setters for tensors
+    Conv_fprop_attributes& set_x(const std::shared_ptr<Tensor_attributes>& value)
```
jerehartAMD, line 64: This is a pattern that we use in many other places, so this definitely isn't gating for this PR, but I think we should take these shared_ptrs by value instead. I've seen discussions on this and agree more strongly with the pass by value arguments for cases like this that take ownership (or using a reference for the underlying class if you don't plan to take/share ownership).

Const ref is forcing at least one copy, and is a little semantically weird for me to see on a shared_ptr here. By value can allow you to avoid all copies and just move

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
```diff
@@ -0,0 +1,229 @@
...
+        if(attributes.get_stride().empty())
+        {
+            return {error_code_t::ATTRIBUTE_NOT_SET,
+                    "ConvolutionNode missing stride for pre-validation"};
+        }
+        if(attributes.get_dilation().empty())
+        {
+            return {error_code_t::ATTRIBUTE_NOT_SET,
+                    "ConvolutionNode missing dilation for pre-validation"};
+        }
```
jerehartAMD, line 206: Not sure how hot of a spot this is for our code, but this is a lot of branches, and we could possibly have an early exit that ands all of them together. If this performance doesn't matter particularly I expect this is 100% fine and easier to maintain.

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
```diff
@@ -0,0 +1,229 @@
...
+            for(size_t i = 2; i < x_dims.size(); ++i)
+            {
+                auto spatial_idx = i - 2; // Index into spatial dimension arrays
+
+                if(spatial_idx >= pre_padding.size() || spatial_idx >= post_padding.size()
+                   || spatial_idx >= stride.size() || spatial_idx >= dilation.size())
+                {
+                    return {error_code_t::ATTRIBUTE_NOT_SET,
+                            "ConvolutionNode: Insufficient padding/stride/dilation parameters for "
+                            "spatial dimensions"};
```
jerehartAMD: Should this be INVALID_VALUE instead?

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
```diff
@@ -0,0 +1,229 @@
...
+                // Validate parameters
+                if(stride_val <= 0)
+                {
+                    return {error_code_t::ATTRIBUTE_NOT_SET,
+                            "ConvolutionNode: Stride must be positive"};
+                }
+                if(dilation_val <= 0)
+                {
+                    return {error_code_t::ATTRIBUTE_NOT_SET,
+                            "ConvolutionNode: Dilation must be positive"};
```
jerehartAMD: Same for these?

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
```diff
@@ -0,0 +1,229 @@
...
+                // Calculate dilated kernel size
+                auto dilated_kernel_size = (dilation_val * (kernel_size - 1)) + 1;
+
+                // Calculate output dimension
+                auto numerator = input_size + pre_pad + post_pad - dilated_kernel_size;
+                if(numerator < 0)
+                {
+                    return {error_code_t::ATTRIBUTE_NOT_SET,
+                            "ConvolutionNode: Invalid convolution parameters result in negative "
+                            "output size"};
```
jerehartAMD: And here?

(There's a couple more, but I'll leave it at this)

frontend/include/hipdnn_frontend/attributes/convolution_fwd_attributes.hpp
BrianHarrisonAMD: Grouped is handled by how dims are specified between filter / input.
I could add an explicit grouped test though!

frontend/tests/test_conv_fprop_node.cpp
BrianHarrisonAMD: Yea I am adding those and agree.

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
BrianHarrisonAMD: Yea lets add that.

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
BrianHarrisonAMD: I also agree with that, and was debating adding it.

jerehartAMD (approved):
I've got some comments but nothing blocking, looks good to me!

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
BrianHarrisonAMD: Should be once per node (which shouldn't be a very hot path).
I think the more descriptive errors is better than any potential minor gains here.

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
```diff
@@ -0,0 +1,382 @@
...
+        {
+            return {error_code_t::ATTRIBUTE_NOT_SET,
+                    "ConvolutionNode missing dilation for pre-validation"};
+        }
+
+        // Get tensor references
+        auto x = attributes.get_x();
+        auto w = attributes.get_w();
+        auto y = attributes.get_y();
+
```
BrianHarrisonAMD, line 62: Blah

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
BrianHarrisonAMD: Yep :(

frontend/tests/test_conv_fprop_node.cpp
```diff
@@ -0,0 +1,2040 @@
...
+    ConvolutionNode node(std::move(conv_attributes), graph_attributes);
+
+    auto error = node.infer_properties_node();
+    EXPECT_EQ(error.code, error_code_t::OK) << error.err_msg;
+
+    // Check inferred dimensions
+    auto inferred_dims = y_tensor->get_dim();
+    EXPECT_EQ(inferred_dims.size(), 4);
+    EXPECT_EQ(inferred_dims[0], 1); // Batch size
+    EXPECT_EQ(inferred_dims[1], 256); // Output channels: 128 * (64/32) = 256
```
SamuelReeder: This changes to 128 with grouped changes. A few other relevant tests below.

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
```diff
@@ -0,0 +1,411 @@
...
+        if(w_strides.size() != w_dims.size())
+        {
+            return {error_code_t::INVALID_VALUE,
+                    "ConvolutionNode: Weight tensor stride count must match dimension count"};
+        }
+
+        // Validate input channels match between input and weight tensors
+        // For regular convolution: x_dims[1] == w_dims[1]
+        // For grouped convolution: x_dims[1] % w_dims[1] == 0
+        if(x_dims[1] % w_dims[1] != 0)
```
mousdahl-amd: Check w_dims[1] is non-zero first?

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
```diff
@@ -0,0 +1,411 @@
...
+            {
+                return {
+                    error_code_t::INVALID_VALUE,
+                    "ConvolutionNode: Output tensor batch size must match input tensor batch size"};
+            }
+
+            // Validate output channels match weight output channels
+            // For regular convolution: y_dims[1] == w_dims[0]
+            // For grouped convolution: y_dims[1] % w_dims[0] == 0
+            if(y_dims[1] % w_dims[0] != 0)
```
mousdahl-amd: Same here for w_dims[0]

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
```diff
@@ -0,0 +1,411 @@
...
+    Conv_fprop_attributes attributes;
+
+    ConvolutionNode(Conv_fprop_attributes&& conv_attrs, const Graph_attributes& graph_attrs)
+        : NodeCRTP(graph_attrs)
+        , attributes(std::move(conv_attrs))
+    {
+    }
+
+    error_t pre_validate_node() const override
+    {
```
mousdahl-amd, line 27: A couple of suggestions:

1. Perhaps add a private method to Node like `has_attribute(bool, msg)` that throws, and catch it here and return.  It might make this method a bit more readable.  A similar method like `valid_attribute(bool, msg)`.  It could pre-pend the `ConvolutionNode: ` to the text.
2. Instead of throwing or returning, would it make sense perhaps to collect all the errors (maybe in two phases, since we're returning two error_code_t).  Then if the node has multiple problems, they can fix them all in one go?

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
```diff
@@ -0,0 +1,411 @@
...
+                if(pre_pad < 0)
+                {
+                    return {error_code_t::INVALID_VALUE,
+                            "ConvolutionNode: Pre-padding must be non-negative"};
+                }
+                if(post_pad < 0)
+                {
+                    return {error_code_t::INVALID_VALUE,
+                            "ConvolutionNode: Post-padding must be non-negative"};
+                }
```
SamuelReeder: `pre_validate_node` checks this too. Whether its useful depends on if `pre_validate_node` will always be called beforehand.

SamuelReeder: There are also some negative values tests for `infer_properties_node` that can be removed if these checks are unnecessary. But up to you if this should be changed; I'm not sure what's best.

SamuelReeder (approved):
LGTM!

Just change the `y_dims[1]` to total output channels alongside relevant checks & tests and can consider my duplication comment.

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
mousdahl-amd: Forget 2.  That's bad, wrong, and complicated.

BrianHarrisonAMD: Okay discussed:

1. We will add a MACRO to make these checks easier to read similar to how we handle the THROW_IF
2. We are going to leave as is, since we are mixing error_code_t types, and it becomes a bit complex since some checks are only valid if you are in a valid state.

frontend/include/hipdnn_frontend/node/convolution_fprop_node.hpp
SamuelReeder: I assume it's not possible to just use unsigned ints? Edit: yeah, probably want to avoid mixing integer types and underflow lol

BrianHarrisonAMD: I think I rather duplicate these simple checks, and ensure it's good.
I also had a similar thought, but since they are public APIs, I would rather not have it get called out of order and explode in a bad way.

SamuelReeder (approved):
LGTM

frontend/include/hipdnn_frontend/error.hpp
```diff
@@ -77,4 +77,41 @@ typedef struct error_object // NOLINT
         return code != other.code;
     }
 } error_t;
+
+
+#define RETURN_IF_NE(x, y, error_status, message) \
```
jerehartAMD: These are in a header file, to avoid conflicts should we add a HIPDNN to the beginning?

BrianHarrisonAMD: Yea probably a good idea.

PR #116, opened 2025-08-08
[ALMIOPEN-144] NHWC tensor support & tests & samples

SamuelReeder:
## Proposed changes

Adds NHWC tensor layout and updates samples and tests to use this layout in some cases.

## Checklist

### DIFF
--- a/plugins/miopen_legacy_plugin/integration_tests/batchnorm_fwd_inference_integration_test.cpp
+++ b/plugins/miopen_legacy_plugin/integration_tests/batchnorm_fwd_inference_integration_test.cpp
@@ -45,16 +45,14 @@ struct Batchnorm_2d_tensor_bundle
 {
     Batchnorm_2d_tensor_bundle(const std::vector<int64_t>& dims,
                                unsigned int seed = 1,
-                               bool use_nhwc = false)
+                               TensorLayout layout = TensorLayout::NCHW)
         : derived_dims({1, dims[1], 1, 1})
-        , x_tensor(use_nhwc ? Tensor::make_nhwc_tensor<Input_type>(dims)
-                            : Tensor::make_nchw_tensor<Input_type>(dims))
-        , y_tensor(use_nhwc ? Tensor::make_nhwc_tensor<Input_type>(dims)
-                            : Tensor::make_nchw_tensor<Input_type>(dims))
-        , scale_tensor(Tensor::make_nchw_tensor<Intermediate_type>(derived_dims))
-        , bias_tensor(Tensor::make_nchw_tensor<Intermediate_type>(derived_dims))
-        , mean_tensor(Tensor::make_nchw_tensor<Intermediate_type>(derived_dims))
-        , variance_tensor(Tensor::make_nchw_tensor<Intermediate_type>(derived_dims))
+        , x_tensor(Tensor::make_tensor<Input_type>(dims, layout))
+        , y_tensor(Tensor::make_tensor<Input_type>(dims, layout))
+        , scale_tensor(Tensor::make_tensor<Intermediate_type>(derived_dims, layout))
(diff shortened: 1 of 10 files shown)


sdk/include/hipdnn_sdk/utilities/tensor.hpp
```diff
@@ -35,12 +60,27 @@ class Tensor
...
+        switch(layout)
+        {
+        case TensorLayout::NCHW:
+            return make_tensor<T>(dims);
+        case TensorLayout::NHWC:
+            return {dims, generate_strides(dims, stride_order_nhwc(dims.size())), sizeof(T)};
+        default:
+            throw std::invalid_argument("Unsupported tensor layout");
+        }
+    }
```
BrianHarrisonAMD: Can just make a default value of NCHW, or call the other one instead of duplicating.

plugins/miopen_legacy_plugin/integration_tests/batchnorm_fwd_inference_integration_test.cpp
```diff
@@ -43,14 +43,16 @@ struct Bn_2d_test_case
...
-        , scale_tensor(Tensor::make_nchw_tensor<Intermediate_type>(derived_dims))
-        , bias_tensor(Tensor::make_nchw_tensor<Intermediate_type>(derived_dims))
-        , mean_tensor(Tensor::make_nchw_tensor<Intermediate_type>(derived_dims))
-        , variance_tensor(Tensor::make_nchw_tensor<Intermediate_type>(derived_dims))
+        , x_tensor(Tensor::make_tensor<Input_type>(dims, layout))
+        , y_tensor(Tensor::make_tensor<Input_type>(dims, layout))
+        , scale_tensor(Tensor::make_tensor<Intermediate_type>(derived_dims))
+        , bias_tensor(Tensor::make_tensor<Intermediate_type>(derived_dims))
+        , mean_tensor(Tensor::make_tensor<Intermediate_type>(derived_dims))
+        , variance_tensor(Tensor::make_tensor<Intermediate_type>(derived_dims))
```
BrianHarrisonAMD, line 55: It might be better to make the derived_dim tensors match the layout to what the other tensors are.
It shouldn't matter since they are 1D tensors and technically work with any layout, but I have seen some checks that fail when mixing in other places.

It's probably fine either way though, and if it's working for the MIOpen plugin, then you can leave it.

BrianHarrisonAMD (approved):
Looking good!

Nice to see that NHWC does in fact work automatically without extra integration work (:

sdk/include/hipdnn_sdk/utilities/tensor.hpp
```diff
@@ -14,6 +15,30 @@ namespace hipdnn_sdk
 namespace utilities
 {
 
+enum class TensorLayout
```
mousdahl-amd: I'm wondering if we could make Tensor_layout a class / struct:
```
class Tensor_layout {
public:
    string name;
    vector<int> stride_order;

    static Tensor_layout NCHW { .name="NCHW", .stride_order={3,2,1,0} };
    static Tensor_layout NHWC{ .name="NCHW", .stride_order=<whatever> };
}
```

and then passing in a layout like `make_tensor(dims, Tensor_layout.NCHW)`

mousdahl-amd (approved):
Looks good, minor suggestion

sdk/include/hipdnn_sdk/utilities/tensor.hpp
SamuelReeder: Yeah good idea, will need no function overloads then.

mousdahl-amd (approved):

sdk/include/hipdnn_sdk/utilities/tensor.hpp
SamuelReeder: Had to put the layout definitions outside due to compiler, but think its fine.

PR #117, opened 2025-08-08
Handle gmock uninteresting call warnings

jerehartAMD:
## Proposed changes

Gmock warns for uninteresting calls when a mocked function is called when expectations aren't noted for it. For this PR I looked at all of the calls outside of the test_graph.cpp file. There were a small number, and fit into two categories:
* Destructors being called - For which I think we want to either state whether we expect them, or when we don't expect them
* Logging functions - which I think are generally uninteresting

The vast majority of these warnings are in the graph_test.cpp file. Those have been turned off selectively by setting`FLAGS_gmock_verbose="error"`

## Checklist

### DIFF
--- a/backend/tests/descriptors/test_engine_config_descriptor.cpp
+++ b/backend/tests/descriptors/test_engine_config_descriptor.cpp
@@ -69,7 +69,7 @@ public:
 
     void make_engine_config_finalized() const
     {
-        set_engine();
+        // TODO: With the call to finalize and get_mock_engine this makes it likely that we're interweaving EXPECTs and function calls
         EXPECT_CALL(*get_mock_engine(), is_finalized()).WillRepeatedly(Return(true));
         EXPECT_CALL(*get_mock_engine(), get_engine_id()).WillRepeatedly(Return(1));
         EXPECT_CALL(*get_mock_engine(), get_graph())
@@ -80,6 +80,8 @@ public:
             .WillOnce(Return(_mock_engine_plugin_resource_manager));
         EXPECT_CALL(*_mock_engine_plugin_resource_manager, get_workspace_size(_, _, _))
             .WillOnce(Return(1024));
+
+        set_engine();
         ASSERT_NO_THROW(get_engine_config_descriptor()->finalize());
     }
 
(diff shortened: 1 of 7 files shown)


backend/tests/descriptors/test_engine_config_descriptor.cpp
```diff
@@ -69,7 +69,10 @@ class Engine_config_descriptor_test : public ::testing::Test
 
     void make_engine_config_finalized() const
     {
-        set_engine();
+        // TODO: These expectations being hidden in here are dangerous.
+        // It's easy to forget they exist, and if any of these functions are called prior to this
+        // call it is undefined behavior. Similarly, any expectations on functions called within
+        // "finalize" or "setup" made after calling this function are UB
```
jerehartAMD, line 75: This function's a bit of a problem, reasoning stated above. It's currently not being used in a problematic way, but it easily can be

BrianHarrisonAMD: It's a bit annoying since we want to be able to include expect calls with the re-usable state that needs to be setup.
We could force defining it in each test, but I think that leads to more problems as well. 

Not sure if there is a good way to resolve this.

BrianHarrisonAMD:
Looks good to me.
The ones with `// uninteresting` could probably be left, but I also dislike the warning output.
I don't want to bulk turn these off either, as they did flag several that looked important to add. 

I guess this is the best we can do.

BrianHarrisonAMD (approved):

frontend/tests/test_graph.cpp
```diff
@@ -264,6 +264,7 @@ static void validate_tensor(const Tensor_attributes& tensor,
 // NOLINTBEGIN
 TEST_F(Graph_test_fixture, BuildAndSerializeBatchnormInferenceGraph)
 {
+    ::testing::FLAGS_gmock_verbose = "error";
```
adickin-amd, line 267: when i refactor this Ill take some time to see if I can figure out a nice way to setup mocks without creating too much overhead.

adickin-amd (approved):
Changes look great, ty for fixing all of these up.

backend/tests/descriptors/test_engine_config_descriptor.cpp
jerehartAMD: I tried splitting this into two different functions, one with the expectations and one with the setup, part of the problem is that the expectations also set the default return values that are required for the setup to work right. If we do that we need to make sure that both functions are always called together (or at least the expectation setting one is called before the setup one)

PR #118, opened 2025-08-11
More tests for engine resource manager

bghimireamd:
## Proposed changes

Add more test suites for backend/src/plugin/engine_plugin_resource_manager.hpp

develop vs bg/test_engine_plugin_resource_manager

- __More tests__: **44** tests vs **33** in develop (11 additional tests)

- __Better coverage__: **506** mismatched data warnings vs **700** in develop

## Checklist


backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -195,9 +200,26 @@ void Engine_plugin_resource_manager::get_engine_details(
...
+    {
+        throw Hipdnn_exception(HIPDNN_STATUS_BAD_PARAM, "Engine details cannot be null");
+    }
+
+    auto it = _engine_id_to_handle.find(engine_id);
+    if(it == _engine_id_to_handle.end())
+    {
+        throw Hipdnn_exception(HIPDNN_STATUS_BAD_PARAM,
+                               "Invalid engine ID: " + std::to_string(engine_id));
+    }
```
BrianHarrisonAMD, line 256: Probably want to make any of these failing checks use the internal error status instead.
The bad param status is meant for when the user provides bad inputs, and any fails in this class would likely be a result of an internal error allowing it to get this far, or a PLUGIN_ERROR if it's related to a plugin issue.

I think there was already a few that were BAD_PARAM, and they should be swapped as well.

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -165,6 +165,11 @@ void Engine_plugin_resource_manager::set_stream(hipStream_t stream) const
 std::vector<int64_t> Engine_plugin_resource_manager::get_applicable_engine_ids(
     const Graph_descriptor* graph_desc) const
 {
+    if(graph_desc == nullptr)
+    {
+        throw Hipdnn_exception(HIPDNN_STATUS_BAD_PARAM, "Graph descriptor cannot be null");
+    }
```
BrianHarrisonAMD: I think these checks can be swapped to the THROW_IF_NULL macro.
Some of the niche checks probably can't really be, but several of these can use the macro's I think.

backend/tests/CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ target_link_libraries(hipdnn_backend_tests
     GTest::gmock
     pthread
     hipdnn_backend_private
+    hipdnn_backend
```
BrianHarrisonAMD: I don't think this should be needed.

Are we using the public backend API directly?

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+    std::vector<std::filesystem::path> initial_paths = {"/test/path1", "/test/path2"};
+    Engine_plugin_resource_manager::set_plugin_paths(initial_paths, HIPDNN_PLUGIN_LOADING_ABSOLUTE);
+
+    std::vector<std::filesystem::path> new_paths = {"/test/path3", "/test/path4"};
+    Engine_plugin_resource_manager::set_plugin_paths(new_paths, HIPDNN_PLUGIN_LOADING_ABSOLUTE);
+
+    auto retrieved_paths = Engine_plugin_resource_manager::get_plugin_paths();
+
+    std::set<std::filesystem::path> expected_paths = {"/test/path3", "/test/path4"};
+    EXPECT_EQ(retrieved_paths, expected_paths);
```
BrianHarrisonAMD: Can probably reuse new_paths here, but not that big of a deal to copy paste the path's in a test.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+{
+    auto rm1 = create_resource_manager();
+    auto rm2 = create_resource_manager();
+    auto original_ptr = rm1.get();
+
+    rm2 = std::move(rm1);
+
+    EXPECT_EQ(rm2.get(), original_ptr);
+    EXPECT_EQ(rm1.get(), nullptr);
+}
```
BrianHarrisonAMD: These are testing the move behavior of std::shared_ptr I believe, and I don't think is needed.

We could change these to check the behavior of moving the resource manager itself if we wanted.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+
+TEST_F(Engine_plugin_resource_manager_good_conditions_test,
+       Construction_ScopedCreationAndDestruction)
+{
+    {
+        auto rm = create_resource_manager();
+        ASSERT_NE(rm, nullptr);
+    }
+}
```
BrianHarrisonAMD: Is this just checking that it doesn't throw during ~? 
Not really sure what this checks, but maybe it's checking the destructor is valid?

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+    Scoped_resource stream_res(stream, [](hipStream_t s) { std::ignore = hipStreamDestroy(s); });
+
+    EXPECT_NO_THROW(rm->set_stream(stream));
+}
+
+TEST_F(Engine_plugin_resource_manager_good_conditions_test, StreamManagement_SetNullStream)
+{
+    auto rm = create_resource_manager();
+
+    EXPECT_NO_THROW(rm->set_stream(nullptr));
```
BrianHarrisonAMD: This is interesting.
I think we probably should be throwing if a null stream is set. 
Might be worth discussing.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+    EXPECT_NO_THROW(rm->set_stream(stream2));
+    EXPECT_NO_THROW(rm->set_stream(nullptr));
+}
+
+TEST_F(Engine_plugin_resource_manager_good_conditions_test,
+       EngineDiscovery_GetApplicableEngineIdsWithNullptr)
+{
+    auto rm = create_resource_manager();
+
+    EXPECT_THROW(rm->get_applicable_engine_ids(nullptr), Hipdnn_exception);
```
BrianHarrisonAMD: I think we have a macro for this that lets you specify the status code as well we should be using.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+        EXPECT_EQ(engine_ids[2], 102);
+    }
+    catch(const Hipdnn_exception& e)
+    {
+        SUCCEED() << "Plugin loading or engine discovery failed as expected: " << e.what();
+    }
+    catch(const std::exception& e)
+    {
+        SUCCEED() << "Plugin loading or engine discovery failed: " << e.what();
+    }
```
BrianHarrisonAMD: Fail instead of success here? 

Also, I think the macro should be used if we are expecting exceptions.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+
+        EXPECT_EQ(engine_ids.size(), 3);
+        EXPECT_EQ(engine_ids[0], 100);
+        EXPECT_EQ(engine_ids[1], 101);
+        EXPECT_EQ(engine_ids[2], 102);
+    }
+    catch(const Hipdnn_exception& e)
+    {
+        SUCCEED() << "Plugin loading or engine discovery failed as expected: " << e.what();
+    }
```
BrianHarrisonAMD: Should this be throwing?

Seems like we are testing that it works, and the try/catch should be removed entirely.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+    catch(const Hipdnn_exception& e)
+    {
+        SUCCEED() << "Integration workflow failed as expected (plugin may not be available): "
+                  << e.what();
+    }
+    catch(const std::exception& e)
+    {
+        SUCCEED() << "Integration workflow failed: " << e.what();
+    }
```
BrianHarrisonAMD: This one for sure looks like we aren't expecting a throw, and there should be no try/catch.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+
+        for(auto engine_id : engine_ids)
+        {
+            auto details = Engine_plugin_resource_manager::get_engine_details(
+                rm, engine_id, graph_desc.get());
+            ASSERT_NE(details, nullptr);
+            EXPECT_EQ(details->get()->engine_id(), engine_id);
+
+            auto ws_size = rm->get_workspace_size(engine_id, &engine_config, graph_desc.get());
+            EXPECT_GT(ws_size, 0);
```
BrianHarrisonAMD: Since it's using a stub, we should check for the exact value we are expecting.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+       StaticPluginPathManagement_SetPathsWithActiveHandle)
+{
+    auto rm = Engine_plugin_resource_manager::create();
+
+    std::vector<std::filesystem::path> plugin_paths = {"/test/path"};
+
+    EXPECT_THROW(Engine_plugin_resource_manager::set_plugin_paths(plugin_paths,
+                                                                  HIPDNN_PLUGIN_LOADING_ABSOLUTE),
+                 Hipdnn_exception);
+}
```
BrianHarrisonAMD: Might want to rename this to SetPathWithActiveResourceManager

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+{
+    auto rm1 = create_resource_manager();
+    auto rm2 = create_resource_manager();
+
+    EXPECT_NO_THROW(rm1->set_stream(nullptr));
+    EXPECT_NO_THROW(rm2->set_stream(nullptr));
+
+    EXPECT_THROW(rm1->get_applicable_engine_ids(nullptr), Hipdnn_exception);
+    EXPECT_THROW(rm2->get_applicable_engine_ids(nullptr), Hipdnn_exception);
+}
```
BrianHarrisonAMD: These are basically the same.
I think the tests is verifying that each time you create one, you get a new shared_ptr, and if that's the case we probably only need one version of this without the extra checks afterwards.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+            ASSERT_NE(rm, nullptr);
+
+            auto rm_moved = std::move(rm);
+            ASSERT_NE(rm_moved, nullptr);
+            EXPECT_EQ(rm, nullptr);
+
+            EXPECT_NO_THROW(rm_moved->set_stream(nullptr));
+        }
+    }
+}
```
BrianHarrisonAMD: This is duplicated several times, and I think it's basically testing that the ~ is working. 
We probably only need one test that verifies ~ works without error.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+    }
+
+    for(size_t i = 1; i < managers.size(); ++i)
+    {
+        if(managers[i] != nullptr)
+        {
+            EXPECT_THROW(managers[i]->get_applicable_engine_ids(nullptr), Hipdnn_exception);
+        }
+    }
+}
```
BrianHarrisonAMD: This is testing the std::move of the std::shared_ptr.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+            EXPECT_NE(rm, nullptr);
+
+            EXPECT_NO_THROW(rm->set_stream(nullptr));
+        }
+        catch(const std::exception& e)
+        {
+            SUCCEED() << "Construction failed with exception: " << e.what();
+        }
+    }
+}
```
BrianHarrisonAMD: I think this can be removed. 

Also, the general SUCCEED() inside caught exceptions is a bit concerning.
If we are expecting exceptions, then we have a macro that can be used to ensure the type + status matches what we expect.

If we aren't expecting exceptions, then we don't need try/catch, it'll fail if it throws unexpectedly.

backend/tests/test_engine_plugin_resource_manager.cpp
BrianHarrisonAMD, line 1: I think these tests end up needing to depend on loading real plugins due to not being able to mock the [Engine_plugin_manager](https://github.com/ROCm/hipDNN/blob/develop/backend/src/plugin/engine_plugin_resource_manager.hpp#L53-L54).

It would probably be worth moving the declaration for Engine_plugin_manager to the header so that it can be mocked. If we did that, then we could make a mocked Engine_plugin_manager and avoid loading real plugins for these tests.
That might simplify some parts of this.

Ideally this test is isolated from the plugin's actually getting loaded, and instead is just focused on the engine_plugin_resource_manager functionality.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+    EXPECT_GT(successful_creations.load(), 0);
+
+    for(const auto& thread_managers : all_managers)
+    {
+        for(const auto& rm : thread_managers)
+        {
+            EXPECT_NE(rm, nullptr);
+        }
+    }
+}
```
BrianHarrisonAMD: I think it's good to have this threaded test just to verify it works properly in a threaded context.

The exceptions shouldn't be ignored, and should be fails. 
Since it's threaded I am not sure the best way to propagate the errors, but we shouldn't be catching and ignoring exceptions. No failures should occur in this case.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+
+TEST_F(Engine_plugin_resource_manager_stress_test, ConcurrentAccess_MultipleThreadsCreatingManagers)
+{
+    const int num_threads = 4;
+    std::vector<std::thread> threads;
+    std::vector<std::vector<std::shared_ptr<Engine_plugin_resource_manager>>> all_managers(
+        static_cast<size_t>(num_threads));
```
BrianHarrisonAMD: Okay this is thread safe since it pre-allocates a separate vector for each thread.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+            }
+        });
+    }
+
+    for(auto& thread : threads)
+    {
+        thread.join();
+    }
+
+    EXPECT_GT(successful_creations.load(), 0);
```
BrianHarrisonAMD: This should check that it's the expected number.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+            if(rm != nullptr)
+            {
+                managers.push_back(rm);
+                rm->set_stream(nullptr);
+            }
+        }
+    }
+
+    SUCCEED();
+}
```
BrianHarrisonAMD: We need to consolidate these.
There is several versions that are doing stress testing on creation / destruction of these.

I think the one threaded version makes sense, and maybe a single non-threaded version that creates multiples to ensure it's not sharing state that can be ran through ASAN.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+    auto original_ptr = rm.get();
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wself-move"
+    rm = std::move(rm);
+#pragma clang diagnostic pop
+
+    EXPECT_EQ(rm.get(), original_ptr);
+    EXPECT_NO_THROW(rm->set_stream(nullptr));
+}
```
BrianHarrisonAMD: This would be good to test on the resource manager itself.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,715 @@
...
+            std::vector<uint32_t> host_output(data_size, 0);
+
+            for(size_t i = 0; i < data_size; ++i)
+            {
+                host_input[i] = static_cast<uint32_t>(i + 1);
+            }
+
+            ASSERT_EQ(
+                hipMemcpy(input_device, host_input.data(), buffer_size, hipMemcpyHostToDevice),
+                hipSuccess);
```
BrianHarrisonAMD: I think this test is missing the execute part, and we probably want to mock this.
Ideally we have very minimal device tests. 

I think it's worth reworking this so we can use mocked objects, and strip out the device dependent tests.

backend/src/plugin/engine_plugin.hpp
```diff
@@ -21,33 +21,38 @@ class Engine_plugin : public Plugin_base
...
-                            int64_t engine_id,
-                            const hipdnnPluginConstData_t* op_graph,
-                            hipdnnPluginConstData_t* engine_details) const;
-    void destroy_engine_details(hipdnnEnginePluginHandle_t handle,
-                                hipdnnPluginConstData_t* engine_details) const;
-    size_t get_workspace_size(hipdnnEnginePluginHandle_t handle,
-                              const hipdnnPluginConstData_t* engine_config,
-                              const hipdnnPluginConstData_t* op_graph) const;
-    [[nodiscard]] hipdnnEnginePluginExecutionContext_t
+    virtual hipdnnEnginePluginHandle_t create_handle() const;
```
mousdahl-amd: Looks like we lost [[nodiscard]] during a merge.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,824 @@
...
+        .WillOnce(::testing::Return(std::vector<int64_t>{200}));
+    EXPECT_CALL(*mock_plugin2, destroy_handle(testing::Eq(hipdnnEnginePluginHandle_t(0xcafebabe))))
+        .Times(testing::AtMost(1));
+
+    Engine_plugin_resource_manager rm1(plugin_manager1);
+    Engine_plugin_resource_manager rm2(plugin_manager2);
+
+    rm2 = std::move(rm1);
+
+    EXPECT_NO_THROW(rm2.set_stream(nullptr));
```
mousdahl-amd, line 199: Wouldn't this succeed even if the move was incorrect?

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,824 @@
...
+    EXPECT_CALL(*mock_plugin, get_all_engine_ids())
+        .WillOnce(::testing::Return(std::vector<int64_t>{100}));
+    EXPECT_CALL(*mock_plugin, set_stream(hipdnnEnginePluginHandle_t(0xdeadbeef), nullptr));
+    EXPECT_CALL(*mock_plugin, destroy_handle(testing::Eq(hipdnnEnginePluginHandle_t(0xdeadbeef))));
+
+    Engine_plugin_resource_manager rm(plugin_manager);
+    
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wself-move"
+    rm = std::move(rm);  // Test self-move of the actual object
```
mousdahl-amd: This seems like a strange test.  What exactly is expected behavior here?

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,824 @@
...
+        EXPECT_CALL(*mock_plugin, destroy_handle(testing::Eq(hipdnnEnginePluginHandle_t(0xdeadbeef))));
+        
+        managers.push_back(new Engine_plugin_resource_manager(plugin_manager));
+    }
+    
+    for(size_t i = 0; i < managers.size(); ++i)
+    {
+        for(size_t j = i + 1; j < managers.size(); ++j)
+        {
+            EXPECT_NE(managers[i], managers[j]);
```
mousdahl-amd: This is testing that we haven't created the same pointer twice somehow?

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,824 @@
...
+            output->size = strlen(fake_engine_details) + 1;
+        }));
+
+    EXPECT_CALL(*mock_plugin, destroy_handle(testing::Eq(hipdnnEnginePluginHandle_t(0xdeadbeef))));
+
+    {
+        Engine_plugin_resource_manager resource_manager(plugin_manager);
+
+        // Test get_engine_details functionality
+        // Note: This will throw because our mock data is not a valid flatbuffer
```
mousdahl-amd: Let's chat about this.  I wonder if there's a way to get a valid flatbuffer engine details.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,824 @@
...
+
+        try
+        {
+            resource_manager.execute_op_graph(nullptr, // execution_plan descriptor
+                                              nullptr // variant_pack descriptor
+            );
+        }
+        catch(const Hipdnn_exception& e)
+        {
+            SUCCEED() << "execute_op_graph properly validated null descriptors: " << e.what();
```
mousdahl-amd: We should probably make an EXPECT_HIPDNN_EXCEPTION that checks that the exception was thrown for the encapsulated code and the result of get_status() on the exception.

mousdahl-amd (commented):
I've got a few questions about some of the tests.

backend/tests/test_engine_plugin_resource_manager.cpp
bghimireamd: basically self-move should not crash. Added more test to insure that.

backend/src/plugin/engine_plugin.hpp
bghimireamd: good catch!

backend/tests/test_engine_plugin_resource_manager.cpp
bghimireamd: added more tests

backend/tests/test_engine_plugin_resource_manager.cpp
bghimireamd: yes. Basically there should be no duplicates. used std::unordered instead to detect duplicate :p

backend/tests/test_engine_plugin_resource_manager.cpp
bghimireamd: good call! added `EXPECT_HIPDNN_EXCEPTION`

backend/src/plugin/engine_plugin.hpp
mousdahl-amd: Looks like the one on create_handle is still missing

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,813 @@
...
+            .WillOnce(::testing::Return(hipdnnEnginePluginHandle_t(0xdeadbeef)));
+        EXPECT_CALL(*mock_plugin, get_all_engine_ids())
+            .WillOnce(::testing::Return(std::vector<int64_t>{static_cast<int64_t>(100 + i)}));
+        EXPECT_CALL(*mock_plugin,
+                    destroy_handle(testing::Eq(hipdnnEnginePluginHandle_t(0xdeadbeef))));
+
+        managers.push_back(new Engine_plugin_resource_manager(plugin_manager));
+    }
+
+    std::unordered_set<Engine_plugin_resource_manager*> unique_pointers(managers.begin(),
```
mousdahl-amd: This is still a really strange test.  We're really just testing that `new` is giving us a unique memory address each time.   I'm having difficulty imagining an instance where this wouldn't be the case.

bghimireamd: I agree. Basically we are just testing if new gives "new" address. May be I can just remove this test.

mousdahl-amd (approved):

backend/src/plugin/plugin_core.hpp
```diff
@@ -286,7 +289,7 @@ class Plugin_manager_base
         }
     }
 
-    std::vector<Plugin> _plugins;
+    std::vector<std::shared_ptr<Plugin>> _plugins;
```
BrianHarrisonAMD, line 292: Swap to std::shared_ptr just to allow for mocking / inheritance?

backend/tests/descriptors/flatbuffer_test_utils.hpp
```diff
@@ -29,4 +30,13 @@ static flatbuffers::FlatBufferBuilder create_valid_graph()
     return builder;
 }
 
+[[maybe_unused]] static flatbuffers::FlatBufferBuilder
```
BrianHarrisonAMD: Do we need these [[maybe_unused]] tags?
Surprised these would be unused.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,767 @@
...
+        EXPECT_CALL(*mock_plugin, get_all_engine_ids())
+            .WillOnce(::testing::Return(std::vector<int64_t>{100}));
+        EXPECT_CALL(*mock_plugin, set_stream(hipdnnEnginePluginHandle_t(0xdeadbeef), nullptr));
+        EXPECT_CALL(*mock_plugin,
+                    destroy_handle(testing::Eq(hipdnnEnginePluginHandle_t(0xdeadbeef))));
+
+        {
+            Engine_plugin_resource_manager rm(plugin_manager);
+
+            EXPECT_NO_THROW(rm.set_stream(nullptr));
```
BrianHarrisonAMD: Why add the stream setter here?
Probably isn't needed.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,767 @@
...
+    EXPECT_EQ(successful_creations.load(), num_threads * managers_per_thread);
+
+    for(const auto& thread_managers : all_managers)
+    {
+        for(const auto& rm : thread_managers)
+        {
+            EXPECT_NE(rm, nullptr);
+            delete rm;
+        }
+    }
```
BrianHarrisonAMD: If you swapped to std::shared_ptr, then I think you could avoid this.

backend/tests/test_engine_plugin_resource_manager.cpp
```diff
@@ -1,42 +1,767 @@
...
+            100,
+            &fake_engine_config,
+            &mock_graph_desc);
+
+        EXPECT_NE(execution_context, nullptr);
+        EXPECT_NE(execution_context->get(), nullptr);
+    }
+}
+
+TEST(Engine_plugin_resource_manager, execute_op_graph)
```
BrianHarrisonAMD: Review some of the names, since I think this is probably meant to check that execute fails properly with nullptr?

BrianHarrisonAMD: I also think we should add a execute that is successful using mocked objects.

BrianHarrisonAMD:
Couple comments, but I think it looks pretty close!

Thanks!

BrianHarrisonAMD (approved):

backend/src/plugin/engine_plugin.hpp
```diff
@@ -21,33 +21,38 @@ class Engine_plugin : public Plugin_base
     // The constructor is protected to prevent direct instantiation of the class.
     Engine_plugin(Shared_library&& lib);
 
+    // We need this to allow mocking this class
+    Engine_plugin();
+
 public:
-    std::vector<int64_t> get_all_engine_ids() const;
```
averinevg, line 25: I placed `get_all_engine_ids()` at the top since it doesn't require the handle. All functions here were arranged in the order they're called.

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -170,23 +138,24 @@ Engine_plugin_resource_manager::Engine_plugin_resource_manager()
 }
 
 Engine_plugin_resource_manager::Engine_plugin_resource_manager(
-    std::shared_ptr<Engine_plugin_manager>& pm)
-    : _pm(pm)
+    std::shared_ptr<Engine_plugin_manager> pm)
 {
+    _pm = std::move(pm);
```
averinevg: Could be moved to the member initializer list.

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -241,6 +210,8 @@ void Engine_plugin_resource_manager::set_stream(hipStream_t stream) const
 std::vector<int64_t> Engine_plugin_resource_manager::get_applicable_engine_ids(
     const Graph_descriptor* graph_desc) const
 {
+    THROW_IF_NULL(graph_desc, HIPDNN_STATUS_BAD_PARAM, "Graph descriptor cannot be null");
```
averinevg: Using `assert()` is probably sufficient here.

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -275,9 +246,19 @@ void Engine_plugin_resource_manager::get_engine_details(
     const Graph_descriptor* graph_desc,
     hipdnnPluginConstData_t* engine_details) const
 {
+    THROW_IF_NULL(graph_desc, HIPDNN_STATUS_BAD_PARAM, "Graph descriptor cannot be null");
+    THROW_IF_NULL(engine_details, HIPDNN_STATUS_BAD_PARAM, "Engine details cannot be null");
```
averinevg: Ditto

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -312,9 +293,19 @@ size_t
                                                        const hipdnnPluginConstData_t* engine_config,
                                                        const Graph_descriptor* graph_desc) const
 {
+    THROW_IF_NULL(engine_config, HIPDNN_STATUS_BAD_PARAM, "Engine config cannot be null");
+    THROW_IF_NULL(graph_desc, HIPDNN_STATUS_BAD_PARAM, "Graph descriptor cannot be null");
```
averinevg: Ditto

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -327,9 +318,19 @@ hipdnnEnginePluginExecutionContext_t Engine_plugin_resource_manager::create_exec
     const hipdnnPluginConstData_t* engine_config,
     const Graph_descriptor* graph_desc) const
 {
+    THROW_IF_NULL(engine_config, HIPDNN_STATUS_BAD_PARAM, "Engine config cannot be null");
+    THROW_IF_NULL(graph_desc, HIPDNN_STATUS_BAD_PARAM, "Graph descriptor cannot be null");
```
averinevg, line 321: Ditto

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -275,9 +246,19 @@ void Engine_plugin_resource_manager::get_engine_details(
...
 {
+    THROW_IF_NULL(graph_desc, HIPDNN_STATUS_BAD_PARAM, "Graph descriptor cannot be null");
+    THROW_IF_NULL(engine_details, HIPDNN_STATUS_BAD_PARAM, "Engine details cannot be null");
+
+    auto it = _engine_id_to_handle.find(engine_id);
+    if(it == _engine_id_to_handle.end())
+    {
+        throw Hipdnn_exception(HIPDNN_STATUS_INTERNAL_ERROR,
+                               "Invalid engine ID: " + std::to_string(engine_id));
+    }
```
averinevg, line 256: As I understand it, this code was modified to use an internal hipDNN exception instead of `std::out_of_range`?

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -312,9 +293,19 @@ size_t
...
 {
+    THROW_IF_NULL(engine_config, HIPDNN_STATUS_BAD_PARAM, "Engine config cannot be null");
+    THROW_IF_NULL(graph_desc, HIPDNN_STATUS_BAD_PARAM, "Graph descriptor cannot be null");
+
+    auto it = _engine_id_to_handle.find(engine_id);
+    if(it == _engine_id_to_handle.end())
+    {
+        throw Hipdnn_exception(HIPDNN_STATUS_INTERNAL_ERROR,
+                               "Invalid engine ID: " + std::to_string(engine_id));
+    }
```
averinevg, line 303: Ditto

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -327,9 +318,19 @@ hipdnnEnginePluginExecutionContext_t Engine_plugin_resource_manager::create_exec
...
```
averinevg, line 328: Ditto

backend/src/plugin/engine_plugin_resource_manager.hpp
```diff
@@ -12,6 +12,7 @@
 #include <hip/hip_runtime.h>
 #include <hipdnn_sdk/plugin/plugin_api_data_types.h>
 
+#include "engine_plugin.hpp"
```
averinevg: [R] I'd prefer to keep the forward declaration.

backend/src/plugin/plugin_core.hpp
```diff
@@ -248,11 +251,11 @@ class Plugin_manager_base
                 return;
             }
 
-            Plugin plugin(std::move(lib));
+            std::shared_ptr<Plugin> plugin = std::shared_ptr<Plugin>(new Plugin(std::move(lib)));
```
averinevg, line 254:
```suggestion
           auto plugin = std::make_shared<Plugin>(std::move(lib));
```

mousdahl-amd: For some reason I wasn't able to get that to compile.  I'll try again to see if I can't figure it out.

backend/tests/descriptors/test_macros.hpp
```diff
@@ -23,4 +23,30 @@
...
+        {                                                                        \
+            FAIL() << "Expected hipdnn_backend::Hipdnn_exception";               \
+        }                                                                        \
+        EXPECT_TRUE(exception_thrown) << "Expected exception not thrown";        \
+        if(exception_thrown)                                                     \
+        {                                                                        \
+            EXPECT_EQ(actual_status, status)                                     \
+                << "Expected status " << status << " but got " << actual_status; \
+        }                                                                        \
+    } while(0)
```
BrianHarrisonAMD: Actually seeing this now, do we need this?

I thought the ASSERT_THROW_HIPDNN_STATUS would cover us.
Surprised there would be cases we want to continue after a throw, but I suppose that's possible.

backend/src/plugin/engine_plugin_resource_manager.cpp
BrianHarrisonAMD: These should be internal errors if we don't swap them to asserts.
BAD_PARAM is for cases where the user provides a bad value, and this is certainly a developer mistake if we get this far with a nullptr.

I am a bit torn on using asserts over the runtime checks.
I think the runtime cost is basically 0 for these checks, and they can prevent crashes if we have a regression that slips through, but I also agree we don't expect this to ever occur.

BrianHarrisonAMD: I guess I am fine either way, but my preference would be keeping the check and throwing an internal error.
If the check was slower, or we ever see these extra checks are impactful E2E, then I would agree with swapping to asserts.

backend/src/plugin/plugin_core.hpp
bghimireamd: I attempted this change but it causes a compilation error because std::make_shared requires public access to the constructor, while Engine_plugin has a protected constructor by design.

backend/src/plugin/engine_plugin_resource_manager.cpp
bghimireamd: yes for better error handling basically

backend/tests/descriptors/test_macros.hpp
bghimireamd: I removed this.

backend/src/plugin/plugin_core.hpp
mousdahl-amd: Ah, yes.  The argument forwarding from make_shared fails for really arcane reasons

PR #120, opened 2025-08-12
Add checking engine IDs on plugin load

averinevg:
## Proposed changes

This PR adds checking engine IDs on plugin load

## Checklist


backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -28,6 +28,37 @@ class Engine_plugin_manager : public Plugin_manager_base<Engine_plugin>
...
+        for(const auto id : engine_ids)
+        {
+            if(_engine_ids.contains(id))
+            {
+                throw Hipdnn_exception(HIPDNN_STATUS_PLUGIN_ERROR,
+                                       "Engine ID " + std::to_string(id)
+                                           + " already exists in the list");
+            }
+        }
+    }
```
SamuelReeder, line 45: I see that `action_after_adding` checks for duplicate engine IDs internal to plugin, but maybe its better to do that here so it's always handled before finishing the `load_plugin_from_file` and the error is caught and logged. I think this would be more consistent.

Then `action_after_adding` can be changed to only insert the engine ID since it will be guaranteed there are no duplicates if this is called first, and you won't have duplicate-duplicate checks.

 `load_plugin_from_file` might have to use a lock on engine IDs though, unless that's not a concern.

SamuelReeder (approved):
Can consider my one suggestion, but LGTM!

backend/src/plugin/engine_plugin.cpp
```diff
@@ -67,6 +71,36 @@ void Engine_plugin::resolve_symbols()
 #endif
 }
 
+std::vector<int64_t> Engine_plugin::get_all_engine_ids() const
+{
+    assert(_initialized);
+
+    uint32_t max_engines = 64;
```
SamuelReeder: Not super important, but the 64 could be a constant somewhere if that is a suitable "first guess" for a buffer size.

backend/src/plugin/engine_plugin_resource_manager.cpp
averinevg: I will add caching to the `get_all_engine_ids()` and keep the duplicate check only in `validate_before_adding()`.

backend/src/plugin/engine_plugin.cpp
BrianHarrisonAMD: Can probably add it to the docs in the future for details on adding a plugin.

backend/src/plugin/engine_plugin_resource_manager.cpp
```diff
@@ -115,6 +146,18 @@ Engine_plugin_resource_manager::Engine_plugin_resource_manager(
...
+        for(const auto id : engine_ids)
+        {
+            if(_engine_id_to_handle.contains(id))
+            {
+                throw Hipdnn_exception(HIPDNN_STATUS_PLUGIN_ERROR,
+                                       "Engine ID " + std::to_string(id)
+                                           + " already exists in the resource manager");
+            }
+            _engine_id_to_handle[id] = handle;
+        }
```
BrianHarrisonAMD: If there is a duplicate we should probably not track the plugin anymore, or we can add the non-overlapping Ids I guess. 

Right now the plugin will be in the handle_to_plugin lookup, and potentially partially in the _engine_id_to_handle look-up after failing.

backend/src/plugin/hipdnn_plugin_base.hpp
```diff
@@ -1,32 +0,0 @@
...
-{
-struct Hipdnn_plugin_base
-{
-    virtual ~Hipdnn_plugin_base() = default;
-    // Interface for plugin interaction using internal types.
-    // Simplifies usage and integration with higher-level API handlers.
-    virtual void execute(const Graph_descriptor* graphdesc,
-                         const Variant_descriptor* vpack,
-                         hipdnnHandle* handle)
-        = 0;
```
BrianHarrisonAMD, line 21: Oh did we just forget to remove this?

plugins/miopen_legacy_plugin/tests/test_miopen_legacy_engine_plugin_api.cpp
```diff
@@ -13,6 +13,34 @@
...
+    std::array<int64_t, 1> engine_ids = {0};
+    uint32_t num_engines = 0;
+
+    //get max 1 engine
+    auto status = hipdnnEnginePluginGetAllEngineIds(engine_ids.data(), 1, &num_engines);
+
+    EXPECT_EQ(status, HIPDNN_PLUGIN_STATUS_SUCCESS);
+    EXPECT_EQ(num_engines, 1u);
+    EXPECT_EQ(engine_ids[0], 1u);
+}
```
BrianHarrisonAMD: I think we should add a test to cover getting the count back, (called with 0 max), since that's an alternative way we are using this API.

sdk/include/hipdnn_sdk/plugin/engine_plugin_api.h
```diff
@@ -27,6 +27,26 @@ extern "C" {
...
+ * @param[out] num_engines A pointer to a variable where the total number of applicable engines will be stored.
+ *                         This value may exceed `max_engines` if more engines are applicable than the array can
+ *                         accommodate.
+ *
+ * @return A value of type `hipdnnPluginStatus_t` indicating the status of the operation.
+ *
+ * @note The caller is responsible for ensuring that the `engine_ids` array is large enough to hold up to
+ *       `max_engines` IDs. If the number of applicable engines exceeds `max_engines`, only the first
+ *       `max_engines` IDs will be returned, and the total count will be stored in `num_engines`.
+ */
```
BrianHarrisonAMD: Might be good to explicitly call out that it's expected you can call this API with 0 max_engines to get back the full count, but it should work that way from this description.

backend/src/plugin/engine_plugin_resource_manager.cpp
BrianHarrisonAMD: I guess it's assumed this cannot happen, and we are just being extra careful.

BrianHarrisonAMD (approved):
Looks good to me with some small feedback to consider.

backend/src/plugin/hipdnn_plugin_base.hpp
averinevg: Yep

mousdahl-amd (approved):

backend/src/plugin/engine_plugin.cpp
```diff
@@ -67,6 +72,51 @@ void Engine_plugin::resolve_symbols()
...
+    }
+
+    uint32_t num_engines = 0;
+    invoke_plugin_function(
+        "get number of engines", _func_get_all_engine_ids, nullptr, 0u, &num_engines);
+
+    if(num_engines == 0)
+    {
+        throw Hipdnn_exception(HIPDNN_STATUS_PLUGIN_ERROR, "No engines found in the plugin");
+    }
```
BrianHarrisonAMD: I think this can be swapped to the THROW_IF_NE macro. 
Same with other exceptions, I think there is ways to swap some of these to the macro usage.

backend/src/plugin/engine_plugin.cpp
```diff
@@ -67,6 +72,51 @@ void Engine_plugin::resolve_symbols()
...
+                           max_engines,
+                           &num_engines);
+
+    if(num_engines != max_engines)
+    {
+        throw Hipdnn_exception(HIPDNN_STATUS_PLUGIN_ERROR,
+                               "Number of engines returned does not match expected count");
+    }
+
+    engine_ids.resize(num_engines);
```
BrianHarrisonAMD: Don't think we need the resize here, because we throw if they aren't equal, and pre-allocate it to be max_engines.

backend/src/plugin/engine_plugin.cpp
```diff
@@ -105,21 +160,33 @@ std::vector<int64_t>
...
+
+    std::ranges::sort(engine_ids);
+    if(std::ranges::adjacent_find(engine_ids) != engine_ids.end())
+    {
+        throw Hipdnn_exception(HIPDNN_STATUS_PLUGIN_ERROR, "Duplicate engine IDs found");
+    }
+
+    for(const auto engine_id : engine_ids)
+    {
+        if(std::ranges::find(_all_engine_ids, engine_id) == _all_engine_ids.end())
```
BrianHarrisonAMD, line 177: I think it would be better to make _all_engine_ids a set since we are forcing it's a unique vector anyways, and has better lookup complexity for contains in general.

That being said, the list should be small, and not sure it really matters a ton in practice. 
I guess it might also make it more annoying to work with in other areas of the code if the vector is useful for populating the returned values through the public API. 

I guess I am fine either way hehe.

backend/src/plugin/engine_plugin.hpp
```diff
@@ -36,7 +37,7 @@ class Engine_plugin : public Plugin_base
     size_t get_workspace_size(hipdnnEnginePluginHandle_t handle,
                               const hipdnnPluginConstData_t* engine_config,
                               const hipdnnPluginConstData_t* op_graph) const;
-    hipdnnEnginePluginExecutionContext_t
+    [[nodiscard]] hipdnnEnginePluginExecutionContext_t
```
BrianHarrisonAMD, line 40: Ah we should keep this in mind in the future if we are returning back something that is expected to be owned by the caller.

BrianHarrisonAMD (approved):
Looks generally good to me.
I think it's worth using the macros for the throws.

PR #121, opened 2025-08-12
[ALMIOPEN-119] - Windows Build

mousdahl-amd:
## Proposed changes
- Modified ClangToolChain.cmake to work on Windows using the dist folders from TheRock
- Edit .clang-tidy to make it more platform independent
- Since we're compiling with clang on windows, __declspec(dllimport / dllexport) don't work for exports, so we had to use a different mechanism for now `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS TRUE)`.  This may expose more than we want, so we can add a story to look into using a `.def` file.
- Added some more platform-portable methods (get_env, set_env, path utilities, etc)
- Fixed up tests that assumed some behaviors

## Checklist


backend/src/CMakeLists.txt
```diff
@@ -69,11 +66,6 @@ set_target_properties(hipdnn_backend PROPERTIES CXX_VISIBILITY_PRESET hidden)
 
 clang_tidy_check(hipdnn_backend)
 
-include( GenerateExportHeader )
-generate_export_header( hipdnn_backend 
-                        EXPORT_FILE_NAME 
-                        ${HIP_DNN_BACKEND_EXPORT_INCLUDE_DIR}/hipdnn_backend_export.h )
```
mousdahl-amd: This really helps if we're using MSVC for compiling on Windows.  But since we're so HIP-heavy, we have to use clang++ from the rocm build, and I could not for the life of me get it to honor __declspec(dllimport), even with the `-fms-extensions` compiler args

frontend/tests/test_graph.cpp
```diff
@@ -1493,10 +1493,14 @@ TEST_F(Graph_test_fixture, ExecutePacksVariantPackAndPassesTheCorrectArguments)
...
-            for(const auto& value : std::views::values(variant_pack))
+            for(int i = 0; i < 4; i++)
             {
-                EXPECT_EQ(data_ptrs[i++], value);
+                auto targetValue = data_ptrs[i];
+                auto it = std::find_if(
+                    variant_pack.begin(), variant_pack.end(), [&targetValue](const auto& pair) {
+                        return pair.second == targetValue;
+                    });
+                EXPECT_TRUE(it != variant_pack.end());
```
mousdahl-amd, line 1499: This was doing an order-dependant check of an unordered map.  It worked by accident before.

plugins/miopen_legacy_plugin/integration_tests/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ if(NOT BUILD_PLUGIN_AS_DEPENDENCY)
 endif()
 
 include(GoogleTest)
+find_package(Threads REQUIRED)
```
mousdahl-amd, line 9: This is the platform-independent way to include threading.

CMakeLists.txt
```diff
@@ -98,9 +99,7 @@ list(PREPEND HIPDNN_WARNING_COMPILE_OPTIONS
     -Wswitch-default                         # Warn if a switch statement does not have a default case
 )
 
-add_compile_options(
-    -fPIC # Position Independent Code
-    ${EXTRA_COMPILE_OPTIONS})
+add_compile_options(${EXTRA_COMPILE_OPTIONS})
```
BrianHarrisonAMD, line 102: -fPIC was causing issues?

sdk/include/hipdnn_sdk/plugin/plugin_api.h
```diff
@@ -6,9 +6,7 @@
 #include <hipdnn_sdk/logging/callback_types.h>
 #include <hipdnn_sdk/plugin/plugin_api_data_types.h>
 
-#ifdef _WIN32
-#define HIPDNN_PLUGIN_EXPORT __declspec(dllexport)
```
mousdahl-amd, line 10: Again, couldn't get __declspec(dllexport) to work.  We can look at using `.def` files on windows.

backend/include/hipdnn_backend_export.h
BrianHarrisonAMD, line 1: Does this need to be committed & not generated now?

I think it was generated before, and surprised to see we need to commit it now.

CMakeLists.txt
BrianHarrisonAMD: Oh looks like we probably didn't need this generally.

backend/src/CMakeLists.txt
BrianHarrisonAMD: Ah okay that explains it.

BrianHarrisonAMD: Alright so we are just make it ourselves now and making it work cross-plat?

backend/tests/test_backend_logger.cpp
```diff
@@ -24,66 +25,46 @@ class Backend_logging_test : public ::testing::Test
...
 
-        // pipe stderr to capture log output
-        _old_stderr = dup(STDERR_FILENO);
-        ASSERT_NE(_old_stderr, -1);
-        ASSERT_EQ(pipe(_stderr_pipe.data()), 0);
-        ASSERT_NE(dup2(_stderr_pipe[1], STDERR_FILENO), -1);
-        ASSERT_EQ(close(_stderr_pipe[1]), 0);
-        setenv("HIPDNN_LOG_LEVEL", "off", 1);
-        unsetenv("HIPDNN_LOG_FILE");
+        testing::internal::CaptureStderr();
```
BrianHarrisonAMD, line 28: I think generally this shouldn't be leveraged / used since it's private, but I think this is probably fine for now.
Maybe we can backlog a task to clean-up usage of this, but I think it's low priority.

cmake/ClangToolChain.cmake
```diff
@@ -12,7 +13,40 @@ if(UNIX)
...
         set(CMAKE_CXX_COMPILER ${ROCM_LLVM_BIN_DIR}/clang++)
         set(CMAKE_SYMBOLIZER ${ROCM_LLVM_BIN_DIR}/llvm-symbolizer)
+        message(STATUS "Using ROCm Clang compilers from ${ROCM_LLVM_BIN_DIR}")
     else()
         message(WARNING "The directory /opt/rocm/llvm/bin does not exist. Cannot auto select clang compilers.")
     endif()
-endif()
\ No newline at end of file
+
+    add_compile_options(-fPIC) # Position Independent Code (not needed/supported on Windows)
```
BrianHarrisonAMD, line 21: Oh this is where it went.

cmake/ClangToolChain.cmake
```diff
@@ -12,7 +13,40 @@ if(UNIX)
...
+
+elseif(WIN32)
+    # Windows: Use Clang from TheRock build
+    set(WINDOWS_ROCM_DIR "C:/src/TheRock/build/dist/rocm" CACHE PATH "Path to Windows ROCm installation")
```
BrianHarrisonAMD, line 25: Hmmm, I guess there is no Windows equivalent we can use, and this is how it has to be for now?

cmake/ClangToolChain.cmake
```diff
@@ -12,7 +13,40 @@ if(UNIX)
...
+    set(WINDOWS_ROCM_LLVM_BIN_DIR "${WINDOWS_ROCM_DIR}/lib/llvm/bin")
+    set(WINDOWS_ROCM_CMAKE_DIR "${WINDOWS_ROCM_DIR}/lib/cmake")
+    set(CMAKE_SYSTEM_VERSION "10.0.22621.0")
+    set(CMAKE_RC_COMPILER rc.exe)
+
+    # Enable exporting all symbols for Windows
+    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
```
BrianHarrisonAMD: How bad are the exports if we don't have this?
I suppose it's fine since we are manually stripping the backend.so, and using an internal static lib that has all symbols exposed. 

Might be able to enable this flag just for the static backend build.

CMakeLists.txt
mousdahl-amd: Originally I was trying to build with msvc, and it didn't like this flag.  When I realized that *everything* needs to be built with clang++ (because HIP), I probably could've put it back.  But if we don't need it, yoink it?

cmake/ClangToolChain.cmake
```diff
@@ -12,7 +13,40 @@ if(UNIX)
...
+        message(WARNING "The directory ${WINDOWS_ROCM_LLVM_BIN_DIR} does not exist. Cannot auto select clang compilers.")
+    endif()
+
+    # Set up CMake package search path for TheRock build
+    if(EXISTS ${WINDOWS_ROCM_CMAKE_DIR})
+        # Add to CMAKE_PREFIX_PATH for find_package() searches
+        list(APPEND CMAKE_PREFIX_PATH ${WINDOWS_ROCM_CMAKE_DIR})
+        message(STATUS "Added Windows ROCm CMake package search path: ${WINDOWS_ROCM_CMAKE_DIR}")
+    else()
+        message(WARNING "Windows ROCm CMake directory not found: ${WINDOWS_ROCM_CMAKE_DIR}")
```
BrianHarrisonAMD: Can we make these error cases?
Seems like it's a busted env if we can't find the ROCm dirs.

cmake/ClangToolChain.cmake
mousdahl-amd: Good idea.  I'll try it!

backend/src/CMakeLists.txt
mousdahl-amd: Pretty much.

docs/Building.md
```diff
@@ -85,4 +85,16 @@ When building with ninja dont forget to add the -jnproc flag for using additiona
...
+1. Follow the instructions for building TheRock
+2. Open up a "x64 Native Tools Command Prompt for VS 2022" command shell from the start menu
+3. From the native tools prompt, clone and build TheRock
+4. Set your HIP_PLATFORM: `set HIP_PLATFORM=amd`
+5. Set your TheRock dist location: `set PATH=C:\src\hipDNN\build\backend\src;C:\src\TheRock\build\dist\rocm\bin;%PATH%`
+    - If your locations differ, you may need to edit \<src\>/cmake/ClangToolChain.cmake
+    - This is done so that the executables / dlls can find their dependency dlls.
+6. From the native tools prompt, clone and build hipDNN
+    - `cmake -GNinja -DHIP_DNN_BUILD_PLUGINS=OFF ..`
+    - `ninja check`
```
BrianHarrisonAMD: Nice!

Should we add a note saying that HIP & GPU related parts are still not working?

frontend/tests/test_graph.cpp
```diff
@@ -1493,10 +1493,14 @@ TEST_F(Graph_test_fixture, ExecutePacksVariantPackAndPassesTheCorrectArguments)
...
             }
```
BrianHarrisonAMD, line 1500: Why did we have to swap this?
No good in Windows?

backend/tests/test_backend_logger.cpp
mousdahl-amd: People were going back and forth on this online.  I tried writing a version of it myself, but it kept behaving oddly (like hanging occasionally) depending on whether or not I was using the commandline tests or C++TestMate.  This seemed like the lesser evil.

mousdahl-amd: I'll add a backlog item.

frontend/tests/test_graph.cpp
BrianHarrisonAMD: Actually, the way it was before seems a bit sussy since it's assuming an order.

cmake/ClangToolChain.cmake
adickin-amd: i think this should stay in the main cmakelists file if possible.  The toolchain file should be to setup the compilers and everything we use to build.  The actual configuration and compile options should live in the main cmakelists.txt.

cmake/ClangToolChain.cmake
adickin-amd: you probably need this for the miopen plugin at the moment.  With symbol hiding I just realized I need to split the plugin build like the backend is done so testing can happen correctly.

sdk/include/hipdnn_sdk/utilities/string_util.hpp
```diff
@@ -16,7 +16,11 @@ static void
         return;
     }
 
+#ifdef _WIN32
+    strncpy_s(destination, max_size, source, max_size - 1);
+#else
     std::strncpy(destination, source, max_size - 1);
+#endif
```
BrianHarrisonAMD, line 23: I think these can have different behaviors, is strncpy problematic on Windows?

I think it's generally a bit problematic, but there wasn't a better cross-plat alternative I could find.

sdk/include/hipdnn_sdk/plugin/plugin_api.h
adickin-amd: did the exporting work for the backend library or just no library on windows?

sdk/include/hipdnn_sdk/utilities/string_util.hpp
BrianHarrisonAMD: Ah nevermind yea it's strncpy_s() should be used.

BrianHarrisonAMD: Can we make it use strncpy_s in both cases, or it's only possible for Windows?

This feels like Déjà vu.

cmake/ClangToolChain.cmake
BrianHarrisonAMD: Do we even need it?
I thought the cmake functions handled this for us automagically now.

BrianHarrisonAMD (approved):
Looks good to me with a few small things you can decide on.

Awesome stuff!

cmake/ClangToolChain.cmake
adickin-amd: dont need the fPIC?  I seem to remember the build failing without it but if we dont need it anymore I dont see why we should keep it.

sdk/include/hipdnn_sdk/plugin/plugin_api.h
mousdahl-amd: I'm going to play around a little more and find out.

cmake/ClangToolChain.cmake
mousdahl-amd: Btw, without this, CMake / clang can't sort out which symbols to import, but I'm going to try a `.def` file to do more targeted exports.

cmake/ClangToolChain.cmake
BrianHarrisonAMD: I thought CMake automagically adds it for you on Linux.

cmake/ClangToolChain.cmake
BrianHarrisonAMD: I seem to remember we hit something similar when @adickin-amd originally added the stripping for public API to the backend libhipdnn.so .
I am fine with windows exporting everything as long as they get stripped in the final public .dll. 
I don't see any value in carefully controlling exports if it's all internal, and it works without needing to manage it manually.

sdk/include/hipdnn_sdk/plugin/plugin_api.h
mousdahl-amd: Okay, @adickin-amd and I figured out how to get this back without exposing a ridiculous amount of symbols in windows!

sdk/include/hipdnn_sdk/utilities/string_util.hpp
mousdahl-amd: Implementation of strncpy_s is iffy on linux distros, it sounds like.

cmake/ClangToolChain.cmake
mousdahl-amd: Got things working without this.

frontend/include/hipdnn_frontend/backend/backend_wrapper.hpp
```diff
@@ -110,20 +110,17 @@ class Hipdnn_backend_wrapper : public Hipdnn_backend_interface
...
-    {
-        std::lock_guard<std::mutex> lock(backend_wrapper_mutex);
-        if(!backend_wrapper)
-        {
-            backend_wrapper = std::make_unique<Hipdnn_backend_wrapper>();
-        }
-    }
-    return *backend_wrapper;
+    return *get_backend_instance();
 }
```
mousdahl-amd, line 122: @BrianHarrisonAMD Adam and I were curious about your thoughts on the thread safety here?  We had to play around a bit since statics in a header file (outside an inline func) are no bueno.

BrianHarrisonAMD: Should be fine as initialization of function scoped static variables is thread safe AFAIK in C++, and it's just used as a handle to call the C API without any state.

BrianHarrisonAMD: Hmm looking at this again, I think if we wanted to be fully safe, we should not have a way to get the Hipdnn_backend_interface&, and only the std::shared_ptr<Hipdnn_backend_interface>.

If this was multi-threaded, and someone sets a new backend_instance, then the & will be cleaned up and is dangerous.

This is obviously a very niche unexpected case though.

BrianHarrisonAMD: To be sure it's safe we would want:
```
inline std::shared_ptr<Hipdnn_backend_interface> get_backend_instance()

inline void set_backend_instance(std::shared_ptr<Hipdnn_backend_interface> backend)
```

That way if someone gets access to a backend, and it *somehow* was replaced, then the old one will still exist, and not cause undefined behavior.

PR #123, opened 2025-08-14
Batchnorm bwd cpu ref

SamuelReeder:
## Proposed changes

Batchnorm bwd cpu ref implementation and tests.

## Checklist


sdk/tests/test_utilities/test_cpu_fp_reference_implementation.cpp
```diff
@@ -13,87 +13,281 @@ using namespace hipdnn_sdk::reference_test_utilities;
...
 
     Cpu_fp_reference_implementation<float, float, float> ref_impl;
 
     ref_impl.batchnorm_fwd_inference(
         input_tensor, scale_tensor, bias_tensor, mean_tensor, variance_tensor, output_tensor, 1e-5);
 }
+
+TEST(CpuFpReferenceImplementation, BatchnormBwdFloatUsage)
+{
+    SKIP_IF_NO_DEVICES();
```
SamuelReeder, line 103: I have the `SKIP_IF_NO_DEVICES()` here because the rest of the ref tests have them, but I don't get why we use them here since it doesn't involve a device.

plugins/miopen_legacy_plugin/integration_tests/batchnorm_fwd_inference_integration_test.cpp
```diff
@@ -284,17 +284,16 @@ class Batchnorm_forward_inference_integration_test
...
+        {.n = 1, .c = 3, .h = 14, .w = 14} // {.n = 2, .c = 3, .h = 14, .w = 14},
+        // {.n = 64, .c = 3, .h = 14, .w = 14},
+        // {.n = 64, .c = 256, .h = 14, .w = 14},
+        // {.n = 64, .c = 256, .h = 28, .w = 28},
+        // {.n = 64, .c = 256, .h = 56, .w = 56},
+        // {.n = 64, .c = 512, .h = 14, .w = 14},
+        // {.n = 64, .c = 512, .h = 28, .w = 28},
+        // {.n = 64, .c = 512, .h = 7, .w = 7},
+        // {.n = 64, .c = 64, .h = 112, .w = 112},
+        // {.n = 64, .c = 64, .h = 56, .w = 56},
```
SamuelReeder: Oops will uncomment.

sdk/tests/test_utilities/test_cpu_fp_reference_implementation.cpp
```diff
@@ -13,87 +13,281 @@ using namespace hipdnn_sdk::reference_test_utilities;
...
+                           x_tensor,
+                           mean_tensor,
+                           inv_variance_tensor,
+                           scale_tensor,
+                           dx_tensor,
+                           dscale_tensor,
+                           dbias_tensor);
+}
+
+TEST(CpuFpReferenceImplementation, BatchnormBwdSanityValidation)
```
SamuelReeder, line 226: This is no golden data but I wanted a clear sanity check. Can remove if its pointless to keep.

sdk/tests/test_utilities/test_cpu_fp_reference_implementation.cpp
mousdahl-amd: Have you tried running ASAN without it?  If it works, that's great.

sdk/tests/test_utilities/test_cpu_fp_reference_implementation.cpp
mousdahl-amd: I like it!

adickin-amd: I think this is great! Very clear comments to show how the expected results are calculated.

mousdahl-amd (approved):
LGTM, remove SKIP_IF_NO_DEVICES if possible

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
```diff
@@ -83,6 +83,103 @@ class Cpu_fp_reference_implementation : public Reference_implementation_interfac
...
+
+            // Calculate dot product for (x - mean) * channel_inv_variance * dy and ∑ dy for this channel
+            Mean_variance_data_type dot_product = 0;
+            Mean_variance_data_type sum_dy = 0;
+
+            for(int row = 0; row < height; row++)
+            {
+                for(int column = 0; column < width; column++)
+                {
+                    for(int bidx = 0; bidx < n_batches; bidx++)
```
SamuelReeder: Would it be better to go batch -> row -> col instead for mem locality?

sdk/tests/test_utilities/test_cpu_fp_reference_implementation.cpp
SamuelReeder: Ooh, got same results in pytorch `batch_norm` too:
```
dbias:  1.000000000000000
dscale: 0.447211806656309

dx tensor:
tensor([[[[-2.1466e-06, -7.1553e-07],
          [ 7.1553e-07,  2.1466e-06]]]], dtype=torch.float64)
```

sdk/tests/test_utilities/test_cpu_fp_reference_implementation.cpp
SamuelReeder: It still fails so I'll keep them, must be `hipHostMalloc` or something?

mousdahl-amd: That makes sense, and reminds me.  Adam was suggesting we change Migratable_memory to use plain old malloc for host allocations, and that would allow us to get rid of the SKIP_IF_NO_DEVICES() checks for these CPU centric tests.  I'm doing a rework of Migratable_memory in my issue, so I'll add a note to try that change and then come back and remove these.

mousdahl-amd: Okay, note added to my story.  It's now my problem!

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
BrianHarrisonAMD: Yea

BrianHarrisonAMD: Best is to go by smallest stride between the 3.

Which will be batch -> row -> col for NCHW & NHWC.

sdk/tests/test_utilities/test_cpu_fp_reference_implementation.cpp
BrianHarrisonAMD: Yea any hip call is problematic even hipHostMalloc.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
```diff
@@ -83,6 +83,103 @@ class Cpu_fp_reference_implementation : public Reference_implementation_interfac
...
+                       const Tensor& invVariance,
+                       const Tensor& scale,
+                       Tensor& dx,
+                       Tensor& dscale,
+                       Tensor& dbias) override
+    {
+        if(x.dims().size() != 4)
+        {
+            throw std::runtime_error("Batchnorm backward requires a 4D tensor.");
+        }
```
BrianHarrisonAMD, line 99: I guess since it's the SDK we can't use our exception class?

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
```diff
@@ -83,6 +83,103 @@ class Cpu_fp_reference_implementation : public Reference_implementation_interfac
         output.memory().mark_host_modified(); // Mark output memory as modified on host
     }
 
+    void batchnorm_bwd(const Tensor& dy,
+                       const Tensor& x,
+                       const Tensor& mean,
+                       const Tensor& invVariance,
```
BrianHarrisonAMD: Does this not flag a format issue?

I thought it needed to be `inv_variance`

BrianHarrisonAMD (approved):
Looks generally good to me.

Can improve the loop ordering, and maybe double check formatting since I am surprised it's not flagging.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
SamuelReeder: Yeah, there's currently no sdk exception class. But there could be!

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_implementation.hpp
SamuelReeder: You're right, but it doesn't flag either way. It is formatting these files, but I don't think all clang tidy rules are being enforced on them... I'll see if its a quick fix.

SamuelReeder: Was a missing `clang_tidy_check(...)`. Turning it on reveals new issues, so I'll either fix here or in another PR depending on how big.

PR #128, opened 2025-08-20
Migratable_memory upgrades

mousdahl-amd:
## Proposed changes
- Added the ability to specify host and device allocators as template args to Migratable_memory
- Added async buffer retrieval
- Added streaming buffer retrieval
- Fixed up all the unit tests, and removed SKIP_IF_NO_DEVICES from many since the default allocator is now malloc instead of hipHostMalloc

## Checklist


.clang-tidy
```diff
@@ -21,7 +21,8 @@ Checks: >
   -readability-uppercase-literal-suffix,
   -readability-avoid-unconditional-preprocessor-if,
   -readability-use-anyofallof,
-  -readability-function-cognitive-complexity
+  -readability-function-cognitive-complexity,
+  -portability-template-virtual-member-function
```
mousdahl-amd: This comes up because I've got a templated base interface inherited by a templated derived class.  It's not an issue in clang, but it warns because some versions of MSVC can't cope.  Since we'll be using clang on Windows (since we have to compile HIP code), I thought this might be okay, but I wanted to check in with everyone else.

BrianHarrisonAMD: Id prefer to suppress this just on the utility class itself for now.
The other cases that flagged this, it made sense to do it in a different way to avoid this issue.
Which file is it?
I am curious if there is an alternative way that we could solve this without suppressing.

BrianHarrisonAMD: Allocators I assume?

mousdahl-amd: No, the reference validators / implementations.  I briefly toyed with CRTP, but then the interface doesn't really gain us anything.

mousdahl-amd: Yeah, the allocators probably will do it too.

mousdahl-amd: I figured out how to tighten up the linter just for the classes.  I was putting it in the wrong place when I tried before.

sdk/include/hipdnn_sdk/utilities/tensor.hpp
```diff
@@ -54,19 +61,6 @@ class Tensor
...
-    {
-        return {dims, generate_strides(dims, layout.stride_order), sizeof(T)};
-    }
-
-    template <typename T>
-    static Tensor make_tensor(const std::vector<int64_t>& dims, const std::vector<int64_t>& strides)
-    {
-        return {dims, strides, sizeof(T)};
-    }
-
```
SamuelReeder, line 69: Need to update samples to use constructor.

sdk/include/hipdnn_sdk/utilities/tensor.hpp
```diff
@@ -34,18 +34,25 @@ inline std::ostream& operator<<(std::ostream& os, const Tensor_layout& layout)
 }
 
 // Wraps vectors of dims/strides and Migratable_memory<T> to provide a common interface for testing
+template <class T>
```
SamuelReeder: Should this have the allocators templated too? Because it hardcodes migratable memory to use `Host_allocator` and I don't think a user can use `Pinned_host_allocator` through tensor class.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
```diff
@@ -287,39 +286,52 @@ class Migratable_memory
...
     void* _device_ptr{nullptr};
     size_t _count;
     size_t _item_size;
     size_t _total_size;
-    Location _current_location{Location::NONE};
+    Memory_location _current_location{Memory_location::NONE};
     bool _host_valid{false};
     bool _device_valid{false};
+    HostAlloc _host_allocator;
+    DeviceAlloc _device_allocator;
```
SamuelReeder, line 368: These new members are missing from move & assignment overloads.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
```diff
@@ -251,32 +256,26 @@ class Migratable_memory
...
         if(!_host_valid && _device_valid && (_device_ptr != nullptr))
         {
-            throw_on_error(hipMemcpy(_host_ptr, _device_ptr, _total_size, hipMemcpyDeviceToHost),
-                           "Failed to copy from device to host");
+            if(async)
+            {
+                throw_on_error(
+                    hipMemcpyAsync(
+                        _host_ptr, _device_ptr, _total_size, hipMemcpyDeviceToHost, stream),
+                    "Failed to copy from device to host");
```
SamuelReeder: I'm a little worried about the user not knowing to synchronize. It would be generally fine but there's a lot of abstraction so I think we should incur the synchronize call.

mousdahl-amd: At the top layer there's explicit host_data_async() and device_data_async(), which I was hoping would be enough of a hint.

.clang-tidy
```diff
@@ -22,7 +22,7 @@ Checks: >
   -readability-avoid-unconditional-preprocessor-if,
   -readability-use-anyofallof,
   -readability-function-cognitive-complexity
-
+
```
BrianHarrisonAMD: nitpick: extra whitespace it looks like

plugins/miopen_legacy_plugin/integration_tests/batchnorm_fwd_inference_integration_test.cpp
```diff
@@ -47,38 +47,38 @@ struct Batchnorm_2d_tensor_bundle
...
-        , scale_tensor(Tensor::make_tensor<Intermediate_type>(derived_dims))
-        , bias_tensor(Tensor::make_tensor<Intermediate_type>(derived_dims))
-        , mean_tensor(Tensor::make_tensor<Intermediate_type>(derived_dims))
-        , variance_tensor(Tensor::make_tensor<Intermediate_type>(derived_dims))
+        , x_tensor(dims, layout)
+        , y_tensor(dims, layout)
+        , scale_tensor(derived_dims)
+        , bias_tensor(derived_dims)
+        , mean_tensor(derived_dims)
+        , variance_tensor(derived_dims)
```
BrianHarrisonAMD, line 55: What magic is this?

plugins/miopen_legacy_plugin/integration_tests/batchnorm_fwd_inference_integration_test.cpp
```diff
@@ -47,38 +47,38 @@ struct Batchnorm_2d_tensor_bundle
...
-    Tensor scale_tensor;
-    Tensor bias_tensor;
-    Tensor mean_tensor;
-    Tensor variance_tensor;
+    Tensor<Input_type> x_tensor;
+    Tensor<Input_type> y_tensor;
+    Tensor<Intermediate_type> scale_tensor;
+    Tensor<Intermediate_type> bias_tensor;
+    Tensor<Intermediate_type> mean_tensor;
+    Tensor<Intermediate_type> variance_tensor;
```
BrianHarrisonAMD: Ah

plugins/miopen_legacy_plugin/integration_tests/batchnorm_fwd_inference_integration_test.cpp
```diff
@@ -132,18 +132,14 @@ class Batchnorm_forward_inference_integration_test
...
+            = tensor_bundle.variance_tensor.memory().device_data();
         variant_pack[scale_tensor_attr.get_uid()]
-            = tensor_bundle.scale_tensor.memory().template device_data<void>();
-        variant_pack[bias_tensor_attr.get_uid()]
-            = tensor_bundle.bias_tensor.memory().template device_data<void>();
-        variant_pack[y_tensor_attr.get_uid()]
-            = tensor_bundle.y_tensor.memory().template device_data<void>();
+            = tensor_bundle.scale_tensor.memory().device_data();
+        variant_pack[bias_tensor_attr.get_uid()] = tensor_bundle.bias_tensor.memory().device_data();
+        variant_pack[y_tensor_attr.get_uid()] = tensor_bundle.y_tensor.memory().device_data();
```
BrianHarrisonAMD, line 142: Nice

samples/batchnorm/bn_backward.cpp
```diff
@@ -68,50 +68,46 @@ void Sample_runner::operator()(const Tensor_layout& layout)
...
+    scale_tensor.fill_with_random_values(static_cast<IntermediateType>(0.0f),
+                                         static_cast<IntermediateType>(1.0f));
+    saved_mean_tensor.fill_with_random_values(static_cast<IntermediateType>(0.0f),
+                                              static_cast<IntermediateType>(1.0f));
+    saved_inv_var_tensor.fill_with_random_values(static_cast<IntermediateType>(0.1f),
+                                                 static_cast<IntermediateType>(1.0f));
 
     std::unordered_map<int64_t, void*> variant_pack;
 
     // TODO: Cleanup syntax when there is a better way to grab these pointers.
```
BrianHarrisonAMD: Looks like this was todone?

samples/batchnorm/bn_inference.cpp
```diff
@@ -64,54 +64,50 @@ void Sample_runner::operator()(const Tensor_layout& layout)
...
-    mean_tensor.template fill_with_value<IntermediateType>(static_cast<IntermediateType>(0.5f));
+    mean_tensor.fill_with_value(static_cast<IntermediateType>(0.5f));
 
-    inv_variance_tensor.template fill_with_value<IntermediateType>(
-        static_cast<IntermediateType>(1.0f));
+    inv_variance_tensor.fill_with_value(static_cast<IntermediateType>(1.0f));
 
     std::unordered_map<int64_t, void*> variant_pack;
 
     // TODO: Cleanup syntax when there is a better way to grab these pointers.
```
BrianHarrisonAMD: Might be more of these that are todone I think.

samples/batchnorm/bn_training.cpp
```diff
@@ -73,55 +73,48 @@ void Sample_runner::operator()(const Tensor_layout& layout)
...
+                                             static_cast<IntermediateType>(1.0f));
+    prev_var_tensor.fill_with_random_values(static_cast<IntermediateType>(0.1f),
+                                            static_cast<IntermediateType>(1.0f));
+
+    momentum_tensor.memory().host_data()[0] = 0.1f;
+    epsilon_tensor.memory().host_data()[0] = 1e-5f;
 
     std::unordered_map<int64_t, void*> variant_pack;
 
     // TODO: Cleanup syntax when there is a better way to grab these pointers.
```
BrianHarrisonAMD: Todone!

sdk/include/hipdnn_sdk/utilities/allocators.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+        auto ptr = static_cast<T*>(std::malloc(n * sizeof(T)));
+        if(!ptr)
+        {
+            throw std::bad_alloc();
+        }
+        return ptr;
+    }
+
+    void deallocate(T* p, std::size_t /*n*/) noexcept override
```
BrianHarrisonAMD: I think we have been using std::ignore instead for these.

sdk/include/hipdnn_sdk/utilities/allocators.hpp
```diff
@@ -0,0 +1,295 @@
...
+bool operator==(const Host_allocator<T>& /*lhs*/, const Host_allocator<U>& /*rhs*/) noexcept
+{
+    return true;
+}
+
+template <typename T, typename U>
+bool operator!=(const Host_allocator<T>& /*lhs*/, const Host_allocator<U>& /*rhs*/) noexcept
+{
+    return false;
+}
```
BrianHarrisonAMD: What are these needed for?

mousdahl-amd: It's part of a stateless allocator pattern.  Apparently allocator-aware containers can make use of this during move operations / etc to know that one allocator can de-allocate memory from another allocator.  A stateful custom allocator would have a more specific implementation for these.

sdk/include/hipdnn_sdk/utilities/allocators.hpp
```diff
@@ -0,0 +1,311 @@
...
+    Host_allocator(const Host_allocator<U>& other) noexcept
+    {
+        std::ignore = other;
+    }
+
+    ~Host_allocator() override = default;
+
+    [[nodiscard]] T* allocate(std::size_t n) override
+    {
+        if(n > std::numeric_limits<std::size_t>::max() / sizeof(T))
```
BrianHarrisonAMD, line 132: Surprised this doesn't generate the () warning for math operations.
Is this file getting checked?

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
```diff
@@ -13,25 +14,34 @@ namespace hipdnn_sdk
...
+                  "DeviceAlloc must derive from Device_allocator_interface<T>");
 
-    explicit Migratable_memory(size_t count = 0, size_t item_size = 0)
+public:
+    explicit Migratable_memory(size_t count = 0)
         : _count(count)
-        , _item_size(item_size)
-        , _total_size(count * item_size)
+        , _item_size(sizeof(T))
+        , _total_size(count * _item_size)
```
BrianHarrisonAMD: What if we made stream as part of the constructor, and then removed it from the member functions?

I think in most cases we would want to use the same stream for all operations, and you can hit several issues if you mix streams between allocation / copy. I guess this is technically a bit more flexible, but I think we would rather make this simple to use, and less error prone. It should only be for tests & samples generally.

sdk/include/hipdnn_sdk/utilities/allocators.hpp
mousdahl-amd: It's getting checked.  This was generating one of the portability issues.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
mousdahl-amd: I think that makes sense.  Doing it.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
```diff
@@ -108,67 +122,68 @@ class Migratable_memory
...
     {
         const_cast<Migratable_memory*>(this)->ensure_device_valid(stream);
         return static_cast<T*>(_device_ptr);
     }
 
+    void* device_data_async(hipStream_t stream = nullptr) const
+    {
+        const_cast<Migratable_memory*>(this)->ensure_device_valid(stream, true);
+        return static_cast<T*>(_device_ptr);
+    }
```
BrianHarrisonAMD: Should this be const?
We either should make these not-const, or make the members mutable.

In this case I think it would make sense to treat these as non-const, (and remove the cast), since we are changing the host state, but if we wanted to treat it as a lazy populated cache, then I suppose mutable.

BrianHarrisonAMD: Or maybe the intent was a const accessor.

I think in that case we should just throw if it's not avaliable, and have a const accessor only no async version.
Basically a const version that will give it to you if it's not going to copy.

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
```diff
@@ -108,67 +122,68 @@ class Migratable_memory
...
 
-    // Get const host pointer (migrates if needed)
-    template <typename T>
-    const T* host_data() const
+    const T* host_data_async(hipStream_t stream = nullptr) const
     {
-        const_cast<Migratable_memory*>(this)->ensure_host_valid();
+        const_cast<Migratable_memory*>(this)->ensure_host_valid(stream, true);
         return static_cast<T*>(_host_ptr);
     }
```
BrianHarrisonAMD: Similar to the device, have a const version that will give it to you if it's not going to copy, or throws if it needs to modify state. 

Or we could do the mutable change, and make it always const, but that seems a bit off to me.

sdk/include/hipdnn_sdk/utilities/tensor.hpp
```diff
@@ -34,18 +34,25 @@ inline std::ostream& operator<<(std::ostream& os, const Tensor_layout& layout)
...
 // Wraps vectors of dims/strides and Migratable_memory<T> to provide a common interface for testing
+template <class T, class HostAlloc = Host_allocator<T>, class DeviceAlloc = Device_allocator<T>>
 class Tensor
 {
-private:
-    Tensor(const std::vector<int64_t>& dims, const std::vector<int64_t>& strides, size_t item_size)
-        : _memory(calculate_item_count(dims), item_size)
+public:
+    Tensor(const std::vector<int64_t>& dims, const std::vector<int64_t>& strides)
+        : _memory(calculate_item_count(dims))
```
BrianHarrisonAMD, line 66: Didn't realize this before, but technically this will only be correct if the strides are packed.
We either need to use the strides for calculating how much memory, or we should add some check to ensure packed strides.

BrianHarrisonAMD: Oh it was private before hehe

sdk/tests/test_utilities/test_cpu_fp_reference_validation.cpp
```diff
@@ -28,7 +28,6 @@ Migratable_memory create_buffer(size_t size, T mult)
 
 TEST(CpuFpReferenceValidation, BasicBFloat16Usage)
 {
-    SKIP_IF_NO_DEVICES();
```
BrianHarrisonAMD, line 31: 🎉

sdk/include/hipdnn_sdk/utilities/migratable_memory.hpp
mousdahl-amd: Looking at mutable, that looks gross.  I think I'll just get rid of the const.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
BrianHarrisonAMD, line 1: Should we swap these to pinned allocator for the performance 💪's?

plugins/miopen_legacy_plugin/integration_tests/batchnorm_fwd_inference_integration_test.cpp
BrianHarrisonAMD, line 1: Should we swap these to pinned allocator for the performance 💪's?

sdk/tests/utilities/test_migratable_memory.cpp
```diff
@@ -25,143 +25,278 @@ void check_buffer(const T* buffer, size_t size, T mult = 1)
...
+    error = hipStreamDestroy(stream);
+    EXPECT_EQ(error, hipSuccess) << "Failed to destroy HIP stream";
+}
+
+TEST(MigratableMemory, MigrateToDeviceAsyncNonDefaultStream)
+{
+    SKIP_IF_NO_DEVICES();
+
+    hipStream_t stream;
+    hipError_t error = hipStreamCreate(&stream);
```
BrianHarrisonAMD: I think we might need to do hipStreamCreateWithFlags, and provide the [non-blocking flag](https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/group___global_defs.html#gaaba9ae995d9b43b7d1ee70c6fa12c57d).

Or maybe a separate test, but I think they will still sync with default stream unless it's made this way.

BrianHarrisonAMD: More details to dig into [here](https://rocm.docs.amd.com/projects/HIP/en/docs-develop/how-to/hip_runtime_api/asynchronous.html).

sdk/tests/utilities/test_migratable_memory.cpp
BrianHarrisonAMD, line 1: If we don't make stream part of the constructor, then we probably want tests that have multiple streams, and we ensure the operations work when we swap which stream it was allocated on vrs used for copying etc.

mousdahl-amd: No, I'm 100% doing the constructor version for this reason.

BrianHarrisonAMD: Okay you can ignore this if we make it part of the constructor.

BrianHarrisonAMD:
Looking really nice!
Some small comments to address, but I think mostly good to go.

plugins/miopen_legacy_plugin/integration_tests/batchnorm_fwd_inference_integration_test.cpp
```diff
@@ -47,38 +47,38 @@ struct Batchnorm_2d_tensor_bundle
...
-    Tensor scale_tensor;
-    Tensor bias_tensor;
-    Tensor mean_tensor;
-    Tensor variance_tensor;
+    Tensor<Input_type, Pinned_host_allocator<Input_type>> x_tensor;
+    Tensor<Input_type, Pinned_host_allocator<Input_type>> y_tensor;
+    Tensor<Intermediate_type, Pinned_host_allocator<Intermediate_type>> scale_tensor;
+    Tensor<Intermediate_type, Pinned_host_allocator<Intermediate_type>> bias_tensor;
+    Tensor<Intermediate_type, Pinned_host_allocator<Intermediate_type>> mean_tensor;
+    Tensor<Intermediate_type, Pinned_host_allocator<Intermediate_type>> variance_tensor;
```
BrianHarrisonAMD: This is fine as is, but I think you could make a template alias for this, and then call it like PinnedTensor or something.

plugins/miopen_legacy_plugin/tests/test_miopen_batchnorm_operations.cpp
```diff
@@ -86,23 +86,23 @@ std::vector<Bn_2d_test_case> get_bn_fwd_inference_test_cases()
...
     hipdnnPluginDeviceBuffer_t buffer;
     buffer.uid = uid;
-    buffer.ptr = tensor.memory().device_data<T>();
+    buffer.ptr = tensor.memory().device_data();
     return buffer;
 }
 
 template <typename T>
-hipdnnPluginDeviceBuffer_t generate_static_device_buffer(Tensor& tensor, int uid, T value)
+hipdnnPluginDeviceBuffer_t generate_static_device_buffer(Tensor<T>& tensor, int uid, T value)
```
BrianHarrisonAMD: Surprised this one doesn't need to be Tensor_interface as well.

mousdahl-amd: Ah, I thought I hunted all these down!

plugins/miopen_legacy_plugin/integration_tests/batchnorm_fwd_inference_integration_test.cpp
```diff
@@ -47,38 +47,38 @@ struct Batchnorm_2d_tensor_bundle
...
-    Tensor scale_tensor;
-    Tensor bias_tensor;
-    Tensor mean_tensor;
-    Tensor variance_tensor;
+    PinnedTensor<Input_type> x_tensor;
+    PinnedTensor<Input_type> y_tensor;
+    PinnedTensor<Intermediate_type> scale_tensor;
+    PinnedTensor<Intermediate_type> bias_tensor;
+    PinnedTensor<Intermediate_type> mean_tensor;
+    PinnedTensor<Intermediate_type> variance_tensor;
```
BrianHarrisonAMD, line 81: Oh yea that looks nice!

samples/batchnorm/bn_inference.cpp
```diff
@@ -64,54 +64,49 @@ void Sample_runner::operator()(const Tensor_layout& layout)
...
-    auto bias_tensor = Tensor::make_tensor<IntermediateType>(bias->get_dim());
-    auto mean_tensor = Tensor::make_tensor<IntermediateType>(mean->get_dim());
-    auto inv_variance_tensor = Tensor::make_tensor<IntermediateType>(inv_variance->get_dim());
-    auto y_tensor = Tensor::make_tensor<InputType>(y->get_dim(), layout);
+    Tensor<InputType> x_tensor(x->get_dim(), layout);
+    Tensor<IntermediateType> scale_tensor(scale->get_dim());
+    Tensor<IntermediateType> bias_tensor(bias->get_dim());
+    Tensor<IntermediateType> mean_tensor(mean->get_dim());
+    Tensor<IntermediateType> inv_variance_tensor(inv_variance->get_dim());
+    Tensor<InputType> y_tensor(y->get_dim(), layout);
```
BrianHarrisonAMD, line 72: Probably fine as is, but maybe Pinned here make sense, (and same for all samples).

BrianHarrisonAMD (approved):
LGTM!

PR #132, opened 2025-08-25
Add testing for unhappy path & integration

jerehartAMD:
## Proposed changes
Added new integration tests
- No plugin can support the graph.
- Include tests for providing incorrect engineIds (non-existent plugin Id)
- Empty plugin directory (no plugins)
- Multiple plugins exist.
- Multiple engines exist.
- Some engines can handle it, and some engines cannot.
- Incomplete plugin API (only partially handling API).
- EngineId conflicts between two different plugins

Tests that should still be added:
- EngineId conflicts within the same plugin
- Plugin name and version conflicts

Did some additional refactoring as well:
- Create Temp_directory class that uses RAII to create temporary directories
- Added the `test_plugin_engine_id_map.hpp` file, which maps the test plugin classes to ids (This might have been overengineered, it might make more sense to just have a named, untemplated function for each)

## Checklist


tests/backend/integration_test_unhappy_plugin_path.cpp
```diff
@@ -0,0 +1,268 @@
...
+    const std::array<const char*, 2> paths
+        = {hipdnn_tests::plugin_constants::test_duplicate_id_a_plugin_path().c_str(),
+           hipdnn_tests::plugin_constants::test_duplicate_id_b_plugin_path().c_str()};
+    ASSERT_EQ(
+        hipdnnSetEnginePluginPaths_ext(paths.size(), paths.data(), HIPDNN_PLUGIN_LOADING_ABSOLUTE),
+        HIPDNN_STATUS_SUCCESS);
+
+    ASSERT_EQ(hipdnnCreate(&_handle), HIPDNN_STATUS_SUCCESS);
+
+    // TODO: Warning is logged, but we don't have means of querying the last warning
```
jerehartAMD: This comment may not be true? I couldn't find a way to test that this is being logged, but I was able to verify it was myself

BrianHarrisonAMD: There is logging tests that can check for logs outputs if needed. 
@SamuelReeder or @adickin-amd should be able to point you at those for reference.

backend/tests/test_plugin.cpp
```diff
@@ -123,10 +130,39 @@ TEST(PluginManagerTest, LoadPluginsFromDirectory)
...
+        Test_plugin_manager plugin_manager;
+        plugin_manager.load_plugins({temp_dir.path()}, HIPDNN_PLUGIN_LOADING_ABSOLUTE);
+
+        ASSERT_TRUE(plugin_manager.get_loaded_plugin_files().empty());
+        ASSERT_TRUE(plugin_manager.get_plugins().empty());
+    }
+    catch(...)
+    {
+        FAIL();
+    }
```
BrianHarrisonAMD: Do we need the try/catch around code we aren't expecting to throw?

I thought if an unexpected exception occured, it would get handled automatically and fail the test inside gtest, but it would be good to get that confirmed.

backend/src/plugin/plugin_core.hpp
```diff
@@ -174,7 +174,12 @@ class Plugin_manager_base
                 {
                     paths_to_load.insert(resolved_path);
                 }
-                // Consider logging `else` here once cognitive complexity is resolved
+                else
+                {
+                    HIPDNN_LOG_WARN(
+                        "Plugin path is not a directory and doesn't have a filepath component '{}'",
```
bghimireamd:
```suggestion
                        "Plugin path '{}' is invalid - expected either a directory containing plugins or a path to a plugin file"

File: tests/backend/integration_test_unhappy_plugin_path.cpp [hunk 4]
```diff
@@ -0,0 +1,268 @@
...
+#include "hipdnn_backend.h"
+#include "hipdnn_sdk/plugin/engine_plugin_api.h"
+#include "hipdnn_sdk/plugin/plugin_api.h"
+#include "hipdnn_sdk/utilities/platform_utils.hpp"
+#include "test_util.hpp"
+#include <hipdnn_backend_attribute_name.h>
+#include <hipdnn_backend_attribute_type.h>
+#include <hipdnn_backend_heuristic_type.h>
+#include <hipdnn_sdk/test_utilities/temporary_directory.hpp>
+#include <iostream> // Don't forget to remove
```
BrianHarrisonAMD: Was this forgot? 
:P

tests/backend/integration_test_unhappy_plugin_path.cpp
```diff
@@ -0,0 +1,268 @@
...
+                                        HIPDNN_TYPE_BACKEND_DESCRIPTOR,
+                                        0,
+                                        &available_engine_count,
+                                        nullptr),
+              HIPDNN_STATUS_SUCCESS);
+
+    EXPECT_EQ(available_engine_count, 1);
+}
+
+TEST_F(Unhappy_plugin_path_tests, MultiplePluginsMultipleApplicableEngines)
```
BrianHarrisonAMD: This test isn't really unahappy path.
Might be worth renaming these tests to something like "integration_test_plugin_loading", or maybe move this to the happy version of this API?

BrianHarrisonAMD (approved):
Looking really good to me!

Couple small comments, but generally good with it as is.

tests/backend/integration_test_unhappy_plugin_path.cpp
jerehartAMD: Indeed it was haha

backend/tests/test_plugin.cpp
jerehartAMD: There was a test above it that had a similar pattern. However, now that I think of it that may have actually been because it needed to catch that so it could remove the temporary file. I don't actually know how the uncaught exception is handled here and if the destructor gets called, but I'll take a look at it! We might be able to kill a couple of these try catch blocks now

jerehartAMD: I removed this thinking that it was a duplicate of the integration test version, I'll double check to see if there's something meaningfully different between the two and if there is I might bring it back

sdk/include/hipdnn_sdk/test_utilities/temporary_directory.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include <filesystem>
+
+class Temp_dir
```
adickin-amd: since were renaming soon we can ignore but normally we want the class name to match the file name.

sdk/include/hipdnn_sdk/test_utilities/temporary_directory.hpp
```diff
@@ -0,0 +1,37 @@
...
+public:
+    Temp_dir(std::filesystem::path path)
+    {
+        if(std::filesystem::create_directory(path))
+        {
+            _path = std::move(path);
+        }
+        else
+        {
+            std::runtime_error("Temp_dir: Directory already exists");
```
adickin-amd: do we have to add `throw` in front of this or does it throw upon creation?

sdk/include/hipdnn_sdk/test_utilities/temporary_directory.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include <filesystem>
+
+class Temp_dir
+{
+    std::filesystem::path _path;
+
+public:
+    Temp_dir(std::filesystem::path path)
```
adickin-amd: Do we have a unit test written for this?

tests/backend/integration_test_unhappy_plugin_path.cpp
adickin-amd: What ends up in the lastErrorString in this case?  Im curious as to why we cant query in this case?

tests/backend/integration_test_unhappy_plugin_path.cpp
adickin-amd: what about multiple plugins, no applicable engines.

sdk/include/hipdnn_sdk/test_utilities/temporary_directory.hpp
jerehartAMD: Oops, we definitely do haha, I missed that

tests/test_plugins/test_no_applicable_engines_plugin.cpp
```diff
@@ -21,7 +21,7 @@ class No_applicable_engines_plugin : public Test_plugin_base
     }
     int64_t get_engine_id() const override
     {
-        return -3;
+        return hipdnn_tests::plugin_constants::engine_id<No_applicable_engines_plugin>();
```
adickin-amd, line 24: love the clean up of all the random magic numbers.

tests/backend/integration_test_unhappy_plugin_path.cpp
jerehartAMD: Haven't tested that one, but I'll add that to the list!

sdk/include/hipdnn_sdk/test_utilities/temporary_directory.hpp
jerehartAMD: Not yet, but we should, I'll add it in a PR after we make the renaming branch

backend/tests/test_plugin.cpp
jerehartAMD: We don't need it, since gtest catches the throw the destructor will be called, I've removed all of the try catch blocks with that pattern

tests/backend/integration_test_unhappy_plugin_path.cpp
jerehartAMD: I'll be turning the plugin load failure logs into an error in a PR after the renaming split off

PR #137, opened 2025-08-25
MIOpen batchnorm backwards integration

BrianHarrisonAMD:
## Proposed changes

Add integration to the MIOpen plugin for batchnorm bwds.

## Checklist


sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_miopen_rms_validation.hpp
```diff
@@ -0,0 +1,92 @@
...
+            // Track maximum magnitudes
+            max_ref_magnitude = std::max(max_ref_magnitude, std::fabs(ref_value));
+            max_impl_magnitude = std::max(max_impl_magnitude, std::fabs(impl_value));
+        }
+
+        // Find the maximum magnitude between reference and implementation
+        double max_magnitude = std::max({max_ref_magnitude, max_impl_magnitude, std::numeric_limits<double>::min()});
+
+        // Calculate and compare to _relative_tolerance
+        return (std::sqrt(square_difference) / (std::sqrt(static_cast<double>(element_count)) * max_magnitude)) < static_cast<double>(_relative_tolerance);
```
mousdahl-amd: You might get nailed for implicit cast when used with half / bfloat16 since it doesn't have a way to coerce to double otherwise.

BrianHarrisonAMD: Oh let me check.

BrianHarrisonAMD: I think it was working since we ran it before.

sdk/include/hipdnn_sdk/test_utilities/cpu_fp_reference_miopen_rms_validation.hpp
BrianHarrisonAMD, line 1: Add tests for this.

mousdahl-amd (approved):
Looks good to me.  Add a ticket for RMS tests later.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_bwd_plan.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+    const Miopen_tensor& x() const;
+    const Miopen_tensor& dy() const;
+    const Miopen_tensor& dx() const;
+    const Miopen_tensor& scale() const;
+    const Miopen_tensor& dscale() const;
+    const Miopen_tensor& dbias() const;
+
+    const std::optional<std::unique_ptr<Miopen_tensor>>& opt_mean() const;
+    const std::optional<std::unique_ptr<Miopen_tensor>>& opt_inv_variance() const;
```
jerehartAMD, line 34: I don't love that this has two ways of being null, and requires two "dereferences". I'm not sure of what you'd do to get around it though. I wish std::optional was like tl::optional and could handle references :(

plugins/miopen_legacy_plugin/integration_tests/batchnorm_bwd_integration_test.cpp
```diff
@@ -0,0 +1,398 @@
...
+{
+    Bn_2d_test_case test_case = GetParam();
+    run_batchnorm_test<hip_bfloat16, float>(test_case, 4e-3_bf);
+}
+
+INSTANTIATE_TEST_SUITE_P(RunBfloat16BwdBatchnormGraph,
+                         Batchnorm_backward_integration_test_bfloat16,
+                         testing::ValuesIn(get_bn_bwd_test_cases()));
+
+TEST_P(Batchnorm_backward_integration_test_half, RunHalfBwdBatchnormGraph)
```
bghimireamd:
```suggestion
TEST_P(Batchnorm_backward_integration_test_half, RunHalfBwdBatchnormGraphNCHW)

**Review by bghimireamd — APPROVED** [review]
LGTM

File: plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_bwd_plan.cpp [hunk 4]
```diff
@@ -0,0 +1,178 @@
...
+namespace miopen_legacy_plugin
+{
+
+// We have made the intentional decision to hardcode the batchnorm mode to miopenBNSpatial
+// rather than making it configurable and adding extra complexity.
+const miopenBatchNormMode_t MIOPEN_BATCHNORM_MODE = miopenBNSpatial;
+
+Batchnorm_bwd_params::Batchnorm_bwd_params(
+    const hipdnn_sdk::data_objects::BatchnormBackwardAttributes& attributes,
+    const std::unordered_map<int64_t, const hipdnn_sdk::data_objects::TensorAttributes*>&
```
jerehartAMD, line 16: Is there a reason this needs to be a map? If there was a way to guarantee they're present that'd make this a lot easier.

It seems that the reason why all the class members are unique_ptrs is because we can have missing fields in this map, but also the object will throw if any of them are missing, so it _can't_ be missing any fields. I wonder if there's a way that we could restructure this to avoid the unique_ptrs entirely

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_plan_builder.cpp
```diff
@@ -43,43 +46,83 @@ size_t Miopen_batchnorm_plan_builder::get_workspace_size(
...
+            HIPDNN_PLUGIN_STATUS_BAD_PARAM,
+            "Failed to convert node attributes to BatchnormInferenceAttributes for node: "
+                + get_node_name(node));
+    }
+
+    auto params
+        = std::make_unique<Batchnorm_fwd_inference_params>(*attr, op_graph.get_tensor_map());
+    auto plan = std::make_unique<Batchnorm_fwd_inference_plan>(std::move(params));
+    execution_context.set_plan(std::move(plan));
+}
```
BrianHarrisonAMD, line 77: Discussed today, that we will add a task to make wrapper objects for anything that needs to use flatbuffer objects without deserializing in the SDK. That way we don't have to put the burden on plugin creators to know all the gotchas, but also don't need to pay the deserialize cost to ensure it works properly.

SamuelReeder: It's added at bottom of backlog.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_bwd_plan.cpp
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    throw hipdnn_plugin::Hipdnn_plugin_exception(HIPDNN_PLUGIN_STATUS_INTERNAL_ERROR,
+                                                 "Failed to find tensor with UID in tensor_map: "
+                                                     + std::to_string(uid));
+}
+
+} // namespace
+
+void Batchnorm_bwd_params::initialize_tensors(
```
jerehartAMD, line 82: Okay! I think if we just called the create_tensor functions in the constructors of each member variable instead of here I think we could entirely avoid the unique_ptrs even while keeping the map. That would also handle the double indirection on the optionals

jerehartAMD (approved):
LGTM! If the suggestion in my comments could work I'd love to see that change made, but it has nothing to do with correctness and isn't blocking

plugins/miopen_legacy_plugin/integration_tests/batchnorm_bwd_integration_test.cpp
```diff
@@ -0,0 +1,398 @@
...
+        ASSERT_EQ(hipdnnSetEnginePluginPaths_ext(
+                      paths.size(), paths.data(), HIPDNN_PLUGIN_LOADING_ABSOLUTE),
+                  HIPDNN_STATUS_SUCCESS);
+
+        // Create handle
+        ASSERT_EQ(hipdnnCreate(&_handle), HIPDNN_STATUS_SUCCESS);
+
+        //todo: bring back stream support once MigratableMemory supports it
+        //ASSERT_EQ(hipStreamCreate(&_stream), hipSuccess);
+        //ASSERT_EQ(hipdnnSetStream(_handle, _stream), HIPDNN_STATUS_SUCCESS);
```
BrianHarrisonAMD, line 120: Maybe we can leave these for now and clean it up in a big sweep since I think this was done.

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_bwd_plan.cpp
adickin-amd: its a map because of how flatbuffers packs the data. We generate a generic map here https://github.com/ROCm/hipDNN/blob/develop/sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/graph_wrapper.hpp#L92-L107

plugins/miopen_legacy_plugin/engines/plans/miopen_batchnorm_bwd_plan.cpp
adickin-amd: Are you saying that instead of std::unique_ptr<Miopen_tensor> we would just have MIOpen_tensor on the class?

BrianHarrisonAMD: Yea seems like we can move this.

How does that fix the optional though?

jerehartAMD: We can store the value as an `optional<Miopen_tensor>`, so our accessors give us references or a reference to an optional. Before we couldn't give an `optional<MIOpen_tensor>&` because there was a unique_ptr in the middle

BrianHarrisonAMD: Oh I see.

BrianHarrisonAMD:
Merged to not block renaming efforts.
Will have a follow-up with suggested feedback.

PR #138, opened 2025-08-25
Backend Rename Leftovers

mousdahl-amd:
## Proposed changes
Renamed the leftover backend pieces

## Checklist


backend/src/logging/logging.cpp
```diff
@@ -25,8 +25,8 @@ namespace
 
 // Could refactor this to a class with a single static instance.
 // The benefit would be a destructor to cleanup logging.
-std::mutex s_logging_init_mutex;
-bool s_logging_initialized = false;
+std::mutex s_loggingInitMutex;
+bool s_loggingInitialized = false;
```
jerehartAMD, line 29: What's the `s_` for?

jerehartAMD (approved):
As long as the `s_` is intended, LGTM!

backend/src/hipdnn_exception.hpp
```diff
@@ -39,40 +39,40 @@ class Hipdnn_exception : public std::exception
     std::string _message;
 };
 
-#define THROW_IF_NE(x, y, failure_status, message)                       \
-    if(x != y)                                                           \
-    {                                                                    \
-        throw hipdnn_backend::Hipdnn_exception(failure_status, message); \
+#define THROW_IF_NE(x, y, failure_status, message)                      \
```
adickin-amd: missed failure_status in macros

adickin-amd (approved):
just 1 fix up in the macros

backend/src/logging/logging.cpp
mousdahl-amd: Some people use it for globals.  I've seen `s_` and `g_`  Maybe that's another thing for the rename docs.  Thoughts @adickin-amd ?

mousdahl-amd: Leaving it for now.

adickin-amd (approved):
approved!

backend/src/logging/logging.cpp
```diff
@@ -25,8 +25,8 @@ namespace
...
 const std::string S_BACKEND_LOGGER_NAME = "hipdnn_backend";
 const std::string S_CALLBACK_RECEIVER_LOGGER_NAME = "hipdnn_callback_receiver";
```
BrianHarrisonAMD, line 31: These probably should be without S_.

PR #162, opened 2025-08-27
Rename sdk tests utilities

BrianHarrisonAMD:
Rename sdk tests utilities to match new naming scheme

### DIFF
--- a/sdk/tests/utilities/test_callback_logger.cpp
+++ b/sdk/tests/utilities/test_callback_logger.cpp
@@ -12,36 +12,36 @@
 #include <hipdnn_sdk/logging/callback_types.h>
 #include <hipdnn_sdk/logging/logger.hpp>
 
-static std::vector<std::string> g_captured_logs;
-static std::mutex g_log_mutex;
+static std::vector<std::string> g_CapturedLogs;
+static std::mutex g_LogMutex;
 
 // Custom callback for testing. It doesn't fully simulate the real logging behavior,
 // but the backend tests use the true callback function. The test could use the backend callback, but then it has to link against the backend.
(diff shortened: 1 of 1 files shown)


sdk/tests/utilities/test_callback_logger.cpp
```diff
@@ -12,131 +12,131 @@
 #include <hipdnn_sdk/logging/callback_types.h>
 #include <hipdnn_sdk/logging/logger.hpp>
 
-static std::vector<std::string> g_captured_logs;
-static std::mutex g_log_mutex;
+static std::vector<std::string> s_capturedLogs;
+static std::mutex s_logMutex;
```
SamuelReeder, line 16: So global would be g_LOG_MUTEX and static is like above?

BrianHarrisonAMD: Only if it's const.

BrianHarrisonAMD: I think
- g_logMutex for non-const.
- g_LOG_MUTEX for const (but also probably try to avoid)

SamuelReeder: Right, realized I need to change my global here then: https://github.com/ROCm/hipDNN/blob/mo/rename_integration/backend/tests/test_plugin.cpp#L56

SamuelReeder (approved):

sdk/tests/utilities/test_callback_logger.cpp
BrianHarrisonAMD: Local, or statics I think are s_

SamuelReeder: Could avoid the global, but outside the scope.

PR #163, opened 2025-08-27
Replace global with test fixture

SamuelReeder:
## Proposed changes

I named a global incorrectly here but opted to make a minimal change to avoid using the global altogether. 

## Checklist

### DIFF
--- a/backend/tests/test_plugin.cpp
+++ b/backend/tests/test_plugin.cpp
@@ -53,13 +53,25 @@ public:
     }
 };
 
-bool g_CALLBACK_CALLED = false;
-void dummyCallback(hipdnnSeverity_t sev, const char* msg)
+class PluginCallbackTest : public ::testing::Test
 {
-    (void)sev;
-    (void)msg;
-    g_CALLBACK_CALLED = true;
-}
+protected:
+    void SetUp() override
+    {
+        s_callbackCalled = false;
+    }
+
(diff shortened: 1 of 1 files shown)


mousdahl-amd (approved):

backend/tests/test_plugin.cpp
```diff
@@ -53,13 +53,25 @@ class TestPluginManager : public plugin::PluginManagerBase<Plugin>
...
 };
 
-bool g_CALLBACK_CALLED = false;
-void dummyCallback(hipdnnSeverity_t sev, const char* msg)
+class PluginCallbackTest : public ::testing::Test
 {
-    (void)sev;
-    (void)msg;
-    g_CALLBACK_CALLED = true;
-}
```
BrianHarrisonAMD, line 62: I think it should have been s_callbackCalled since it was in the unnamed namespace, and it was local.
This also works though.

BrianHarrisonAMD (approved):

backend/tests/test_plugin.cpp
BrianHarrisonAMD: g_ would be if it was declared in a header, and accessible everywhere I believe.

SamuelReeder: Ah, I wasn't sure if it would be considered a global with internal linkage or a static.

PR #169, opened 2025-08-28
Rename tests and files in test/backend + refactoring

jerehartAMD:
## Proposed changes

Renaming tests and files to conform with our new testing standards. I attempted to reduce redundancy between the test suite and test names, as well as making the language for the test names a little more consistent.

## Checklist


tests/backend/TestCreateAndDeserializeGraphExt.cpp
```diff
@@ -0,0 +1,15 @@
+#include "hipdnn_backend.h"
+#include <gtest/gtest.h>
+#include <hipdnn_sdk/data_objects/graph_generated.h>
+#include <hipdnn_sdk/utilities/PlatformUtils.hpp>
+
+// TODO: Move this into backend/tests
```
jerehartAMD: I think the TODO I added here is incorrect as this is testing just the API and public interface

tests/backend/IntegrationBackendDescriptor.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+#include "hipdnn_backend.h"
+#include <gtest/gtest.h>
+#include <hipdnn_sdk/data_objects/graph_generated.h>
+#include <hipdnn_sdk/utilities/PlatformUtils.hpp>
+#include <vector>
+
+namespace fs = std::filesystem;
+
+TEST(IntegrationBackendDescriptor, CreateAndDestroy)
```
jerehartAMD, line 12: Tests moved from the TestBackendPulbicFunctions.cpp file to here

tests/backend/TestCreateAndDeserializeGraphExt.cpp
jerehartAMD: This test moved from the TestBackendPublicFunctions.cpp file to here

tests/backend/IntegrationBackendDescriptor.cpp
```diff
@@ -0,0 +1,142 @@
...
+TEST(IntegrationBackendDescriptor, Finalize)
+{
+    hipdnnBackendDescriptor_t descriptor = nullptr;
+
+    hipdnnStatus_t status = hipdnnBackendFinalize(descriptor);
+
+    EXPECT_EQ(status, HIPDNN_STATUS_BAD_PARAM_NULL_POINTER);
+}
+
+TEST(IntegrationBackendDescriptor, GetAttribute)
```
adickin-amd: not sure how picky we want to be but this test case name should probably be more like `GetAttributeFailsWithNullDescriptor` or something.

tests/backend/IntegrationBackendDescriptor.cpp
```diff
@@ -0,0 +1,142 @@
...
+                                                      attributeName,
+                                                      attributeType,
+                                                      requestedElementCount,
+                                                      &elementCount,
+                                                      arrayOfElements);
+
+    EXPECT_EQ(status, HIPDNN_STATUS_BAD_PARAM_NULL_POINTER);
+}
+
+TEST(IntegrationBackendDescriptor, SetAttribute)
```
adickin-amd: similar comment to above

tests/backend/IntegrationBackendExecuteApi.cpp
```diff
@@ -106,7 +106,28 @@ TEST_F(ExecutionBackendEndApiTests, TestBackendExecuteWithNullDescriptors)
...
+    test_util::createTestEngineConfig(
+        &_engineConfig, &_engine, &_graphDescriptor, _handle, GIDX, true);
+
+    EXPECT_EQ(hipdnnBackendCreateDescriptor(HIPDNN_BACKEND_EXECUTION_PLAN_DESCRIPTOR, &_plan),
+              HIPDNN_STATUS_SUCCESS);
+
+    test_util::populateTestExecutionPlan(
+        &_plan, &_engineConfig, &_engine, &_graphDescriptor, _handle, GIDX, true);
+
+    ASSERT_EQ(hipdnnBackendExecute(_handle, nullptr, _variantPack),
```
adickin-amd, line 114: is the _variantPack created here?  I think its null or is it setup in a helper somewhere?

tests/backend/TestCreateAndDeserializeGraphExt.cpp
adickin-amd: I think id put this test with the descriptor tests.  At the end of the day its a shortcut to making a graph descriptor directly from a flatbuffer graph.

adickin-amd: There is even a test in that integration test, `TEST(IntegrationBackendDescriptor, SetOperationGraph)` which uses this same API

adickin-amd: could also decide on moving all the graph extension api tests here i suppose.  Whatever you feel is cleaner Id say.

tests/backend/IntegrationBackendDescriptor.cpp
jerehartAMD: Agree on both this and the below one, I don't think I looked closely at the test itself when i was renaming this. For consistency I've been writing the scenario the test is testing and omitting the outcome, so I'd be renaming these as: `Get/SetAttributeWithNullDescriptor`.

Do we have a preference on whether or not we want to include outcomes as well?

tests/backend/TestCreateAndDeserializeGraphExt.cpp
jerehartAMD: Sounds good to me, I'm happy to move it since it's already being tested there too

tests/backend/IntegrationBackendExecuteApi.cpp
jerehartAMD: It looks like it is null, and I think several of the other tests here are working with a null _variantPack

tests/backend/IntegrationBackendDescriptor.cpp
adickin-amd: I leave that up to the developer.  I think we wanted to leave lots of flexibility in test case naming.

tests/backend/IntegrationBackendExecuteApi.cpp
adickin-amd: maybe thats okay then.  im not 100% sure tbh

jerehartAMD: I'll take a quick look into them, I should clarify when I say "working with" I mean using, they're all tests designed to fail, but they're supposed to fail in other ways

jerehartAMD: I made a fix that ensures that they're not null in all the tests checking for other things, but I don't know if it's sufficient because the variant pack is also not finalized or filled with any values. I'm considering making a task for this

adickin-amd (approved):
Approved as long as we decide the api tests are indeed integration tests

BrianHarrisonAMD (approved):

PR #175, opened 2025-08-29
Rename frontend tests

SamuelReeder:
## Proposed changes

Rename frontend tests and a couple conv files for consistency.

## Checklist


frontend/include/hipdnn_frontend/node/ConvolutionNode.hpp
SamuelReeder, line 1: The conv fwd naming is very inconsistent. It's called `ConvFprop`, `ConvolutionFwd`, sometimes just `Convolution`, etc. We could make it more consistent, however, some places would not match as well and we might have to regenerate schema files. For now, I opted to only rename this file and its tests because it's minimal and the class is named `ConvolutionNode`.

frontend/tests/TestConvolutionNode.cpp
```diff
@@ -482,7 +482,7 @@ TEST(ConvolutionFwdNodeTests, StrideInferenceDimensionMismatch)
     EXPECT_EQ(error.code, error_code_t::ATTRIBUTE_NOT_SET);
 }
 
-TEST(ConvolutionFwdNodeTests, StrideInferenceNCHWLayoutSuccess)
```
SamuelReeder, line 485: Need to change NCHW casing.

frontend/tests/TestConvolutionNode.cpp
```diff
@@ -520,7 +520,7 @@ TEST(ConvolutionFwdNodeTests, StrideInferenceNCHWLayoutSuccess)
     EXPECT_EQ(inferredStrides[3], 1); // W stride should be 1 (contiguous)
 }
 
-TEST(ConvolutionFwdNodeTests, StrideInferenceNHWCLayoutSuccess)
+TEST(TestConvolutionNode, StrideInferenceNHWCLayoutSuccess)
```
SamuelReeder: Here too.

There's a few more I'll change as well.

frontend/include/hipdnn_frontend/node/ConvolutionNode.hpp
BrianHarrisonAMD: Maybe we can name it ConvolutionFwdNode?
We do want to make it clear it's FWD.

BrianHarrisonAMD: I guess leave it for now.
We are going to do the naming pass on the frontend anyways.

BrianHarrisonAMD (approved):
LGTM

PR #177, opened 2025-09-02
Namespace consolidation

mousdahl-amd:
## Proposed changes
Renamed and consolidated some namespaces.  Generally, the namespace layout I aimed for was:

- hipdnn_<component> - root namespace for a component (eg. hipdnn_sdk), contains primary interface for a component
    - utilities - namespace containing utilities that make the component easier to use
    - test_utilities - namespace containing utilities that aid in testing

Some other namespaces still exist, such as platform_utilities (which I wanted to keep separate as an indicator that a dev should consider multi-platform choices when working here)

_Note, ASAN is currently failing due to an unrelated issue, I've added a bug to the backlog to fix it._

## Checklist

- [x] ⚠️ I have ran the tests with ASAN, and they are all passing locally


backend/src/PlatformUtils.linux.cpp
```diff
@@ -8,7 +8,7 @@
 #include "HipdnnException.hpp"
 #include <dlfcn.h>
 
-namespace hipdnn_backend::platform_utils
+namespace hipdnn_backend::platform_utilities
```
BrianHarrisonAMD, line 11: platform_utilities seems a bit odd to me with the others, but I can't think of a better name.
I am fine with it as is, because I can't think of a better name hehe.

mousdahl-amd: Yeah, I can't think of a better one either.  I even asked Sline, and it said it was pretty good.

sdk/include/hipdnn_sdk/logging/CallbackSink.hpp
```diff
@@ -92,9 +92,9 @@ template <typename Factory = spdlog::synchronous_factory>
 inline std::shared_ptr<spdlog::logger> createCallbackLoggerMt(hipdnnCallback_t callback,
                                                               const std::string& source)
 {
-    auto logger = Factory::template create<hipdnn::logging::CallbackSinkMt>(source, callback);
+    auto logger = Factory::template create<hipdnn_sdk::logging::CallbackSinkMt>(source, callback);
```
jerehartAMD, line 95: Not that this is a change from prior behavior, but do we prefer to explicitly include the namespace even when you're in it? There's a number of examples of this. I generally prefer not to for conciseness, but I could imagine there being some benefits to being explicit

edit: Just checked here specifically, it's not necessary for compilation

mousdahl-amd: I think I tend to agree with you.  I'll take a swing at it.

BrianHarrisonAMD (approved):
Sorry got distracted mid-way review.
LGTM.

sdk/include/hipdnn_sdk/logging/CallbackSink.hpp
mousdahl-amd: I took a look, and it touches too many lines of code, and this PR is big enough already.  I'll add a follow-on ticket.

PR #178, opened 2025-09-02
Rename miopen_legacy_plugin/tests

jerehartAMD:
## Proposed changes

Updated tests to conform with new naming rules

## Checklist


plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
```diff
@@ -225,6 +225,6 @@ void BatchnormBwdExecuteGraphTest::runBwdBatchnormGraph(
         cpuRefValidationIntermediate.allClose(dbiasTensorCpu.memory(), dbiasTensor.memory()));
 }
 
-INSTANTIATE_TEST_SUITE_P(RunBwdBatchnormGraphWithParams,
-                         BatchnormBwdExecuteGraphTest,
+INSTANTIATE_TEST_SUITE_P(,
```
jerehartAMD, line 333: Left this blank here because there's only one set of parameters we're using to instantiate this with and they seem pretty general. This will omit the first argument from the test name

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormFwdInferenceOperations.cpp
```diff
@@ -48,14 +48,14 @@ class BatchnormFwdInferExecuteGraphTest : public ::testing::TestWithParam<Batchn
...
-TEST_P(BatchnormFwdInferExecuteGraphTest, RunFloatFwdBatchnormGraphNCHW)
+TEST_P(TestGpuBatchnormFwdInferExecuteGraph, Fp32Nchw)
 {
     auto testCase = GetParam();
     runFwdBatchnormGraph<float, float>(
         testCase, hipdnn_sdk::data_objects::DataType::DataType_FLOAT, 1e-6f, TensorLayout::NCHW);
 }
 
-TEST_P(BatchnormFwdInferExecuteGraphTest, RunBfloat16FwdBatchnormGraphNCHW)
+TEST_P(TestGpuBatchnormFwdInferExecuteGraph, Bfp16Nchw)
```
BrianHarrisonAMD: I think the data type should be part of the TestSuite name, and might be worth adding the layout as well.

`TestGpuBatchnormFwdInferExecuteGraph` -> `TestGpuBatchnormFwdInferExecuteGraphBfp16`

or

`TestGpuBatchnormFwdInferExecuteGraph` -> `TestGpuBatchnormFwdInferExecuteGraphNchwBfp16`

BrianHarrisonAMD: For test name not really sure.
I was putting generic Correctness for the integration versions of these. 
These are like integration light tests.

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
```diff
@@ -48,14 +48,14 @@ class BatchnormBwdExecuteGraphTest : public ::testing::TestWithParam<Batchnorm2d
     hipdnnEnginePluginHandle_t _handle = nullptr;
 };
 
-TEST_P(BatchnormBwdExecuteGraphTest, RunFloatBwdBatchnormGraphNCHW)
+TEST_P(TestGpuMiopenBatchnormBwdExecuteGraph, Fp32Nchw)
```
SamuelReeder: All these need to have datatype at end of test suite.

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormFwdInferenceOperations.cpp
jerehartAMD: Responding to both yours and Samuel's comments here: Does this mean that as a rule a test suite can only ever test one data type and configuration? That feels a little constrained to me, and having a bunch of single test test suites instead of a single test suite that shares setup and varies on the types feels like we're changing the testing code structure for the naming instead of using the naming to represent the tests

jerehartAMD: That being said, maybe this should be refactored on a larger scale to mimic the full Integration tests. 

Edit: Looking into those changes specifically, I see how you changed the integration tests and the inheritance method is a pretty small scale change on the code

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
jerehartAMD: Replied to yours and Brian's comments in his comment chain

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormFwdInferenceOperations.cpp
BrianHarrisonAMD: I think the way we have it defined now, that is the case. 
In other files we have a shared base, and then extend each for specific configurations.
I don't think its too restrictive, but we could discuss and revisit this.

BrianHarrisonAMD: In this case, we have a templated method we are calling.
We can instead make the testing class templated, and have instances for each configuration.

That's similar to how it was done [here](https://github.com/ROCm/hipDNN/pull/180).

SamuelReeder: Is there a realistic case where we would need more than one datatype in the test name? Even mixed precision is assumed to be float32 accumulation so that is probably fine.

BrianHarrisonAMD: I think we would probably want a primary data-type (input / outputs?), and could have tests that are for different compute / accumulate types or w/e. 

Can always adjust if we find that the mixed data-types makes us want to include more.

jerehartAMD: Changed it to group by datatype and layout!

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
```diff
@@ -21,9 +21,17 @@
...
 
-class BatchnormBwdExecuteGraphTest : public ::testing::TestWithParam<Batchnorm2dTestCase>
+template <typename InputType, typename IntermediateType>
+class BatchnormBwdExecuteGraphBase : public ::testing::TestWithParam<Batchnorm2dTestCase>
 {
 protected:
+    TensorLayout _layout;
+
+    BatchnormBwdExecuteGraphBase(TensorLayout layout)
+        : _layout(std::move(layout))
```
BrianHarrisonAMD, line 31: Oh that's a good way to do it.

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
```diff
@@ -39,206 +47,289 @@ class BatchnormBwdExecuteGraphTest : public ::testing::TestWithParam<Batchnorm2d
...
+    {
+    }
+};
+
+TEST_P(TestGpuMiopenBatchnormBwdExecuteGraphFp32Nchw, Correctness)
 {
     Batchnorm2dTestCase testCase = GetParam();
-    runBwdBatchnormGraph<float, float>(
-        testCase, hipdnn_sdk::data_objects::DataType::DataType_FLOAT, 4e-3f, TensorLayout::NCHW);
+    runBwdBatchnormGraph(testCase, hipdnn_sdk::data_objects::DataType::DataType_FLOAT, 4e-3f);
```
BrianHarrisonAMD, line 257: You can also move the GetParam() into the run method, but not a big deal either way.

BrianHarrisonAMD (approved):
LGTM!

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
```diff
@@ -39,206 +47,289 @@ class BatchnormBwdExecuteGraphTest : public ::testing::TestWithParam<Batchnorm2d
...
-        testCase, hipdnn_sdk::data_objects::DataType::DataType_FLOAT, 4e-3f, TensorLayout::NHWC);
+    runBwdBatchnormGraph(testCase, hipdnn_sdk::data_objects::DataType::DataType_FLOAT, 4e-3f);
 }
 
 // TODO: add unique test suite and conform to naming rules
 
 // MIOpen segfaults for this case, re-enable when fix is released:
 // https://github.com/ROCm/rocm-libraries/pull/1197
-TEST_P(BatchnormBwdExecuteGraphTest, DISABLED_RunBfloat16BwdBatchnormGraphNHWC)
+TEST_P(TestGpuMiopenBatchnormBwdExecuteGraphBfp16Nhwc, DISABLED_Correctness)
```
SamuelReeder: Datatype should be at the very end of the test suite name, layout can be anywhere.

jerehartAMD: Oops, got that backwards, thanks!

SamuelReeder (approved):
LGTM

PR #180, opened 2025-09-03
Cleanup integration tests reducing duplication

BrianHarrisonAMD:
Cleanup integration tests post rename to reduce duplication.
Based off the rename branch.


plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuBatchnormBackward.cpp
```diff
@@ -301,27 +295,27 @@ class BatchnormBackward : public ::testing::TestWithParam<Batchnorm2dTestCase>
     int _deviceId = 0;
 };
 
-class IntegrationGpuBatchnormBackwardNchwFp32 : public BatchnormBackward
+class IntegrationGpuBatchnormBackwardNchwFp32 : public BatchnormBackward<float, float>
 {
 };
```
BrianHarrisonAMD, line 300: You could technically make the layout a template arg as well, but it requires a change to the TensorLayout struct that I didn't like. You need to make the struct a structural class type with everything being constexpr which would require swapping to a fixed size array for the layout name + stride order.

SamuelReeder (approved):
Looks good!

PR #181, opened 2025-09-03
Modified Tensor to handle flexible dimensions

mousdahl-amd:
## Proposed changes

- Modified Tensor to handle flexible dimensions
- Since variadic templated methods can't be virtual / overridden, I remade ITensor into TensorBase
- The root problem is that `Tensor::getIndex()` was running past the end of the strides array in some cases

## Checklist


plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
```diff
@@ -127,7 +127,7 @@ void BatchnormBwdExecuteGraphTest::runBwdBatchnormGraph(
 
     std::vector<int64_t> dims = {testCase.n, testCase.c, testCase.h, testCase.w};
 
-    std::vector<int64_t> derivedDims = {1, dims[1], 1, 1};
+    std::vector<int64_t> derivedDims = {1, dims[1]};
```
BrianHarrisonAMD, line 130: Can't these be 1D now (just dims[1])?

BrianHarrisonAMD:
What change fixed the ASAN error?
I was expecting to see some index out of bounds, but it wasn't instantly obvious.

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
```diff
@@ -101,26 +145,6 @@ class Tensor : public ITensor<T>
...
-        int64_t index = getIndex(nidx, cidx, hidx, widx);
-        auto* data = _memory.hostData();
-        data[index] = value;
-    }
-
-    int64_t getIndex(int64_t nidx, int64_t cidx, int64_t hidx, int64_t widx) const override
-    {
-        return (nidx * _strides[0]) + (cidx * _strides[1]) + (hidx * _strides[2])
-               + (widx * _strides[3]);
-    }
```
mousdahl-amd, line 122: This is where the out-of-bounds was happening.  It was accessing _strides[2] and [3] even if the Tensor only had two dims.

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
mousdahl-amd: I'll try it!

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
BrianHarrisonAMD: Ah okay that makes sense. 
Didn't realize we had cases with only 2 dims.

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
```diff
@@ -34,29 +34,73 @@ inline std::ostream& operator<<(std::ostream& os, const TensorLayout& layout)
...
+        if(indexVector.size() > strides().size())
+        {
+            throw std::invalid_argument("Number of indices (" + std::to_string(indexVector.size())
+                                        + ") must not be greater than the number of dimensions ("
+                                        + std::to_string(strides().size()) + ")");
+        }
+
+        return std::inner_product(
+            indexVector.begin(), indexVector.end(), strides().begin(), int64_t{0});
+    }
```
SamuelReeder, line 72: I think this could be done recursively to avoid vector alloc. Not that it matters much and not even sure if we like recursion 🤔

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
```diff
@@ -34,29 +34,73 @@ inline std::ostream& operator<<(std::ostream& os, const TensorLayout& layout)
...
+    template <typename... Args>
+    void setHostValue(T value, Args... indices)
+    {
+        static_assert(all_of_types<std::is_integral, Args...>::value,
+                      "Indices must be an integral type!");
+
+        int64_t index = getIndex(indices...);
+        auto* data = memory().hostData();
+        data[index] = value;
+    }
```
SamuelReeder: The `static_assert` in each of these can be removed I believe since the same is in `getIndex`.

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
```diff
@@ -34,29 +34,73 @@ inline std::ostream& operator<<(std::ostream& os, const TensorLayout& layout)
 
 // NOLINTBEGIN(portability-template-virtual-member-function)
 
+// Helper to check if all types in a parameter pack satisfy a predicate
+template <template <typename> class Predicate, typename... Ts>
+struct all_of_types : std::conjunction<Predicate<Ts>...>
+{
+};
```
SamuelReeder: This is cool!

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
SamuelReeder: Think we need another constructor overload for this, but probably worth it.

SamuelReeder (approved):
LGTM, comments are all minor.

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
BrianHarrisonAMD: I think this has to be PascalCaseNow doesn't it?

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
BrianHarrisonAMD: For which class?

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
BrianHarrisonAMD: Honestly surprised this isn't flagging.
Maybe out of date, or are we not checking the sdk utilities for tidy issues?

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
SamuelReeder: `Tensor` class for setting dims as an integer. Unless you're saying a single value vector, which is probably the case lol

mousdahl-amd: So MIOpen doesn't play well with non-4D tensors (even for the derived), hence why a lot of these are {1, c, 1, 1}

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
mousdahl-amd: Thanks Sline!  🎆

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
BrianHarrisonAMD: We probably should fix that in the integration then, and still allow it generally.
These should be 1D tensors AFAIK.

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
mousdahl-amd: I'm going to leave it for now as it made the vector size checks a little more complicated.

plugins/miopen_legacy_plugin/tests/TestMiopenBatchnormBwdOperations.cpp
BrianHarrisonAMD: > `Tensor` class for setting dims as an integer. Unless you're saying a single value vector, which is probably the case lol

Yea single value vector.

BrianHarrisonAMD: Nevermind I think 4D is fine for now.

BrianHarrisonAMD (approved):
LGTM

PR #182, opened 2025-09-03
Test name checker

SamuelReeder:
## Proposed changes

Adds a python script to validate test names. This will typically be run during ninja check, and depends on ninja check so that it runs whenever a testing binary is changed.

I opted to add optional unit tests in the script file to ensure the logic is correct. I could also test this script via gtest child processes that run the python, but that doesn't seem great.

## Checklist


SamuelReeder:
This might need a separate PR to fix all the broken test names, but I think we are still awaiting some renames for now.

cmake/scripts/test_name_validator.py
```diff
@@ -0,0 +1,445 @@
...
+
+    print(f"\nWarning: {invalid_count} test(s) have non-conforming names")
+    print(
+        " - For detailed hipDNN test naming rules, see: docs/CodingStyleAndNamingGuidelines.md\n"
+    )
+
+    return 1 if args.strict else 0
+
+
+class TestTestNameValidator(unittest.TestCase):
```
SamuelReeder, line 350: Hehe

CMakeLists.txt
```diff
@@ -120,10 +120,16 @@ endif()
...
 # using append_test_to_check_target
+
+create_test_name_validation_target()
+
 finalize_custom_check_target()
 finalize_unit_check_target()
 finalize_integration_check_target()
 
+add_dependencies(check validate_test_names)
+add_dependencies(check_ctest validate_test_names)
```
SamuelReeder, line 131: Only a dependency of the full `check` targets.

cmake/scripts/test_name_validator.py
SamuelReeder, line 1: This filename could be PascalCase, but is it very unconventional for python in my experience.

BrianHarrisonAMD (approved):
Looks good to me.

How many tests were flagging with this? 
We probably need to fix these as part of this PR, or before we turn this on anyways.

cmake/scripts/test_name_validator.py
BrianHarrisonAMD, line 350: We might want to add these to the CTest test run at some point, but for now it's probably fine.

SamuelReeder:
> Looks good to me.
> 
> How many tests were flagging with this? We probably need to fix these as part of this PR, or before we turn this on anyways.

226 tests! I've skimmed through and they all seem valid, so will wait and see what needs to be fixed when the rest of the renaming is done.

Right now this only gives warnings for non-conforming names so that tests can run. However, if desired, I could make it blocking once we're in a stable state?

cmake/scripts/test_name_validator.py
```diff
@@ -0,0 +1,445 @@
...
+from typing import List, Dict
+
+
+class TestNameValidator:
+
+    KEYWORDS = {
+        "test_types": ["Test", "Integration"],
+        "gpu": ["Gpu"],
+        "datatypes": ["Bfp16", "Fp16", "Fp32", "Fp64"],
+        "shapes": ["Nhwc", "Nchw", "Ndhwc", "Ncdhw"],
```
adickin-amd, line 23: Do these shapes get used anywhere.  I think we removed the requirement to enforce when/where shape names are used.

cmake/Tests.cmake
```diff
@@ -22,6 +22,25 @@ set(UNIT_CHECK_DEPENDS_GLOBAL "" CACHE INTERNAL "Accumulated unit check depends"
 set(INTEGRATION_CHECK_COMMAND_GLOBAL "" CACHE INTERNAL "Accumulated integration check commands" FORCE)
 set(INTEGRATION_CHECK_DEPENDS_GLOBAL "" CACHE INTERNAL "Accumulated integration check depends" FORCE)
 
+function(create_test_name_validation_target)
+    add_custom_command(
+        OUTPUT ${CMAKE_BINARY_DIR}/test_names_validated
+        COMMAND ${CMAKE_CTEST_COMMAND} --show-only=json-v1 > ${CMAKE_BINARY_DIR}/ctest_tests.json
```
adickin-amd: I have a feeling this might break when we swap from using `gtest_discover_tests` to `add_test'.  Problem for future us.

cmake/scripts/test_name_validator.py
SamuelReeder: They are only used in duplicate and case sensitivity checks. Because it flags a lot of layouts which are incorrectly all uppercase.

BrianHarrisonAMD:
> Member

Yea lets get everything merged, then we can do a PR to fix them up and make this a breaking check.

cmake/Tests.cmake
adickin-amd: ah yes, once I swapped its this

```
}root@4c264b28ea16:/data/hipDNN/build# ctest --show-only
Test project /data/hipDNN/build
  Test #1: hipdnn_sdk_tests
  Test #2: hipdnn_backend_tests
  Test #3: hipdnn_frontend_tests
  Test #4: public_hipdnn_backend_tests
  Test #5: public_hipdnn_frontend_tests
  Test #6: miopen_legacy_plugin_tests
  Test #7: miopen_plugin_integration_test

Total Tests: 7
```

SamuelReeder: I'll see if there is quick solution.

SamuelReeder:
```
[1/2] Validating test names by running executables with --gtest_list_tests
Warning: Failed to list tests from tests/frontend/public_hipdnn_frontend_tests: terminate called after throwing an instance of 'std::runtime_error'
  what():  Plugin file not found: ../test_plugins/libtest_good_plugin.so


Test Name Validation Report
============================================================
Total tests found: 845
Valid test names: 723
Invalid test names: 122
```
Get this when using gtest, but I think this fix is best saved for the plugin path stuff.

Edit: fixed by changing working dir to parent of each test binary, but will still investigate more for my next ticket.

BrianHarrisonAMD:
Heads up, I merged the test rename PR for flagged issues into this one.

adickin-amd (approved):

PR #183, opened 2025-09-03
Fix clean rebuild by adding dependency on the generated files

adickin-amd:
## Proposed changes

Clean rebuild could fail if the build order happened incorrectly.  This dependency addition should make it so the files are generated before any other part of the build starts.

## Checklist

- [x ] I have added automated tests relevant to the introduced functionality
- [ x] I have sufficient test coverage for the changes, and code coverage hasn't decreased as a result of my PR
- [ x] I have ran the tests, and they are all passing locally
- [ x] I have ran the tests with ASAN, and they are all passing locally
- [ x] I have added relevant documentation for the changes
- [ x] I have removed the stale documentation which is no longer relevant after this pull request
- [ x] I have ran `make format` & `make check_format` to ensure the changes have been formatted

### DIFF
--- a/sdk/CMakeLists.txt
+++ b/sdk/CMakeLists.txt
@@ -36,13 +36,13 @@ _save_var(FLATBUFFERS_FLATC_SCHEMA_EXTRA_ARGS)
 
 SET(FLATBUFFERS_FLATC_SCHEMA_EXTRA_ARGS "--gen-object-api;--gen-mutable;--gen-compare;--defaults-json")
 build_flatbuffers(
-    "${SCHEMA_FILES}" 
-    "" 
-    generate_hipdnn_sdk_headers 
-    "" 
-    ${HIP_DNN_SDK_INCLUDE_DIR}/hipdnn_sdk/data_objects 
-    "" 
-    ""
+    "${SCHEMA_FILES}" #flatbuffers_schemas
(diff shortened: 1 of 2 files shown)


sdk/CMakeLists.txt
```diff
@@ -57,6 +57,7 @@ hipdnn_add_dependency(spdlog VERSION 1.15.2)
 
 add_library(hipdnn_sdk INTERFACE)
 target_link_libraries(hipdnn_sdk INTERFACE FlatBuffers spdlog_header_only)
+add_dependencies(hipdnn_sdk generate_hipdnn_sdk_headers)
```
BrianHarrisonAMD, line 60: :O

BrianHarrisonAMD: Oh we have it in the frontend/tests lmao.
We should remove it there, and just add this here.
I think I added this just for the tests since there wasn't an SDK target back then.

BrianHarrisonAMD (approved):
LGTM thanks!

Can probably remove the frontend/tests dependency, but it's non-blocking.

sdk/CMakeLists.txt
adickin-amd: removed

mousdahl-amd (approved):

PR #189, opened 2025-09-04
[ALMIOPEN-168] - Adding bfloat16 and half to TensorAttribute value

mousdahl-amd:
## Proposed changes
- Reworking how we use hip half and bfloat16 to remove hip::device link
- Reworked constructors to share the set_value method

## Checklist

### DIFF
--- a/frontend/tests/CMakeLists.txt
+++ b/frontend/tests/CMakeLists.txt
@@ -43,7 +43,6 @@ target_link_libraries(hipdnn_frontend_tests
     GTest::gtest 
     GTest::gmock
     hip::host
-    hip::device
     Threads::Threads 
     hipdnn_sdk 
     hipdnn_frontend
--- a/sdk/CMakeLists.txt
+++ b/sdk/CMakeLists.txt
@@ -63,7 +63,7 @@ target_include_directories(hipdnn_sdk INTERFACE
     $<BUILD_INTERFACE:${HIP_DNN_SDK_INCLUDE_DIR}>
     $<INSTALL_INTERFACE:${HIP_DNN_SDK_INSTALL_INCLUDE_DIR}>
 )
-
+target_compile_definitions(hipdnn_sdk INTERFACE __HIPCC__)
 install( 
     TARGETS hipdnn_sdk FlatBuffers spdlog_header_only
     EXPORT hipdnn_sdk_targets
--- a/sdk/CMakeLists.txt
+++ b/sdk/CMakeLists.txt
@@ -63,7 +63,11 @@ target_include_directories(hipdnn_sdk INTERFACE
     $<BUILD_INTERFACE:${HIP_DNN_SDK_INCLUDE_DIR}>
     $<INSTALL_INTERFACE:${HIP_DNN_SDK_INSTALL_INCLUDE_DIR}>
 )
+
+# This is required to use hip/hip_fp16.h and hip/hip_bfloat16.h in any code that includes the sdk
+# By doing this, we don't need to link to hip::device
 target_compile_definitions(hipdnn_sdk INTERFACE __HIPCC__)
+
 install( 
     TARGETS hipdnn_sdk FlatBuffers spdlog_header_only
     EXPORT hipdnn_sdk_targets
--- a/sdk/include/hipdnn_sdk/utilities/HalfUtils.hpp
+++ b/sdk/include/hipdnn_sdk/utilities/HalfUtils.hpp
@@ -7,21 +7,54 @@
 #include <hipdnn_sdk/logging/Logger.hpp>
 #include <string>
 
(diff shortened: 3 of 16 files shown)


sdk/CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ target_include_directories(hipdnn_sdk INTERFACE
     $<BUILD_INTERFACE:${HIP_DNN_SDK_INCLUDE_DIR}>
     $<INSTALL_INTERFACE:${HIP_DNN_SDK_INSTALL_INCLUDE_DIR}>
 )
-
+target_compile_definitions(hipdnn_sdk INTERFACE __HIPCC__)
```
mousdahl-amd, line 69: This is the compile definition that gets us over the hump of being able to use hip/hip_fp16 and hip/hip_bfloat16.  Since it's interface, it gets pushed up to consumers.

SamuelReeder: I like this!

frontend/tests/CMakeLists.txt
```diff
@@ -43,7 +43,6 @@ target_link_libraries(hipdnn_frontend_tests
     GTest::gtest 
     GTest::gmock
     hip::host
-    hip::device
```
BrianHarrisonAMD, line 47: It's so beautiful.

sdk/CMakeLists.txt
BrianHarrisonAMD: I think we should add a comment just to make it clear why it's here, but I like it!

adickin-amd: I wonder if we comment this line as well to stay this definition make it so we can use bfp16 and fp16 without linking hip::device.  just thinking future us might forget why we did this.

BrianHarrisonAMD (approved):
This looks really good.
Is it possible to do a quick check that using just the frontend and the half / bfp16 types works without any extra effort? 

I think the Samples cover us, but it also includes the SDK.

adickin-amd:
> This looks really good. Is it possible to do a quick check that using just the frontend and the half / bfp16 types works without any extra effort?
> 
> I think the Samples cover us, but it also includes the SDK.

Pretty sure if you add the frontend library its going to add hipdnn_sdk library since the interface library defines it that way

```
target_link_libraries(hipdnn_frontend INTERFACE hipdnn_sdk hipdnn_backend)
```

BrianHarrisonAMD:
> > This looks really good. Is it possible to do a quick check that using just the frontend and the half / bfp16 types works without any extra effort?
> > I think the Samples cover us, but it also includes the SDK.
> 
> Pretty sure if you add the frontend library its going to add hipdnn_sdk library since the interface library defines it that way
> 
> ```
> target_link_libraries(hipdnn_frontend INTERFACE hipdnn_sdk hipdnn_backend)
> ```

Yea I think so as well, but wasn't sure if the define workaround would handle it the way we want.

mousdahl-amd:
> > > This looks really good. Is it possible to do a quick check that using just the frontend and the half / bfp16 types works without any extra effort?
> > > I think the Samples cover us, but it also includes the SDK.
> > 
> > 
> > Pretty sure if you add the frontend library its going to add hipdnn_sdk library since the interface library defines it that way
> > ```
> > target_link_libraries(hipdnn_frontend INTERFACE hipdnn_sdk hipdnn_backend)
> > ```
> 
> Yea I think so as well, but wasn't sure if the define workaround would handle it the way we want.

I just tried it with the samples.  We now only need to find / link frontend now.

sdk/include/hipdnn_sdk/utilities/HalfUtils.hpp
```diff
@@ -3,25 +3,58 @@
...
+    if(__hisnan(a) && !__hisnan(b))
+        return b;
+    if(!__hisnan(a) && __hisnan(b))
+        return a;
+    if(__hisnan(a) && __hisnan(b))
+        return HIPRT_NAN_FP16;
+    if(static_cast<__half_raw>(a).x > static_cast<__half_raw>(b).x)
+        return __half_raw{static_cast<__half_raw>(a).x};
+    return __half_raw{static_cast<__half_raw>(b).x};
+}
```
BrianHarrisonAMD: I suppose we might want `__HOST_DEVICE__` since it's SDK, but we could also just get away with `__HOST__` right?

mousdahl-amd: SDK is why I stuck with host/device.  Damn, now that I think about it, a plugin dev might use `hip/hip_fp16.h` with hip::device defined, and they'll get some collisions here.  I think I'll hide these in a namespace and make them host-only.

sdk/CMakeLists.txt
mousdahl-amd: Comment added.

PR #191, opened 2025-09-04
Followup on unhappy path testing PR

jerehartAMD:
## Proposed changes
- Adds testing for Temp_directory class
- Adds test for multiple plugins, no applicable engines
- Changes warnings into errors and test logs in the Plugin_loading_tests::DuplicateEngineIds and Plugin_loading_tests::IncompleteAPI tests

## Checklist

### DIFF
--- a/sdk/include/hipdnn_sdk/test_utilities/TempDirectory.hpp
+++ b/sdk/include/hipdnn_sdk/test_utilities/TempDirectory.hpp
@@ -2,6 +2,9 @@
 
 #include <filesystem>
 
+namespace hipdnn_sdk::test_utilities
+{
+
 class TempDirectory
 {
     std::filesystem::path _path;
@@ -35,3 +38,5 @@ public:
         }
     }
 };
+
+}
--- a/tests/backend/IntegrationPluginLoading.cpp
+++ b/tests/backend/IntegrationPluginLoading.cpp
@@ -141,9 +141,8 @@ TEST_F(IntegrationPluginLoading, IncorrectEngineID)
 
     ASSERT_EQ(hipdnnBackendFinalize(_engine), HIPDNN_STATUS_BAD_PARAM);
 
-    constexpr size_t BUFFER_SIZE = 512;
-    std::array<char, BUFFER_SIZE> buffer;
-    hipdnnGetLastErrorString(buffer.data(), BUFFER_SIZE);
+    std::array<char, HIPDNN_ERROR_STRING_MAX_LENGTH> buffer;
+    hipdnnGetLastErrorString(buffer.data(), buffer.size());
 
     ASSERT_EQ(
         std::string{buffer.data()},
@@ -161,14 +160,13 @@ TEST_F(IntegrationPluginLoading, DuplicateEngineIds)
 
     ASSERT_EQ(hipdnnCreate(&_handle), HIPDNN_STATUS_SUCCESS);
 
-    constexpr size_t BUFFER_SIZE = 2048;
-    std::array<char, BUFFER_SIZE> buffer;
-    hipdnnGetLastErrorString(buffer.data(), BUFFER_SIZE);
+    std::array<char, HIPDNN_ERROR_STRING_MAX_LENGTH> buffer;
+    hipdnnGetLastErrorString(buffer.data(), buffer.size());
 
     std::string expectedError
         = fmt::format("Engine ID {} already exists",
                       hipdnn_tests::plugin_constants::engineId<DuplicateIdBPlugin>());
-    ;
+
     EXPECT_NE(std::string{buffer.data()}.find(expectedError), std::string::npos);
 
     EXPECT_EQ(test_util::getLoadedPlugins(_handle).size(), 1);
@@ -186,9 +184,8 @@ TEST_F(IntegrationPluginLoading, IncompleteAPI)
 
     ASSERT_EQ(hipdnnCreate(&_handle), HIPDNN_STATUS_SUCCESS);
 
-    constexpr size_t BUFFER_SIZE = 2048;
-    std::array<char, BUFFER_SIZE> buffer;
-    hipdnnGetLastErrorString(buffer.data(), BUFFER_SIZE);
+    std::array<char, HIPDNN_ERROR_STRING_MAX_LENGTH> buffer;
+    hipdnnGetLastErrorString(buffer.data(), buffer.size());
 
     EXPECT_NE(std::string{buffer.data()}.find("Failed to get symbol"), std::string::npos);
     EXPECT_EQ(test_util::getLoadedPlugins(_handle).size(), 0);
--- a/tests/backend/IntegrationPluginLoading.cpp
+++ b/tests/backend/IntegrationPluginLoading.cpp
@@ -14,8 +14,10 @@
 #include <HipdnnBackendHeuristicType.h>
 #include <hipdnn_sdk/test_utilities/TempDirectory.hpp>
 #include <test_plugins/TestPluginConstants.hpp>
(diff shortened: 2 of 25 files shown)


tests/backend/IntegrationPluginLoading.cpp
```diff
@@ -189,7 +161,15 @@ TEST_F(IntegrationPluginLoading, DuplicateEngineIds)
...
     ASSERT_EQ(hipdnnCreate(&_handle), HIPDNN_STATUS_SUCCESS);
 
-    // TODO: Warning is logged, but we don't have means of querying the last warning
+    constexpr size_t BUFFER_SIZE = 2048;
+    std::array<char, BUFFER_SIZE> buffer;
+    hipdnnGetLastErrorString(buffer.data(), BUFFER_SIZE);
+
+    std::string expectedError
+        = fmt::format("Engine ID {} already exists",
+                      hipdnn_tests::plugin_constants::engineId<DuplicateIdBPlugin>());
```
jerehartAMD, line 168: This is sensitive to the plugins being loaded and processed in the order provided to `hipdnnSetEnginePluginPaths_ext` I'm not sure whether or not this is an assumption we should make. It's our code thankfully, so we do get to choose the expected behavior, but this is a little white-boxy as a result unless it's an API expectation

SamuelReeder: Plugin paths are stored in a set so I don't think this can be guaranteed. However, now that you mention it I think it would be good to use a vector so then there is some passive mechanic for priority.

tests/test_plugins/TestNoApplicableEnginesBPlugin.cpp
```diff
@@ -0,0 +1,46 @@
...
+// SPDX-License-Identifier: MIT
+
+#include "TestPluginCommon.hpp"
+#include "TestPluginEngineIdMap.hpp"
+// NOLINTNEXTLINE
+thread_local char
+    hipdnn_plugin::PluginLastErrorManager::s_lastError[HIPDNN_PLUGIN_ERROR_STRING_MAX_LENGTH]
+    = "";
+
+class NoApplicableEnginesAPlugin : public TestPluginBase
```
SamuelReeder: Should this be `NoApplicableEnginesBPlugin`?

tests/backend/IntegrationPluginLoading.cpp
jerehartAMD: Good point, though since we're using an ordered set I think it guarantees that we'll iterate over A before B (regardless of the order we put them in)

tests/test_plugins/TestNoApplicableEnginesBPlugin.cpp
jerehartAMD: It definitely should, and it's not good that the test didn't catch that. I think one of these engines should have been rejected for a duplicate ID in this case

tests/backend/IntegrationPluginLoading.cpp
SamuelReeder: Yeah I'm just thinking it's more useful for the load order to correspond to the user's input ordering.

tests/test_plugins/TestNoApplicableEnginesBPlugin.cpp
jerehartAMD: Unfortunately I don't think I can find a way to ensure that there were no errors in the load, because the only way to query that is to check the last error which is a non-resettable global state. I can imagine that this would also be challenging for a user who wants to be able to respond in code to one of the plugins not being loaded

sdk/include/hipdnn_sdk/test_utilities/TempDirectory.hpp
```diff
@@ -2,6 +2,9 @@
 
 #include <filesystem>
 
+namespace hipdnn_sdk::test_utilities
+{
+
```
BrianHarrisonAMD, line 7: Ah nice catch!

tests/backend/IntegrationPluginLoading.cpp
BrianHarrisonAMD: Yea I think if you consume this, then you would expect order matters.

BrianHarrisonAMD: Can just add a separate ticket for this, or maybe @SamuelReeder can make it part of the plugin path clean-up work.

tests/backend/IntegrationPluginLoading.cpp
```diff
@@ -206,15 +186,79 @@ TEST_F(IntegrationPluginLoading, IncompleteAPI)
 
     ASSERT_EQ(hipdnnCreate(&_handle), HIPDNN_STATUS_SUCCESS);
 
-    // TODO: Warning is logged, but we don't have means of querying the last warning
+    constexpr size_t BUFFER_SIZE = 2048;
+    std::array<char, BUFFER_SIZE> buffer;
+    hipdnnGetLastErrorString(buffer.data(), BUFFER_SIZE);
```
BrianHarrisonAMD: I thought we added a define for max buffer size that could be used here instead (not a big deal though).

jerehartAMD: I didn't know that existed! I'll use that instead

tests/backend/IntegrationPluginLoading.cpp
```diff
@@ -189,7 +161,15 @@ TEST_F(IntegrationPluginLoading, DuplicateEngineIds)
...
+
+    std::string expectedError
+        = fmt::format("Engine ID {} already exists",
+                      hipdnn_tests::plugin_constants::engineId<DuplicateIdBPlugin>());
+    ;
```
BrianHarrisonAMD: extra ;?

tests/test_plugins/TestNoApplicableEnginesBPlugin.cpp
```diff
@@ -0,0 +1,46 @@
...
+thread_local char
+    hipdnn_plugin::PluginLastErrorManager::s_lastError[HIPDNN_PLUGIN_ERROR_STRING_MAX_LENGTH]
+    = "";
+
+class NoApplicableEnginesBPlugin : public TestPluginBase
+{
+public:
+    const char* getPluginName() const override
+    {
+        return "test_NoApplicableEnginesPlugin";
```
BrianHarrisonAMD: Should it be test_NoApplicableEnginesBPlugin?

tests/test_plugins/TestNoApplicableEnginesAPlugin.cpp
BrianHarrisonAMD, line 1: Should the name be test_NoApplicableEnginesPlugin -> test_NoApplicableEnginesAPlugin now?

BrianHarrisonAMD (approved):
Looks good to me!

Just some minor comments.

tests/backend/IntegrationPluginLoading.cpp
jerehartAMD: I assume this is MAX_ERROR_STRING_SIZE? That's set to 256 right now, and with filepaths I've seen the error strings get over the limit already. Should we consider increasing this or tending not to include filepaths?

In particular, in our build process there is something that specifically tries to make sufficiently long paths, and that plays very poorly with that

BrianHarrisonAMD: I would be fine increasing it.
I think I wanted it to be bigger originally as well.
256 is tiny.

BrianHarrisonAMD: Just make sure all the spots that init the error size are updated, and the plugins as well.

PR #193, opened 2025-09-05
Add exports + ensure install paths are relative

AaronStGeorge:
## Proposed changes

This PR adds exports so that a consuming project can build against hipDNN from either the build or install directory. It also updates install targets to use relative rather than absolute install locations. ${CMAKE_INSTALL_PREFIX} in path created an absolute path. With absolute paths, `cmake --install build --prefix <something>` installs at the absolute path irrespective of the prefix provided.

## Checklist

### DIFF
--- a/backend/src/CMakeLists.txt
+++ b/backend/src/CMakeLists.txt
@@ -79,6 +79,11 @@ install(
     EXPORT hipdnn_backend_targets
 )
 
+export(
+    TARGETS hipdnn_backend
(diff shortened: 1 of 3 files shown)


BrianHarrisonAMD:
Thanks!

Looks good to me, but lets get @adickin-amd to review first.
Probably @mousdahl-amd as well since he's been touching the CMake files recently.

adickin-amd (approved):

BrianHarrisonAMD (approved):

PR #194, opened 2025-09-05
[ALMIOPEN-256] Transitioned from std::ignore to [[maybe_unused]]

mousdahl-amd:
## Proposed changes
- Transitioned away from using `std::ignore` and swapped to `[[maybe_unused]]`
- Did a brief regex search for commented-out function / method args, didn't find any

## Checklist


jerehartAMD (approved):
LGTM!

BrianHarrisonAMD:
I think there was also some [[maybe_unused]]'s that were incorrectly added that we saw in the frontend Types.hpp, but we can get that in another PR.

PR #198, opened 2025-09-08
Logging initialization fixes & general cleanup

SamuelReeder:
## Proposed changes

Logging initialization needs to use `COMPONENT_NAME` compile-time definition so that logs can be associated with the correct logger based on their origin. However, some places did not correctly use this variable, so I fixed these spots.

In addition, this add lazy initialization to frontend logging and modifies some tests to use the frontend logger instead, since our logging mechanism is not intended for consumers to write their own logs.

Lastly, it adds tests for untested logging helpers and add logging messages in some critical areas.

## Checklist


frontend/tests/CMakeLists.txt
```diff
@@ -36,9 +36,6 @@ add_executable(hipdnn_frontend_tests
 )
 
 target_compile_options(hipdnn_frontend_tests PRIVATE ${HIPDNN_WARNING_COMPILE_OPTIONS})
-target_compile_definitions(hipdnn_frontend_tests PRIVATE
```
SamuelReeder, line 39: The frontend already defines this in its header, so it propagates to consumers. Therefore, we can use the existing define from the frontend, with the downside being that logs directly from the tests are labelled as `hipdnn_frontend` logs. I'm thinking of other ways to do this, but its a bit tricky with the frontend having the `INTERFACE` property.

BrianHarrisonAMD: I think that's correct.
Since the logs are coming from the frontend right?

BrianHarrisonAMD: We should only need to define this in apps that aren't using the frontend directly.
We don't want consumers needing to define this in their project in order to get the frontend logs.

plugins/miopen_legacy_plugin/tests/CMakeLists.txt
```diff
@@ -47,6 +47,8 @@ target_link_libraries(miopen_legacy_plugin_tests PUBLIC
     hip::host
     miopen_legacy_plugin
     miopen_legacy_plugin_private
+    hipdnn_sdk
+    hipdnn_backend
```
BrianHarrisonAMD: I don't think the unit tests should depend on hipdnn_backend.
Is this required?

plugins/miopen_legacy_plugin/tests/main.cpp
```diff
@@ -5,16 +5,15 @@ SPDX-License-Identifier: MIT
 
 #include <gtest/gtest.h>
 
+#include <hipdnn_backend.h>
```
BrianHarrisonAMD: Why is this needed?

plugins/miopen_legacy_plugin/tests/CMakeLists.txt
BrianHarrisonAMD: SDK is interesting that we didn't have already, but I assume it must be just due to being header only?

plugins/miopen_legacy_plugin/integration_tests/main.cpp
```diff
@@ -5,16 +5,16 @@ SPDX-License-Identifier: MIT
 
 #include <gtest/gtest.h>
 
+#include <hipdnn_backend.h>
```
BrianHarrisonAMD: Is this needed? 

I thought it would only need the SDK includes, and not need backend here.

plugins/miopen_legacy_plugin/tests/main.cpp
SamuelReeder: I didn't get all them yet, but I was changing the tests to use the actual callback so they can correctly use the env vars. I figured it was fine to link backend only for testing binaries, but I can revert if not.

BrianHarrisonAMD: Well for integration tests I could see it, but for these they shouldn't be going through the backend at all.

frontend/tests/CMakeLists.txt
SamuelReeder: The frontend has it's own logs and then some tests that consume the frontend log too, so I now changed it to initialize frontend logging and then initialize their own logging so logs can come from either source. However, I still don't think its great because the `COMPONENT_NAME` needs to be undefined and redefined.

plugins/miopen_legacy_plugin/integration_tests/main.cpp
BrianHarrisonAMD: (since these are integration tests, this could make sense.)

But then we should be using the frontend method to init. and not backend directly.

frontend/tests/CMakeLists.txt
SamuelReeder: The thing is that I don't think the tests which have their own logs put them all under the `hipdnn_frontend` umbrella, since that should probably be for the logs directly within the frontend.

frontend/include/hipdnn_frontend/Utilities.hpp
```diff
@@ -73,7 +74,7 @@ inline TensorAttributes
 
 }
 
-inline int32_t initializeFrontendLogging(hipdnnCallback_t fn)
+inline int32_t initializeFrontendLogging(hipdnnCallback_t fn = hipdnnLoggingCallback_ext)
```
BrianHarrisonAMD, line 80: Lets add a task to have two versions of this.
One that does the default behavior without an extra override.
And another one that allows providing a custom callback logging sink that consumers can use to swap our entire logging system over.

I.E if I provide a custom callback logging function, then all logs (Frontend, backend, SDK, Plugins) will go through that callback function instead.

frontend/include/hipdnn_frontend/Utilities.hpp
```diff
@@ -73,8 +76,28 @@ inline TensorAttributes
 
 }
 
-inline int32_t initializeFrontendLogging(hipdnnCallback_t fn)
+inline bool isLoggingEnabled()
 {
+    static bool loggingEnabled = []() {
+        auto logLevel = hipdnn_sdk::utilities::getEnv("HIPDNN_LOG_LEVEL", "off");
+        return !logLevel.empty() && logLevel != "off";
```
BrianHarrisonAMD: Does the empty check match what we do in the backend?
From a quick check, it sort of looks like it since it'll skip setting the level if it's not matching what we expect.

Maybe we should just instead check it's one of ["info", "warn", "error", "fatal"] to match backend?
Could maybe throw a utility for this in the SDK and share it between both places, but not the biggest deal.

frontend/include/hipdnn_frontend/Utilities.hpp
```diff
@@ -73,8 +76,28 @@ inline TensorAttributes
...
+    static bool loggingInitialized = false;
+    if(loggingInitialized)
+    {
+        return 0;
+    }
+
+    if(!isLoggingEnabled())
+    {
+        return 0;
+    }
```
BrianHarrisonAMD: Should set to true here right?
We don't need to keep rechecking once it's not enabled.

I guess it doesn't really matter since it caches the result there as well actually.
Probably can leave it actually.

frontend/include/hipdnn_frontend/Utilities.hpp
SamuelReeder: Oh yes should check the value is valid. That's a good catch.

frontend/include/hipdnn_frontend/Utilities.hpp
```diff
@@ -86,9 +109,38 @@ inline int32_t initializeFrontendLogging(hipdnnCallback_t fn)
...
+    } while(0)
+
+#define HIPDNN_FE_LOG_ERROR(...)                      \
+    do                                                \
+    {                                                 \
+        hipdnn_frontend::initializeFrontendLogging(); \
+        HIPDNN_LOG_ERROR(__VA_ARGS__);                \
+    } while(0)
+
+#define HIPDNN_FE_LOG_CRITICAL(...)                   \
```
BrianHarrisonAMD: I don't see anywhere flipped to these macros. 
I think there was a spot for ScopedHipdnnBackendDescriptor, but we probably should add a couple info logs to graph at least.

SamuelReeder: This should be FATAL.

BrianHarrisonAMD: Heh our macro is FATAL, but the enum is critical for spdlog it looks like.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuBatchnormForwardInference.cpp
```diff
@@ -243,8 +242,6 @@ class BatchnormForwardInference : public ::testing::TestWithParam<Batchnorm2dTes
         auto intermediateDataType = getDataTypeEnumFromType<IntermediateType>();
 
         unsigned int seed = std::random_device{}();
-        //log the random seed in case we need to reproduce the test
-        HIPDNN_LOG_INFO("Test is using {} for its random seed", seed);
```
BrianHarrisonAMD: I do think we should keep logging these random seeds even though it'll say "frontend". 

These can be nice to have for debugging.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuBatchnormBackward.cpp
```diff
@@ -264,7 +261,6 @@ class BatchnormBackward : public ::testing::TestWithParam<Batchnorm2dTestCase>
         auto intermediateDataType = getDataTypeEnumFromType<IntermediateType>();
 
         unsigned int seed = std::random_device{}();
-        HIPDNN_LOG_INFO("Test is using {} for its random seed", seed);
```
BrianHarrisonAMD: Same here, probably can keep this.

frontend/include/hipdnn_frontend/Utilities.hpp
SamuelReeder: I could just switch to critical lol. There aren't many uses of that macro.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuBatchnormForwardInference.cpp
adickin-amd: honestly if the test says frontend I dont think its weird at all.  Its nice to have test logging and I dont want to have to have another mechanism setup in order to log in tests.

adickin-amd: as long as we can still log when we need then Im all good with the changes

frontend/include/hipdnn_frontend/Utilities.hpp
SamuelReeder: Added to backlog.

backend/tests/TestBackendLogger.cpp
```diff
@@ -11,6 +11,7 @@
 #include <thread>
 #include <vector>
 
+#include <hipdnn_sdk/test_utilities/EnvironmentVariableGuard.hpp>
```
BrianHarrisonAMD: Probably want to call this Scoped to match the other RAII style classes, but this is a nice addition!

sdk/include/hipdnn_sdk/test_utilities/EnvironmentVariableGuard.hpp
```diff
@@ -0,0 +1,46 @@
...
+{
+
+class EnvironmentVariableGuard
+{
+public:
+    explicit EnvironmentVariableGuard(const std::string& varName)
+        : _varName(varName)
+    {
+        _originalValue = hipdnn_sdk::utilities::getEnv(varName.c_str(), "");
+        _hadOriginalValue = !hipdnn_sdk::utilities::getEnv(varName.c_str()).empty();
```
BrianHarrisonAMD: Can reuse the same getEnv I think instead of querying twice.

sdk/include/hipdnn_sdk/test_utilities/EnvironmentVariableGuard.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include <hipdnn_sdk/utilities/PlatformUtils.hpp>
+#include <string>
+
+namespace hipdnn_sdk::test_utilities
+{
+
+class EnvironmentVariableGuard
+{
+public:
+    explicit EnvironmentVariableGuard(const std::string& varName)
```
BrianHarrisonAMD: Maybe `ScopedEnvironmentVariableSetter`, or something similar to match other RAII styles instead?

BrianHarrisonAMD (approved):
Looking really nice!

I think those minor changes to EnvironmentVariableGuard would be good, but not blocking.

PR #200, opened 2025-09-09
Stabilize loading of plugins for tests and create test project for running tests elsewhere

adickin-amd:
## Proposed changes

- Cleanup tests to load plugins relative to the backend library or test binary (when necessary) rather than the working directory
- During build create a test CmakeLists.txt thats installed into build/test_project.  This project can then be used to run the test suite anywhere you copy our build folder to as all paths are relative
- Add CMake variables to specify where to dump plugins
- Separate build and install dirs into bin and lib.
- Allow CMake to resolve some paths at runtime.

## Checklist


backend/tests/TestBackendLogger.cpp
```diff
@@ -14,29 +14,56 @@
 #include <hipdnn_sdk/utilities/PlatformUtils.hpp>
 #include <logging/Logging.hpp>
 
+namespace
+{
+constexpr const char* LOG_LEVEL_ENV_VAR = "HIPDNN_LOG_LEVEL";
```
adickin-amd: remove all changes to this file

backend/tests/main.cpp
```diff
@@ -5,10 +5,13 @@ SPDX-License-Identifier: MIT
 
 #include "logging/Logging.hpp"
 #include <gtest/gtest.h>
+#include <hipdnn_sdk/utilities/PlatformUtils.hpp>
 
 int main(int argc, char** argv)
 {
-    ::testing::InitGoogleTest(&argc, argv);
+    hipdnn_sdk::utilities::setEnv("HIPDNN_LOG_LEVEL", "info");
```
SamuelReeder: Will want to remove this.

backend/src/plugin/PluginCore.hpp
```diff
@@ -161,7 +161,16 @@ class PluginManagerBase
...
+                auto resolvedPath = path;
+                if(path.is_relative())
+                {
+                    // If the path is relative, resolve it to our current module directory instead of allowing
+                    // std::filesystem to resolve it relative to the current working directory.
+                    resolvedPath = hipdnn_backend::platform_utilities::getCurrentModuleDirectory()
+                                   / resolvedPath;
+                }
+
+                resolvedPath = std::filesystem::weakly_canonical(resolvedPath);
```
SamuelReeder, line 173: I think we can leave this for now, and I will refine it in follow up PR. I fixed the tets by using absolute paths from cmake for now.

sdk/tests/utilities/TestPlatformUtils.cpp
```diff
@@ -0,0 +1,12 @@
+// Copyright © Advanced Micro Devices, Inc., or its affiliates.
+// SPDX-License-Identifier:  MIT
+
+#include <gtest/gtest.h>
+#include <hipdnn_sdk/utilities/PlatformUtils.hpp>
+#include <iostream>
+
+// TODO: Additional tests
+TEST(TestPlatformUtils, Stub)
```
SamuelReeder: Need to test PlatformUtils.

backend/tests/TestPlugin.cpp
```diff
@@ -77,13 +77,15 @@ bool TestPluginCallback::s_callbackCalled = false;
 const std::string PLUGIN_NAME1 = "hipdnn_test_plugin1";
 const std::string PLUGIN_NAME2 = "hipdnn_test_plugin2";
 
-const std::filesystem::path PLUGIN_PATH1 = std::filesystem::path(".") /= PLUGIN_NAME1;
-const std::filesystem::path PLUGIN_PATH2 = std::filesystem::path(".") /= PLUGIN_NAME2;
+const std::filesystem::path PLUGIN_PATH1 = "../lib/test_plugins/" + PLUGIN_NAME1;
+const std::filesystem::path PLUGIN_PATH2 = "../lib/test_plugins/" + PLUGIN_NAME2;
```
SamuelReeder: ~~Move to a variable.~~

tests/test_plugins/TestPluginConstants.hpp
```diff
@@ -13,25 +13,38 @@ namespace hipdnn_tests
...
+        = fs::path(getTestPluginCustomDir()) / hipdnn_sdk::utilities::getLibraryName(pluginName);
 
     return pluginFile.string();
 }
 
+inline std::string getDefaultPluginPath()
+{
+    namespace fs = std::filesystem;
+    return (fs::path(getTestPluginDefaultDir())
+            / hipdnn_sdk::utilities::getLibraryName("test_good_default_plugin"))
```
SamuelReeder: Compile definition for this.

tests/backend/IntegrationSetPluginPathsExt.cpp
```diff
@@ -90,17 +93,26 @@ TEST(IntegrationSetPluginPathsExt, GetLoadedPluginPathsLoadsDefault)
...
-    fs::path expectedPluginPath = fs::path("../../backend/src/hipdnn_plugins/engines")
-                                  / getLibraryName("test_good_default_plugin");
+    std::string expectedPluginPath = getDefaultPluginPath();
+
+    std::cout << "Expected default plugin path: " << expectedPluginPath << "\n";
+    std::cout << "Loaded plugins:\n";
+    for(const auto& plugin : loadedPlugins)
+    {
+        std::cout << "  " << plugin << "\n";
+    }
```
SamuelReeder: Remove.

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/GraphWrapper.hpp
```diff
@@ -58,7 +58,12 @@ class GraphWrapper : public IGraph
...
     {
         throwIfNotValid();
-        return _shallowGraph->nodes()->size();
+        auto nodes = _shallowGraph->nodes();
+        if(nodes == nullptr)
+        {
+            return 0;
+        }
+        return static_cast<uint32_t>(nodes->size());
     }
```
SamuelReeder, line 67: This sometimes segfaulted when plugin weren't loaded, and I want to look into it more.

backend/tests/TestSharedLibrary.cpp
```diff
@@ -16,14 +16,19 @@ using namespace hipdnn_backend;
...
+const auto LIBRARY_PATH_LIB_NAME_ONLY
+    = ".." / TEST_PLUGIN_DIR / hipdnn_sdk::utilities::getLibraryName(TEST_PLUGIN1_NAME);
+
+const auto WRONG_LIBRARY_PATH = std::filesystem::path("./wrong_path");
+const auto SYMBOL_NAME = std::string("hipdnnPluginGetName");
+const auto WRONG_SYMBOL_NAME = std::string("wrong_symbol_name");
+
+const auto FULL_LIBRARY_PATH
+    = (hipdnn_backend::platform_utilities::getCurrentModuleDirectory().parent_path()
+       / TEST_PLUGIN_DIR / hipdnn_sdk::utilities::getLibraryName(TEST_PLUGIN1_NAME));
```
SamuelReeder: This and the global constants in `backend/tests/TestPlugin.cpp` could probably be done better with a test fixture, but this PR may already have too much going on.

sdk/tests/utilities/TestPlatformUtils.cpp
SamuelReeder: I tested one new function but we should add a separate ticket to write tests for the rest -- with specific test cases for Windows and Linux.

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/GraphWrapper.hpp
SamuelReeder: Will look into this today, as it may be somewhat unrelated.

backend/tests/main.cpp
```diff
@@ -5,10 +5,12 @@ SPDX-License-Identifier: MIT
 
 #include "logging/Logging.hpp"
 #include <gtest/gtest.h>
+#include <hipdnn_sdk/utilities/PlatformUtils.hpp>
 
 int main(int argc, char** argv)
 {
-    ::testing::InitGoogleTest(&argc, argv);
+    hipdnn_backend::logging::initialize();
```
adickin-amd, line 12: will this log info for the test by default now?

SamuelReeder: No, only if `HIPDNN_LOG_LEVEL` is set.

backend/tests/TestPlugin.cpp
```diff
@@ -340,3 +353,47 @@ TEST_F(TestPluginCallback, SetLoggingCallback)
...
+
+    ASSERT_EQ(result.size(), 2);
+    EXPECT_EQ(result, defaultPaths);
+}
+
+TEST(TestPluginCore, GetPluginSearchPathsWithEmptyEnvVar)
+{
+    const char* envVarName = "TEST_PLUGIN_PATH_EMPTY";
+
+    ScopedEnvironmentVariableSetter envSetter(envVarName);
```
adickin-amd, line 388: feels like the constructor for this should just take in the new value we want set.

SamuelReeder: Yeah I like that, and will add a method to change its value after initialization too.

sdk/tests/utilities/TestPlatformUtils.cpp
```diff
@@ -0,0 +1,33 @@
+// Copyright © Advanced Micro Devices, Inc., or its affiliates.
+// SPDX-License-Identifier:  MIT
+
+#include <filesystem>
+#include <gtest/gtest.h>
+#include <hipdnn_sdk/utilities/PlatformUtils.hpp>
+#include <iostream>
+
+// TODO: More tests for PlatformUtils since it is untested
```
adickin-amd, line 9: do we need todo this?

SamuelReeder: Yeah we haven't directly tested it yet. I will add a ticket if no objections.

CMakeLists.txt
```diff
@@ -28,6 +28,27 @@ project(
...
+# RPATH settings for build portability
+# So binaries can find libraries when the build directory is moved
+set(CMAKE_BUILD_WITH_INSTALL_RPATH OFF)
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH FALSE)
+
+set(CMAKE_BUILD_RPATH_USE_ORIGIN TRUE)
+# Set default build RPATH for executables in bin/ to find libraries in lib/
+set(CMAKE_BUILD_RPATH 
+    "$ORIGIN/../lib"
+)
```
BrianHarrisonAMD: Do we need these for everything, or just for the testing executables?

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/GraphWrapper.hpp
BrianHarrisonAMD: Was this fixed with the nullptr check?
Or still seeing random issues?

BrianHarrisonAMD (approved):
Looks good to me.

Nice!

CMakeLists.txt
SamuelReeder: It's only for testing executables, but I thought it would be easier to keep it global since those are the only executables we produce. I can move if you prefer it be more specific.

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/GraphWrapper.hpp
SamuelReeder: Seemed to be fixed with nullptr check.

CMakeLists.txt
BrianHarrisonAMD: I think it might be nice to make a global function called like, "MakeExecutableRelocatable", set these, and then call it in each spot we want it, but I am also fine leaving it as is.

backend/tests/main.cpp
adickin-amd: i guess im the only one who likes the logging on by default.

sdk/include/hipdnn_sdk/plugin/flatbuffer_utilities/GraphWrapper.hpp
SamuelReeder: I think the issue is that in GraphWrapper there are several dereferences of optional attributes that are not populated for  all tests. I think this was because I was sometimes loading the MIOpen plugin instead of test plugins, and the MIOpen plugin uses GraphWrapper and the test plugins don't. So for cases where the graph had optional attributes absent, it would throw whereas the test plugins never accessed the graph. 

I'm adding some checks so that these nullptrs are handled.

PR #201, opened 2025-09-11
Update frontend pointwise API to include all modes, and include additional valid logic

BrianHarrisonAMD:
## Proposed changes

Update pointwise frontend API to include the additional modes, and also add additional verification logic.
Pointwise now checks that inputs are correctly matching the operation.
Pointwise also now checks that the dimensions are broadcastable. 

## Checklist


SamuelReeder:
Are we going to add pointwise support to the miopen plugin soon?

frontend/include/hipdnn_frontend/Types.hpp
```diff
@@ -161,4 +299,70 @@ inline std::ostream& operator<<(std::ostream& os, const DataType& type)
     return os << to_string(type);
 }
 
+inline const std::set<PointwiseMode>& unaryPointwiseModes()
+{
+    static const std::set<PointwiseMode> unaryPointwiseModes = {PointwiseMode::ABS,
```
jerehartAMD: I'm not sure if we expect pointwise to get significantly more operations, but currently these *could* be implemented using a bitmask. We could even just make a small helper class that takes a list of enums, sets the corresponding bits, and has a contains operation, or just use std::bitset maybe with a helper function that constructs it

frontend/include/hipdnn_frontend/node/PointwiseNode.hpp
```diff
@@ -78,17 +133,17 @@ class PointwiseNode : public BaseNode<PointwiseNode>
...
         {
-            std::vector<std::shared_ptr<TensorAttributes>> allInputs
-                = {in0, attributes.get_input_1(), attributes.get_input_2()};
-
-            for(const auto& tensor : allInputs)
+            for(const auto& [_, tensor] : attributes.inputs)
             {
                 if(tensor && tensor->get_dim() == out->get_dim())
                 {
+                    HIPDNN_LOG_INFO(
```
SamuelReeder: Merged logging PR so will want to change to `HIPDNN_FE_LOG_INFO`.

frontend/tests/TestPointwiseNode.cpp
```diff
@@ -3,234 +3,619 @@
...
+protected:
+    void SetUp() override
+    {
+        _mode = GetParam();
+    }
+
+    PointwiseMode _mode;
+};
+
+TEST_P(TestPointwiseNodeUnaryOps, validWithOneInput)
```
SamuelReeder: This file has wrong style for test case name. Should be PascalCase.

SamuelReeder: Should emphasize this more in cline rules maybe.

SamuelReeder (approved):
LGTM besides those minor things.

frontend/tests/TestPointwiseNode.cpp
SamuelReeder: I need to make sure the test name validator caught this, but maybe that's why CI failed? Will check when able.

Edit: yeah it did.

frontend/include/hipdnn_frontend/node/PointwiseNode.hpp
```diff
@@ -37,6 +35,65 @@ class PointwiseNode : public BaseNode<PointwiseNode>
...
+
+        // Check that ALL inputs are compatible with output dimensions
+        auto output = attributes.get_output_0();
+        if(output && !output->get_dim().empty())
+        {
+            const auto& outputDims = output->get_dim();
+
+            for(const auto& [_, inputTensor] : attributes.inputs)
+            {
+                if(inputTensor)
```
jerehartAMD: Should an error be returned if it's null? This appears elsewhere too

frontend/include/hipdnn_frontend/Types.hpp
jerehartAMD: I think I like bitset the most, especially since it is trivially extendable as opposed to a u64

frontend/tests/TestPointwiseNode.cpp
BrianHarrisonAMD: Ah didn't realize for test case names, nice catch!

frontend/include/hipdnn_frontend/node/PointwiseNode.hpp
BrianHarrisonAMD: Honestly, I was debating this, and I think I agree error makes more sense.

frontend/include/hipdnn_frontend/Types.hpp
BrianHarrisonAMD: I think it's worth leaving it flexible like this, but any strong reason it would be better to use a bitmask?

frontend/include/hipdnn_frontend/node/PointwiseNode.hpp
SamuelReeder: I thought it was fine for input tensors to be null? Since don't always need all of them.

SamuelReeder: Well I guess operation type dictates the size of the input now, so a null check can't hurt.

BrianHarrisonAMD: If they are ever specified in the input map, then they should be set I think.

BrianHarrisonAMD: In other nodes I think we can explicitly check, since they have defined inputs.
Pointwise is probably the only case where this type of check is needed.

frontend/include/hipdnn_frontend/Types.hpp
jerehartAMD: Not a strong one, but a bitset would be a fair bit faster than the set and, with a function to create it from a list of enums, wouldn't result in more complex code. I definitely think it's fine as is, but it is an easy performance boost

frontend/include/hipdnn_frontend/node/PointwiseNode.hpp
```diff
@@ -37,6 +35,65 @@ class PointwiseNode : public BaseNode<PointwiseNode>
             return {ErrorCode::ATTRIBUTE_NOT_SET,
                     "PointwiseNode missing operation for pre-validation"};
         }
+        if(attributes.outputs.size() != 1)
+        {
+            return {ErrorCode::INVALID_VALUE, "PointwiseNode must have exactly one output"};
+        }
+
+        // Validate input count matches operation type
+        switch(attributes.inputs.size())
```
adickin-amd: considering the comment above.  why not refactor into a function `validateInputCountMatchesOperationType`.  Same with below `checkInputsAreCompatableWithOutputDimensions`

Up to you, its just my own preferences to have named functions rather than loose comments.

frontend/include/hipdnn_frontend/Types.hpp
```diff
@@ -23,7 +23,53 @@ typedef ConvolutionMode ConvolutionMode_t; // NOLINT(readability-identifier-nami
 enum class PointwiseMode
 {
     NOT_SET = 0,
-    RELU_FWD = 1,
+    ABS = 1,
+    ADD = 2,
+    ADD_SQUARE = 3,
```
adickin-amd, line 31: How many of these do we plan to support?  Would it make sense to comment out the ones we wont support?

frontend/tests/TestPointwiseNode.cpp
adickin-amd: Did we actually specify its PascalCase for TestCaseNames?  We might want to add that, I do think it looks better in PascalCase for Tests.

frontend/include/hipdnn_frontend/Types.hpp
BrianHarrisonAMD: Bitset makes it expandable as well.
Ill give that a go, thanks!

sdk/include/hipdnn_sdk/utilities/ShapeUtils.hpp
```diff
@@ -12,6 +12,40 @@ namespace hipdnn_sdk
 namespace utilities
 {
 
+// Check broadcastability: input can be broadcast to output
```
adickin-amd, line 15: You didnt name the files but ShapeUtils.hpp doesnt match the test file name `TestShapeUtiltites.cpp` can we make it match?

frontend/tests/TestPointwiseNode.cpp
adickin-amd: slick, the check works already.  Nice work @SamuelReeder 

```TestPointwiseNode.missingOutputError               [FAIL]    

  → Test name does not match expected PascalCase format 'TestSuite.TestCase' or 'TestSuite/Instance.TestCase' without special characters.
```

frontend/include/hipdnn_frontend/Types.hpp
BrianHarrisonAMD: I think we eventually will plan to support them all.
We will have limited support that gets expanded over time.

I think prefer leaving them defined for now, and documenting the patterns we do support as they become available, but I am open to suggestions. Since it's a graphAPI with limited support, we will always have configurations that might not have an engine that can support them, and this seems like a similar case.

frontend/include/hipdnn_frontend/node/PointwiseNode.hpp
BrianHarrisonAMD: Yea that seems decent.
It's a pretty long function as is, and this change makes it more readable.

Will do!

BrianHarrisonAMD:
> Are we going to add pointwise support to the miopen plugin soon?

Probably not right away, but instead just focus on priority fusion cases that need pointwise.

.clinerules
```diff
@@ -39,8 +39,8 @@
 - Use TEST(), TEST_F(), or TEST_P() macros as appropriate
 
 #### Test Naming Guidelines
-
-Rules below apply ONLY to the TestSuite name (first parameter of `TEST` / `TEST_F` / `TEST_P`). The TestCase (second parameter) can be descriptive but should still avoid the reserved keywords where noted.
+TestCase should be PascalCase.
```
BrianHarrisonAMD: Whoops white space, and maybe order swap.

frontend/include/hipdnn_frontend/Types.hpp
```diff
@@ -161,4 +302,109 @@ inline std::ostream& operator<<(std::ostream& os, const DataType& type)
...
+{
+    static const PointwiseModeBitset ternaryModes = []() {
+        PointwiseModeBitset bitset;
+        bitset.set(toBitPosition(PointwiseMode::BINARY_SELECT));
+        return bitset;
+    }();
+    return ternaryModes;
+}
+
+// Fast O(1) check functions
```
BrianHarrisonAMD: aaa

frontend/include/hipdnn_frontend/node/PointwiseNode.hpp
```diff
@@ -37,19 +35,22 @@ class PointwiseNode : public BaseNode<PointwiseNode>
             return {ErrorCode::ATTRIBUTE_NOT_SET,
                     "PointwiseNode missing operation for pre-validation"};
         }
+        if(attributes.outputs.size() != 1)
+        {
+            return {ErrorCode::INVALID_VALUE, "PointwiseNode must have exactly one output"};
+        }
```
jerehartAMD, line 41: This feels a lot cleaner

jerehartAMD (approved):
Looks good to me!

SamuelReeder (approved):
Looks good! Bitset is a nice enhancement.

frontend/tests/TestPointwiseNode.cpp
```diff
@@ -3,234 +3,788 @@
...
-    auto offset = node.pack_node(builder);
-    EXPECT_NE(offset.o, 0);
+    auto error = node.infer_properties_node();
+    EXPECT_EQ(error.code, ErrorCode::INVALID_VALUE);
+}
 
-    builder.Finish(offset);
-    auto bufferPointer = builder.GetBufferPointer();
-    auto nodeFlatbuffer = flatbuffers::GetRoot<hipdnn_sdk::data_objects::Node>(bufferPointer);
+TEST(TestPointwiseNode, nullInputAmongMultipleInferProperties)
```
SamuelReeder: Ah some are still wrong. Saw in CI failure.

SamuelReeder:
```
TestPointwiseNode.nullInputAmongMultipleInferProperties [FAIL]    
  → Test name does not match expected PascalCase format 'TestSuite.TestCase' or 'TestSuite/Instance.TestCase' without special characters.
TestPointwiseNode.nullInputAmongMultipleInputsPreValidation [FAIL]    
  → Test name does not match expected PascalCase format 'TestSuite.TestCase' or 'TestSuite/Instance.TestCase' without special characters.
TestPointwiseNode.nullInputTensorInferProperties   [FAIL]    
  → Test name does not match expected PascalCase format 'TestSuite.TestCase' or 'TestSuite/Instance.TestCase' without special characters.
TestPointwiseNode.nullInputTensorPreValidation     [FAIL]    
  → Test name does not match expected PascalCase format 'TestSuite.TestCase' or 'TestSuite/Instance.TestCase' without special characters.
TestPointwiseNode.nullSecondInputInferProperties   [FAIL]    
  → Test name does not match expected PascalCase format 'TestSuite.TestCase' or 'TestSuite/Instance.TestCase' without special characters.
  ```

adickin-amd (approved):
approved, thanks for the refactor

PR #203, opened 2025-09-12
Add new dockerfile using TheRock for environment init instead of ROCm release

BrianHarrisonAMD:
## Proposed changes

This change updates the hipDNN dockerfile for dev environment set up to use TheRock instead of the released ROCm artifacts.
I also removed the almalinux option, as this isn't being utilized currently, and in the future we will use TheRock for many linux builds.

Note looks like the build from source option has a break due to this [issue](https://github.com/ROCm/TheRock/issues/1481) in TheRock.

## Checklist


dockerfiles/Dockerfile.ubuntu24
```diff
@@ -0,0 +1,171 @@
...
+    libegl1-mesa-dev \
+    curl \
+    unzip \
+    patchelf \
+    libtool \
+    python3-pip \
+    vim \
+    rsync \
+    ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
```
SamuelReeder: It seems a bit redundant to install these deps twice. Could probably make a common base stage with these deps to use for both build types.

BrianHarrisonAMD: Yea good point.
Ill make that change.

dockerfiles/README.md
```diff
@@ -1,55 +1,77 @@
...
+The Dockerfile supports two build types: **prebuilt** (using nightly tarballs) and **fullbuild** (building from source).
+
+### Build Arguments
+
+#### 🔧 Common Arguments (Both Build Types)
+
+| Argument | Default | Description | Valid Values |
+|----------|---------|-------------|--------------|
+| `BUILD_TYPE` | `prebuilt` | Selects build method | `prebuilt`, `fullbuild` |
+| `THEROCK_ASIC` | `gfx94X` (prebuilt)<br>`gfx90a` (fullbuild) | GPU architecture target | `gfx94X`, `gfx950`, `gfx110X`, `gfx90a`, etc. |
```
SamuelReeder: Maybe it would be helpful to provide this [link](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/reference/system-requirements.html) so users can match their GPU to an arch target. However, if it's expected that this is easy to find then I wouldn't worry.

SamuelReeder (approved):
LGTM!

dockerfiles/README.md
BrianHarrisonAMD: I think the list for TheRock is actually a bit different, since I see gfx94X on here.

Ill see if I can get a link for both.

BrianHarrisonAMD: Probably this [link](https://github.com/ROCm/TheRock/blob/main/RELEASES.md) for the prebuilt.
And the link you provided for self build.

dockerfiles/Dockerfile.ubuntu24
```diff
@@ -0,0 +1,124 @@
...
+    curl \
+    unzip \
+    patchelf \
+    libtool \
+    python3-pip \
+    vim \
+    rsync \
+    ca-certificates \
+    wget \
+    && rm -rf /var/lib/apt/lists/*
```
SamuelReeder: Should this have gdb? I also see other ROCm-related deps we had in the old image are absent here, but I assume those are outdated with TheRock?

BrianHarrisonAMD: No I think it's all grabbed from the docker build of TheRock.

Ill re-test, but when I had this set up, my build + tests were all working nicely from CLI and vscode with debugging through vscode.

BrianHarrisonAMD: Hmm actually different this time. Let me check.

BrianHarrisonAMD: Hmm I think you are right, ill add it back.

dockerfiles/README.md
adickin-amd: You may also want this one. https://github.com/ROCm/TheRock/blob/main/cmake/therock_amdgpu_targets.cmake#L45-L115 And also link to here https://github.com/ROCm/TheRock/blob/main/ROADMAP.md#rocm-on-linux since all asics are not fully supported.

BrianHarrisonAMD: Ill add the ROADMAP one, but linking to cmake file itself seems unlikely to remain valid hehe.

adickin-amd (approved):

PR #205, opened 2025-09-16
Update tensor to support vector based indices

BrianHarrisonAMD:
## Proposed changes

Update Tensor.hpp to support get / set by providing a vector for indices.
This change makes it easier to handle n-dim support for normalization in a future PR.

## Checklist

### DIFF
--- a/sdk/include/hipdnn_sdk/utilities/Tensor.hpp
+++ b/sdk/include/hipdnn_sdk/utilities/Tensor.hpp
@@ -71,6 +71,26 @@ public:
             indexVector.begin(), indexVector.end(), strides().begin(), int64_t{0});
     }
 
+    template <typename IndexType>
+    int64_t getIndex(const std::vector<IndexType>& indices) const
+    {
+        static_assert(std::is_integral<IndexType>::value, "Index type must be integral!");
+
+        if(indices.size() > strides().size())
+        {
+            throw std::invalid_argument("Number of indices (" + std::to_string(indices.size())
+                                        + ") must not be greater than the number of dimensions ("
+                                        + std::to_string(strides().size()) + ")");
+        }
+
+        int64_t result = 0;
+        for(size_t i = 0; i < indices.size(); ++i)
+        {
+            result += static_cast<int64_t>(indices[i]) * strides()[i];
+        }
+        return result;
+    }
+
     template <typename... Args>
     T getHostValue(Args... indices) const
     {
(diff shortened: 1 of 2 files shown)


sdk/include/hipdnn_sdk/utilities/Tensor.hpp
```diff
@@ -79,6 +99,16 @@ class TensorBase
         return data[index];
     }
 
+    template <typename IndexType>
+    T getHostValue(const std::vector<IndexType>& indices) const
+    {
+        static_assert(std::is_integral<IndexType>::value, "Index type must be integral!");
```
SamuelReeder: `static_assert` might not be necessary since `getIndex` does it.

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
```diff
@@ -71,6 +71,26 @@ class TensorBase
...
+    int64_t getIndex(const std::vector<IndexType>& indices) const
+    {
+        static_assert(std::is_integral<IndexType>::value, "Index type must be integral!");
+
+        if(indices.size() > strides().size())
+        {
+            throw std::invalid_argument("Number of indices (" + std::to_string(indices.size())
+                                        + ") must not be greater than the number of dimensions ("
+                                        + std::to_string(strides().size()) + ")");
+        }
```
SamuelReeder: Does the class enforce that `dims().size() == strides().size()`? Since this only checks the strides.

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
```diff
@@ -71,6 +71,26 @@ class TensorBase
             indexVector.begin(), indexVector.end(), strides().begin(), int64_t{0});
     }
 
+    template <typename IndexType>
+    int64_t getIndex(const std::vector<IndexType>& indices) const
```
BrianHarrisonAMD, line 67: Oh the variadic version above makes it into an index vector.
I could unify these a bit.

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
BrianHarrisonAMD: Ill have to check what the concrete impls do, but this base doesn't.
Not sure it really matters in this case though, since we are just leveraging strides for this calc.

I could add in a check that basically fails gracefully if we try to access elements outside the elementCount.

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
```diff
@@ -79,6 +99,16 @@ class TensorBase
         return data[index];
     }
 
+    template <typename IndexType>
+    T getHostValue(const std::vector<IndexType>& indices) const
```
SamuelReeder, line 90: Having an `operator()` overload that calls this function could be syntactically nice, but it's just an option.

SamuelReeder (approved):
LGTM with some minor comments.

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
BrianHarrisonAMD: Looks like the concrete impl below does ensure they match.
Not sure if we need to add extra checks here, since I think this is intended to be like array accessors.

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
BrianHarrisonAMD: Yea we should have that.

BrianHarrisonAMD: Probably make it a future thing though, and get this in as is.

PR #206, opened 2025-09-16
Add clang format from apt-get instead of forcing /opt/rocm version

BrianHarrisonAMD:
## Proposed changes

With swapping docker to use TheRock version, clang-format is no longer included in /opt/rocm/llvm/bin. 
Swapping to instead use installed clang-format in the environment, and provide the apt-get version when installing.

Note:

May need to see how clang-format works in CI, and see if it flags these new fixes as problematic.

## Checklist


jerehartAMD (approved):
LGTM!

PR #207, opened 2025-09-16
Add BN fwd training ref impl

BrianHarrisonAMD:
## Proposed changes

Add ref implementation for BN fwd training.

## Checklist

### DIFF
--- a/sdk/include/hipdnn_sdk/test_utilities/CpuFpReferenceBatchnorm.hpp
+++ b/sdk/include/hipdnn_sdk/test_utilities/CpuFpReferenceBatchnorm.hpp
@@ -128,34 +128,37 @@ public:
             MeanVarianceDataType varianceAccum = 0.0;
 
             // Calculate mean and variance for this channel
-            iterateChannelElements(x, cidx, [&](const std::vector<int64_t>& indices) {
-                auto inVal = x.getHostValue(indices);
-                meanAccum += inVal;
-                varianceAccum += inVal * inVal;
-            });
+            iterateChannelElements(
+                x, cidx, elementsPerChannel, [&](const std::vector<int64_t>& indices) {
+                    auto inVal = x.getHostValue(indices);
+                    meanAccum += inVal;
+                    varianceAccum += inVal * inVal;
+                });
 
-            meanAccum /= nhw;
-            varianceAccum /= nhw;
-            varianceAccum -= (meanAccum * meanAccum);
+            MeanVarianceDataType channelMean = meanAccum /= nhw;
+            MeanVarianceDataType channelVariance
+                = (varianceAccum / nhw) - (channelMean * channelMean);
 
             auto invVar
-                = static_cast<MeanVarianceDataType>(1.0) / sqrtInternal(varianceAccum + epsilon);
+                = static_cast<MeanVarianceDataType>(1.0) / sqrtInternal(channelVariance + epsilon);
 
             // Apply normalization with scale and bias
-            iterateChannelElements(x, cidx, [&](const std::vector<int64_t>& indices) {
-                auto xVal = static_cast<MeanVarianceDataType>(x.getHostValue(indices));
-                auto xHat = (xVal - static_cast<MeanVarianceDataType>(meanAccum)) * invVar;
+            iterateChannelElements(
+                x, cidx, elementsPerChannel, [&](const std::vector<int64_t>& indices) {
+                    auto xVal = static_cast<MeanVarianceDataType>(x.getHostValue(indices));
+                    auto xHat = (xVal - channelMean) * invVar;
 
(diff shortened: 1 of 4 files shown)


sdk/tests/test_utilities/TestCpuFpReferenceBatchnorm.cpp
```diff
@@ -293,3 +293,392 @@ TEST(TestCpuFpReferenceBatchnormFp64, BatchnormBwdSanityValidationNchw)
...
+
+    CpuFpReferenceBatchnormImpl<float, float>::batchnormFwdTraining(
+        inputTensor, scaleTensor, biasTensor, outputTensor, 1e-5f, 0.1f);
+}
+
+TEST(TestCpuFpReferenceBatchnormFp32, BatchnormFwdTraining5D)
+{
+    // Test with 5D tensor (batch, channel, depth, height, width)
+    Tensor<float> inputTensor({2, 3, 4, 5, 6});
+    Tensor<float> outputTensor({2, 3, 4, 5, 6});
```
BrianHarrisonAMD: Trying to figure out why these are having issues with ASAN.

sdk/include/hipdnn_sdk/test_utilities/CpuFpReferenceBatchnorm.hpp
```diff
@@ -190,6 +306,60 @@ class CpuFpReferenceBatchnormImpl
     {
         return sqrtf(value);
     }
+
+    // Utility method to iterate over all elements for a specific channel in an N-dimensional tensor
+    static void iterateChannelElements(const TensorBase<InputDataType>& tensor,
```
SamuelReeder, line 314: Doing something similar for conv doesn't seem easy, but that's a problem for another day :)

BrianHarrisonAMD: Yea absolutely not hehe.
You can probably at best do it over groups.

sdk/include/hipdnn_sdk/test_utilities/CpuFpReferenceBatchnorm.hpp
```diff
@@ -190,6 +306,60 @@ class CpuFpReferenceBatchnormImpl
...
+                                       int64_t channelIdx,
+                                       std::function<void(const std::vector<int64_t>&)> func)
```
SamuelReeder: Could have an optional arg to pass in a value for `totalIterations`, since it looks like it's calculated the same way as `elementsPerChannel` in the ref.

sdk/include/hipdnn_sdk/test_utilities/CpuFpReferenceBatchnorm.hpp
```diff
@@ -89,6 +89,122 @@ class CpuFpReferenceBatchnormImpl
...
+            {
+                constexpr MeanVarianceDataType one = static_cast<MeanVarianceDataType>(1.0f);
+                auto currentMean = prevRunningMean->getHostValue(0, cidx);
+                auto newMean = (one - momentum) * currentMean + momentum * meanAccum;
+                nextRunningMean->setHostValue(newMean, 0, cidx);
+
+                auto currentVar = prevRunningVariance->getHostValue(0, cidx);
+                // Apply Bessel's correction for unbiased variance estimate
+                auto adjustedVariance
+                    = (nhw == one) ? varianceAccum : (nhw / (nhw - one)) * varianceAccum;
```
SamuelReeder: Super tiny nit: `meanAccum` and `varianceAccum` variable names give the impression these variables are only for accumulating the sums.

SamuelReeder (approved):
LGTM.

sdk/tests/test_utilities/TestCpuFpReferenceBatchnorm.cpp
BrianHarrisonAMD: Turns out it was silently problematic due to assuming NCHW layout, and providing 5D dims.
When we add full 5D support we probably need to change the constructor to instead default to decreasing stride order.

sdk/include/hipdnn_sdk/test_utilities/CpuFpReferenceBatchnorm.hpp
BrianHarrisonAMD: Ill just add it since no reason to not provide it in this private utility.

sdk/include/hipdnn_sdk/test_utilities/CpuFpReferenceBatchnorm.hpp
BrianHarrisonAMD: Agreed, and some extra useless casts.
Fixing.

PR #209, opened 2025-09-16
Initial CPU fusion prototype

adickin-amd:
## Proposed changes

Initial cpu graph execution prototype.  The prototype can take the flatbuffers serialized graph type and a variant pack and find a CPU executor for each node type as long as the signature is one of the defined ones in BatchnormSignature.hpp.

This initial prototype only handles fwd batchnorm inference but I think the current batchnorm registry can be expanded to handle all batch norm operations inside the same registry because the `BatchnormSignatureKey` can be used for all ops. 

Todo
- Tests

## Checklist


sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormBuilder.hpp
```diff
@@ -0,0 +1,73 @@
...
+template <>
+struct DataTypeToNative<hipdnn_sdk::data_objects::DataType::DataType_HALF>
+{
+    using type = half;
+};
+
+template <hipdnn_sdk::data_objects::DataType T>
+concept BatchnormDataType = (T == hipdnn_sdk::data_objects::DataType::DataType_FLOAT)
+                            || (T == hipdnn_sdk::data_objects::DataType::DataType_HALF)
+                            || (T == hipdnn_sdk::data_objects::DataType::DataType_BFLOAT16);
```
adickin-amd: todo, add bfloat16 support

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormBuilder.hpp
```diff
@@ -0,0 +1,73 @@
...
+concept BatchnormSignatureDescriptor = requires(T t) {
+    { t.INPUT_DATA_TYPE } -> std::convertible_to<hipdnn_sdk::data_objects::DataType>;
+    { t.SCALE_BIAS_DATA_TYPE } -> std::convertible_to<hipdnn_sdk::data_objects::DataType>;
+    { t.MEAN_VARIANCE_DATA_TYPE } -> std::convertible_to<hipdnn_sdk::data_objects::DataType>;
+    { t.NODE_ATTRIBUTES_TYPE } -> std::convertible_to<hipdnn_sdk::data_objects::NodeAttributes>;
+};
+
+template <BatchnormSignatureDescriptor auto SIGNATURE>
+    requires ValidBatchnormSignature<SIGNATURE>
+struct BatchnormBuilder
```
adickin-amd: todo, rename to BatchnormFwdInferenceBuilder

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistrar.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+namespace hipdnn_sdk
+{
+namespace test_utilities
+{
+
+using BatchnormFwdInferenceFn
+    = std::function<void(std::any&, std::any&, std::any&, std::any&, std::any&, std::any&, double)>;
+
+// Registry keyed by BatchnormSignatureKey
```
adickin-amd: todo, add comments explaining what each function in here is doing.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/CpuReferenceGraphExecutor.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+    static void
+        execute(void* graphBuffer, size_t size, std::unordered_map<int64_t, void*>& variantPack)
+    {
+        auto graphWrap = hipdnn_plugin::GraphWrapper(graphBuffer, size);
+
+        for(uint32_t i = 0; i < graphWrap.nodeCount(); i++)
+        {
+            auto& node = graphWrap.getNode(i);
+            const auto* nodeAttributes = node.attributes_as_BatchnormInferenceAttributes();
```
adickin-amd, line 35: can probably refactor each node type out into separate functions

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/CpuReferenceGraphExecutor.hpp
```diff
@@ -0,0 +1,98 @@
...
+public:
+    CpuReferenceGraphExecutor() = default;
+    ~CpuReferenceGraphExecutor() = default;
+
+    static void
+        execute(void* graphBuffer, size_t size, std::unordered_map<int64_t, void*>& variantPack)
+    {
+        auto graphWrap = hipdnn_plugin::GraphWrapper(graphBuffer, size);
+
+        for(uint32_t i = 0; i < graphWrap.nodeCount(); i++)
```
adickin-amd, line 32: todo, make test with 2 nodes

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/CpuReferenceGraphExecutor.hpp
```diff
@@ -0,0 +1,98 @@
...
+        {
+            auto& node = graphWrap.getNode(i);
+            const auto* nodeAttributes = node.attributes_as_BatchnormInferenceAttributes();
+            if(nodeAttributes != nullptr)
+            {
+                const auto& tensorMap = graphWrap.getTensorMap();
+                auto xTensorAttr = tensorMap.at(nodeAttributes->x_tensor_uid());
+                auto scaleTensorAttr = tensorMap.at(nodeAttributes->scale_tensor_uid());
+                //todo, its optional so use scale if not provided
```
adickin-amd, line 41: add optional support for mean/variance

sdk/include/hipdnn_sdk/utilities/ShallowTensor.hpp
```diff
@@ -0,0 +1,160 @@
...
+#include <hipdnn_sdk/utilities/ShapeUtils.hpp>
+#include <hipdnn_sdk/utilities/Tensor.hpp>
+
+namespace hipdnn_sdk
+{
+namespace utilities
+{
+
+template <class T>
+class ShallowHostOnlyMigratableMemory : public IMigratableMemory<T>
```
adickin-amd: split into own file.

sdk/include/hipdnn_sdk/utilities/ShallowTensor.hpp
```diff
@@ -0,0 +1,160 @@
...
+#pragma once
+
+#include <iostream>
+#include <numeric>
+#include <random>
+#include <vector>
+
+#include <hipdnn_sdk/utilities/MigratableMemory.hpp>
+#include <hipdnn_sdk/utilities/ShapeUtils.hpp>
+#include <hipdnn_sdk/utilities/Tensor.hpp>
```
adickin-amd, line 13: test both classes in here

sdk/include/hipdnn_sdk/utilities/Tensor.hpp
```diff
@@ -46,6 +46,8 @@ class TensorBase
 public:
     virtual ~TensorBase() = default;
 
+    using value_type = T;
```
adickin-amd, line 53: check if I still need this.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistrar.hpp
```diff
@@ -0,0 +1,131 @@
...
+template <typename F>
+inline void forEachBatchnormSignatureAlt(F&& f)
+{
+    using Variant = BatchnormSignatureVariants;
+    _forEachVariantAltImpl<Variant>(std::forward<F>(f),
+                                    std::make_index_sequence<std::variant_size_v<Variant>>{});
+}
+
+template <typename BatchnormSignature>
+inline BatchnormSignatureKey makeKeyFromSigTag(std::type_identity<BatchnormSignature>)
```
adickin-amd: rename this SigTag stuff to just Signature

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormBuilder.hpp
```diff
@@ -0,0 +1,73 @@
...
+struct DataTypeToNative<hipdnn_sdk::data_objects::DataType::DataType_FLOAT>
+{
+    using type = float;
+};
+
+template <>
+struct DataTypeToNative<hipdnn_sdk::data_objects::DataType::DataType_HALF>
+{
+    using type = half;
+};
```
BrianHarrisonAMD: Can we generalize this?
I feel like we should be able to take in a DataType enum, and use a constexpr method for extracting the type.

I was pretty sure we have either type -> enum, or enum -> type method already in one of the utilities.

BrianHarrisonAMD: Hmm looks like a macro, or a modified constexpr method is the only way.

```

namespace {
    template <typename T>
    struct TypeWrapper {
        using type = T;
    };
    
    // Mapping function
    template <hipdnn_sdk::data_objects::DataType DT>
    constexpr auto getTypeWrapper() {
        using DataType = hipdnn_sdk::data_objects::DataType;
        
        if constexpr (DT == DataType::DataType_FLOAT) {
            return TypeWrapper<float>{};
        } else if constexpr (DT == DataType::DataType_HALF) {
            return TypeWrapper<half>{};
        } else if constexpr (DT == DataType::DataType_DOUBLE) {
            return TypeWrapper<double>{};
        } else if constexpr (DT == DataType::DataType_INT32) {
            return TypeWrapper<int32_t>{};
        } else if constexpr (DT == DataType::DataType_INT64) {
            return TypeWrapper<int64_t>{};
        } else if constexpr (DT == DataType::DataType_BFP16) {
            return TypeWrapper<hip_bfloat16>{};
        } else {
            static_assert(DT != DT, "Unsupported DataType");
        }
    }
}

template <hipdnn_sdk::data_objects::DataType DT>
struct DataTypeToNative {
    using type = typename decltype(detail::getTypeWrapper<DT>())::type;
};

template <hipdnn_sdk::data_objects::DataType DT>
using DataTypeToNative_t = typename DataTypeToNative<DT>::type;
```

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/CpuReferenceGraphExecutor.hpp
adickin-amd: save for fusions

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/CpuReferenceGraphExecutor.hpp
adickin-amd: dont need it, were removing optional support for those from graphs.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormBuilder.hpp
BrianHarrisonAMD: Regardless of which option we pick, this should probably move to a shared location.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormBuilder.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+template <hipdnn_sdk::data_objects::DataType T>
+concept BatchnormDataType = (T == hipdnn_sdk::data_objects::DataType::DataType_FLOAT)
+                            || (T == hipdnn_sdk::data_objects::DataType::DataType_HALF)
+                            || (T == hipdnn_sdk::data_objects::DataType::DataType_BFLOAT16);
+
+template <hipdnn_sdk::data_objects::NodeAttributes T>
+concept NodeAttributesType
+    = (T == hipdnn_sdk::data_objects::NodeAttributes_BatchnormInferenceAttributes)
+      || (T == hipdnn_sdk::data_objects::NodeAttributes_BatchnormBackwardAttributes);
```
BrianHarrisonAMD: Should this be only BatchnormInference, or should we also add BatchnormAttributes here for training as well?

Unclear if this is fwd specific, or general for all BN.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormBuilder.hpp
```diff
@@ -0,0 +1,73 @@
...
+template <typename T>
+concept BatchnormSignatureDescriptor = requires(T t) {
+    { t.INPUT_DATA_TYPE } -> std::convertible_to<hipdnn_sdk::data_objects::DataType>;
+    { t.SCALE_BIAS_DATA_TYPE } -> std::convertible_to<hipdnn_sdk::data_objects::DataType>;
+    { t.MEAN_VARIANCE_DATA_TYPE } -> std::convertible_to<hipdnn_sdk::data_objects::DataType>;
+    { t.NODE_ATTRIBUTES_TYPE } -> std::convertible_to<hipdnn_sdk::data_objects::NodeAttributes>;
+};
+
+template <BatchnormSignatureDescriptor auto SIGNATURE>
+    requires ValidBatchnormSignature<SIGNATURE>
```
BrianHarrisonAMD: Concepts so nice.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignature.hpp
```diff
@@ -0,0 +1,94 @@
...
+                           DataType_FLOAT,
+                           DataType_FLOAT,
+                           DataType_FLOAT,
+                           NodeAttributes_BatchnormInferenceAttributes);
+
+DEFINE_BATCHNORM_SIGNATURE(FwdBatchnormSignatureHalf,
+                           DataType_HALF,
+                           DataType_HALF,
+                           DataType_HALF,
+                           NodeAttributes_BatchnormInferenceAttributes);
```
BrianHarrisonAMD, line 73: bf16 as well right?

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignature.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+using BatchnormSignatureVariants
+    = std::variant<FwdBatchnormSignatureFloat, FwdBatchnormSignatureHalf>;
+
```
BrianHarrisonAMD: bf16 as well for these I think.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignature.hpp
```diff
@@ -0,0 +1,94 @@
...
+template <>
+struct hash<hipdnn_sdk::test_utilities::BatchnormSignatureKey>
+{
+    std::size_t operator()(const hipdnn_sdk::test_utilities::BatchnormSignatureKey& k) const
+    {
+        return std::hash<int>()(static_cast<int>(k.inputDataType))
+               ^ (std::hash<int>()(static_cast<int>(k.scaleBiasDataType)) << 1)
+               ^ (std::hash<int>()(static_cast<int>(k.meanVarianceDataType)) << 1)
+               ^ (std::hash<int>()(static_cast<int>(k.nodeAttributesType)) << 1);
+    }
```
BrianHarrisonAMD: I think we want to shift these by different amounts when constructing the hash.

Got this spicy impl from copilot:
```
template <>
struct hash<hipdnn_sdk::test_utilities::BatchnormSignatureKey>
{
    std::size_t operator()(const hipdnn_sdk::test_utilities::BatchnormSignatureKey& k) const
    {
        // Use a prime multiplier for better distribution
        constexpr std::size_t prime1 = 73856093;
        constexpr std::size_t prime2 = 19349663;
        constexpr std::size_t prime3 = 83492791;
        constexpr std::size_t prime4 = 47055833;
        
        auto h1 = std::hash<int>{}(static_cast<int>(k.inputDataType));
        auto h2 = std::hash<int>{}(static_cast<int>(k.scaleBiasDataType));
        auto h3 = std::hash<int>{}(static_cast<int>(k.meanVarianceDataType));
        auto h4 = std::hash<int>{}(static_cast<int>(k.nodeAttributesType));
        
        // Combine using prime multiplication and rotation
        std::size_t result = h1 * prime1;
        result = (result << 7) | (result >> (sizeof(std::size_t) * 8 - 7));  // Rotate left by 7
        result ^= h2 * prime2;
        result = (result << 11) | (result >> (sizeof(std::size_t) * 8 - 11)); // Rotate left by 11
        result ^= h3 * prime3;
        result = (result << 13) | (result >> (sizeof(std::size_t) * 8 - 13)); // Rotate left by 13
        result ^= h4 * prime4;
        
        return result;
    }
};
```

Probably overkill, but using the same bit seems undesirable.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistrar.hpp
```diff
@@ -0,0 +1,131 @@
...
+#include <hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignature.hpp>
+#include <hipdnn_sdk/test_utilities/cpu_graph_executor/TensorVariant.hpp>
+
+namespace hipdnn_sdk
+{
+namespace test_utilities
+{
+
+using BatchnormFwdInferenceFn
+    = std::function<void(std::any&, std::any&, std::any&, std::any&, std::any&, std::any&, double)>;
```
BrianHarrisonAMD: Might want to also make epsilon a tensor that is passed by value, but I guess you can see how it works in practice.
The ref impls pass by value without the tensor, but the frontend API passes by value through tensors. 

Might make it more consistent / easier if every arg is a tensor.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormBuilder.hpp
adickin-amd: this will be generalized but we dont support it yet so its not there.

adickin-amd: next PR you would expect bwd and train here.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistrar.hpp
```diff
@@ -0,0 +1,131 @@
...
+    BatchnormRegistryInitializer()
+    {
+        forEachBatchnormSignatureAlt([](auto signature) {
+            auto key = makeKeyFromBatchnormSignature(signature);
+            batchnormRegistry()[key] = makeBatchnormFwdInferenceFnForSignature(signature);
+        });
+    }
+};
+
+static BatchnormRegistryInitializer _batchnormRegistryInitializer;
```
BrianHarrisonAMD: I think this should be inline instead to avoid ODR issues.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormBuilder.hpp
shumway: Yes, I'd recommend Brian's suggestion, but you can simplify this a lot (CoPilot and I agree on this):


template <hipdnn_sdk::data_objects::DataType DT>
consteval auto data_type_to_native() {
    using enum hipdnn_sdk::data_objects::DataType;
    if constexpr (DT == DataType_FLOAT) return float{};
    else if constexpr (DT == DataType_HALF) return half{};
    else if constexpr (DT == DataType_DOUBLE) return double{};
    else if constexpr (DT == DataType_INT32) return int32_t{};
    else if constexpr (DT == DataType_INT64) return int64_t{};
    else if constexpr (DT == DataType_BFP16) return hip_bfloat16{};
    else static_assert(DT != DT, "Unsupported DataType");
}

template <hipdnn_sdk::data_objects::DataType DT>
using DataTypeToNative_t = decltype(data_type_to_native<DT>());

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormBuilder.hpp
adickin-amd: adding later, hit issue with bfloat16 that needs to be fixed first.  I filed  story for it

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistrar.hpp
adickin-amd: sure

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistrar.hpp
adickin-amd: we can do this, but I think ill add it to next PR

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignature.hpp
adickin-amd: oh geez thats complexs, ill just change bits

adickin-amd: i just shifted by 1, 2, then 3

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignature.hpp
adickin-amd: next PR, after resolving bfloat16 issues

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignature.hpp
```diff
@@ -0,0 +1,94 @@
...
+#include <hipdnn_sdk/data_objects/data_types_generated.h>
+#include <hipdnn_sdk/data_objects/graph_generated.h>
+#include <hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormBuilder.hpp>
+
+namespace hipdnn_sdk
+{
+namespace test_utilities
+{
+
+struct BatchnormSignatureKey
```
shumway: I think this is all you want, just a BatchnormSignature struct.

You can make constexpr values of this type. You don't need all that macro stuff below.
1. Just one type: BatchnormSignature
2. Each supported signature is a constexpr value of that type.
3. You can even just put them in a `constexpr std::Array<BatchnormSignature, N> BATCHNORM SIGNTURE VARIANTS`

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistrar.hpp
```diff
@@ -0,0 +1,131 @@
...
+// Registry keyed by BatchnormSignatureKey
+inline std::unordered_map<BatchnormSignatureKey, BatchnormFwdInferenceFn>& batchnormRegistry()
+{
+    static std::unordered_map<BatchnormSignatureKey, BatchnormFwdInferenceFn> _reg;
+    return _reg;
+}
+
+// Compile-time iteration over variant alternatives
+template <typename Variant, typename F, std::size_t... Is>
+inline void _forEachVariantAltImpl(F&& f, std::index_sequence<Is...>)
```
shumway: horrors!!! what is this!?!?

Just use `constexpr std::array` to hold your variants and use a loop in a constexpr function.

The whole point of NTTP is that you can just have structs and arrays and functions. None of this template metaprogramming.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormBuilder.hpp
adickin-amd: moved to shared location.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/CpuReferenceGraphExecutor.hpp
adickin-amd: later once we add more

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/CpuReferenceGraphExecutor.hpp
```diff
@@ -0,0 +1,100 @@
...
+    static void
+        execute(void* graphBuffer, size_t size, std::unordered_map<int64_t, void*>& variantPack)
+    {
+        auto graphWrap = hipdnn_plugin::GraphWrapper(graphBuffer, size);
+
+        for(uint32_t i = 0; i < graphWrap.nodeCount(); i++)
+        {
+            auto& node = graphWrap.getNode(i);
+            const auto* nodeAttributes = node.attributes_as_BatchnormInferenceAttributes();
+            if(nodeAttributes != nullptr)
```
shumway, line 36: Nit: if you write this as a guard clause, you reduce indentation and make logic more clear:

if (nodeAttributes == nullptr) {
    throw std::runtime_error("Unsupported node attributes type");
}
// normal execution continues...
const auto& tensorMp = ...

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/CpuReferenceGraphExecutor.hpp
```diff
@@ -0,0 +1,100 @@
...
+                auto meanTensorAttr = tensorMap.at(nodeAttributes->mean_tensor_uid().value());
+
+                BatchnormSignatureKey key{
+                    .inputDataType = xTensorAttr->data_type(),
+                    .scaleBiasDataType = scaleTensorAttr->data_type(),
+                    .meanVarianceDataType = meanTensorAttr->data_type(),
+                    .nodeAttributesType
+                    = hipdnn_sdk::data_objects::NodeAttributes::BatchnormInferenceAttributes,
+                };
+                auto it = batchnormRegistry().find(key);
```
shumway, line 48: Nit: guard clause?

if (it == batchnormRegistry().end()) {
   throw...
}
// continue outside if
BatchnormInferenceFn fn = ...

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/CpuReferenceGraphExecutor.hpp
```diff
@@ -0,0 +1,100 @@
...
+#include <hipdnn_sdk/utilities/UtilsFp16.hpp>
+
+#include <hipdnn_sdk/test_utilities/cpu_graph_executor/TensorVariant.hpp>
+
+namespace hipdnn_sdk
+{
+namespace test_utilities
+{
+
+class CpuReferenceGraphExecutor
```
shumway, line 21: This is really confusing me. First, this looks like a function CpuReferenceGraphExecutor::execute, not a class.

That function immediately uses its two arguments to get a "graphWrap", which appears to be an ordered container of nodes. So this is really just a function that operates on a vector-like graphWrap and a map variantPack.

That looks like we are just mapping over the graphWrap and getting node attributes and  functions. That to me sounds like a good function: (graphWrap, variantPack) -> a list of node attributes their functions (plus throwing errors.

Then you can make an execute function that takes that list of node attributes and functions and executes each one.

What I'm concerned about here is that there are a lot of generic, reusable functions (or at least simple functions), but instead you have a class. There's little chance for refactoring and reducing code duplication with this implementation.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistrar.hpp
```diff
@@ -0,0 +1,131 @@
...
+        using MeanVarianceT =
+            typename DataTypeToNative<BatchnormSignature::MEAN_VARIANCE_DATA_TYPE>::type;
+
+        auto& inTensor = *std::get<std::unique_ptr<TensorBase<InputT>>>(inVar);
+        auto& scTensor = *std::get<std::unique_ptr<TensorBase<ScaleBiasT>>>(scVar);
+        auto& biTensor = *std::get<std::unique_ptr<TensorBase<ScaleBiasT>>>(biVar);
+        auto& meTensor = *std::get<std::unique_ptr<TensorBase<MeanVarianceT>>>(meVar);
+        auto& vaTensor = *std::get<std::unique_ptr<TensorBase<MeanVarianceT>>>(vaVar);
+        auto& outTensor = *std::get<std::unique_ptr<TensorBase<InputT>>>(outVar);
+
```
bghimireamd: We could also call `isApplicable` just before `batchnormFwdInference` here in future.

adickin-amd: might even be able to do it before we call into the registry since we likely arnt going to do a ton of tensor checks,

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormRegistry.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <variant>
+
+#include <hipdnn_sdk/test_utilities/cpu_graph_executor/GenericBatchnormExecutor.hpp>
+
+namespace hipdnn_sdk
+{
+namespace test_utilities
+{
+
+inline std::unordered_map<BatchnormSignatureKey, std::unique_ptr<IGenericBatchnormExecutor>>&
```
adickin-amd: next pr: Plan to revise how we construct and store the registries.  Im thinking of moving the registries to a single place so users can go to one place to get a op registry. The the Graph executor would also change to use the new container to get access to each registry.

sdk/include/hipdnn_sdk/test_utilities/FlatbufferDatatypeMapping.hpp
```diff
@@ -0,0 +1,100 @@
+// Copyright © Advanced Micro Devices, Inc., or its affiliates.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <hipdnn_sdk/data_objects/data_types_generated.h>
+#include <hipdnn_sdk/utilities/UtilsBfp16.hpp>
+#include <hipdnn_sdk/utilities/UtilsFp16.hpp>
+
+namespace
```
shumway: I don't think an unnamed namespace in a header file is a common pattern. Unnamed namespaces limit linkage to file-scope with cpp files (similar to the static keyword), but for a header file the "file-scope" is determined by how it's included. You really want to use a project namespace.

sdk/include/hipdnn_sdk/test_utilities/FlatbufferDatatypeMapping.hpp
```diff
@@ -0,0 +1,100 @@
...
+{
+template <typename T>
```
shumway: Please explain the purpose of this struct with a comment. It's not clear to me what this is useful for, it looks like an extra layer of indirection. Literally "TypeWrapper<T> has an alias "type" that is an alias for T." Why do this?

sdk/include/hipdnn_sdk/test_utilities/FlatbufferDatatypeMapping.hpp
```diff
@@ -0,0 +1,100 @@
...
+struct TypeWrapper
+{
+    using type = T;
```
shumway: Do you have a naming convention? If classes have UpperCase names, why would an alias to a type be lower case? I just want to make sure I understand this project's style guide.

sdk/include/hipdnn_sdk/test_utilities/FlatbufferDatatypeMapping.hpp
```diff
@@ -0,0 +1,100 @@
...
+};
+
+template <hipdnn_sdk::data_objects::DataType DT>
+constexpr auto getTypeWrapper()
+{
+    using DataType = hipdnn_sdk::data_objects::DataType;
+
+    if constexpr(DT == DataType::FLOAT)
+    {
+        return TypeWrapper<float>{};
```
shumway: This might be simpler as `return float{};`.

sdk/include/hipdnn_sdk/test_utilities/FlatbufferDatatypeMapping.hpp
```diff
@@ -0,0 +1,100 @@
...
+        return TypeWrapper<hip_bfloat16>{};
+    }
+    else
+    {
+        static_assert(DT != DT, "Unsupported DataType");
+    }
+}
+
+template <typename T>
+constexpr auto toDataType()
```
shumway: It's probably more clear to readers of this code to specify that this always returns an enum value of type hipdnn_sdk::data_objects::DataType. That gives you some extra type safety, too.

sdk/include/hipdnn_sdk/test_utilities/FlatbufferDatatypeMapping.hpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+    else
+    {
+        static_assert(sizeof(T) == 0, "Unsupported native type");
+    }
+}
+
+}
+
+namespace hipdnn_sdk
```
shumway: This is kind of verbose, especially since you put the braces on the next line.

What about:
namespace hipdnn_sdk::test_utilities
{

sdk/include/hipdnn_sdk/test_utilities/FlatbufferDatatypeMapping.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+// Struct for DataType to native type conversion
+template <hipdnn_sdk::data_objects::DataType DT>
+struct DataTypeToNative
+{
+    using type = typename decltype(getTypeWrapper<DT>())::type;
+};
+
+// Struct for native type to DataType conversion
+template <typename T>
```
shumway, line 76: I'm getting more confused here. What are objects of type DataType<float> used for. They don't have any values? What does an instantiated struct give us that the type doesn't?

sdk/include/hipdnn_sdk/test_utilities/FlatbufferDatatypeMapping.hpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+namespace hipdnn_sdk
+{
+namespace test_utilities
+{
+
+// Struct for DataType to native type conversion
+template <hipdnn_sdk::data_objects::DataType DT>
+struct DataTypeToNative
```
shumway: This is repeating the pattern I questions before. Now we can represent "float" as:
* float type
* TypeWrapper<float>::type
* DataTypeToNative<TypeWrapper<float>>::type
It's actually even worse, because none of these types have any concepts or constraints, so I can probably come up all sorts of weird ways to say "type float".

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormRegistry.hpp
```diff
@@ -0,0 +1,50 @@
...
+    batchnormRegistry()
+{
+    static std::unordered_map<BatchnormSignatureKey, std::unique_ptr<IGenericBatchnormExecutor>>
+        _reg;
```
shumway: What's the naming convention here? Are all local variables in functions preffixed with _? I've never seen that convention. Do we also write `for (int _i = 0; _i < _n, ++_i)`?

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormRegistry.hpp
```diff
@@ -0,0 +1,50 @@
...
+    return _reg;
+}
+
+template <typename InputT, typename ScaleBiasT, typename MeanVarianceT>
+struct BatchnormSignature
```
shumway: This is exactly the kind of pattern I'm trying to remove from CK. I don't know why you want to encode information in types. Enumerations have a lot of advantages for this. We can discuss offline.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormRegistry.hpp
BrianHarrisonAMD: I think this is a case where it should be reg, or registry since it's not a private member.

adickin-amd: Yea, we have a known issue with clang tidy not properly flagging on the header only files.  We have a task to look into that but I still miss some things.  Especially something like this which isnt getting called anywhere else.  I also dont like this registry setup so ill likely refactor things once the signatures and type conversion stuff is all decided upon.  I dont want to spend time refactoring something that is still getting lots of comments.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormRegistry.hpp
BrianHarrisonAMD: I think we could potentially change this signature to use the enum data types as constexpr:

```
template<hipdnn_sdk::data_objects::DataType InputDT,
         hipdnn_sdk::data_objects::DataType ScaleBiasDT,
         hipdnn_sdk::data_objects::DataType MeanVarianceDT>
struct BatchnormSignature
{
    static constexpr hipdnn_sdk::data_objects::DataType INPUT_DATA_TYPE = InputDT;
    static constexpr hipdnn_sdk::data_objects::DataType SCALE_BIAS_DATA_TYPE = ScaleBiasDT;
    static constexpr hipdnn_sdk::data_objects::DataType MEAN_VARIANCE_DATA_TYPE = MeanVarianceDT;

...    
```

And then our registry declares which signatures are valid.
We would need some mechanism to construct a BatchnormExecutor with the proper types from this definition, but then it makes it constexpr, and the user just has to add the signature they want generated when adding support for a new executor type combination.

This would unify the BatchnormSignature & BatchnormSignatureKey structs I believe.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/CpuReferenceGraphExecutor.hpp
adickin-amd: This class is currently the simplest executor I could make work to test the registry system.  This PR is just to setup an initial system.  In order to limit the scope of this PR, Ive pushed the refactor of this class to another story.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormRegistry.hpp
adickin-amd: but that still requires us to define the signature as a templated struct doesnt it? 

So at runtime I need to go `auto sig = BatchnormSignature<attr.input, attr2.whatever, attr3.other>()`?

BrianHarrisonAMD: Yea I think you need another runtime key, and the map would be the runtime version for the key.
```
// Registry key: compile-time template
template<hipdnn_sdk::data_objects::DataType InputDT,
         hipdnn_sdk::data_objects::DataType ScaleBiasDT,
         hipdnn_sdk::data_objects::DataType MeanVarianceDT>
struct BatchnormSignatureRegistryKey
{
    static constexpr hipdnn_sdk::data_objects::DataType INPUT_DATA_TYPE = InputDT;
    static constexpr hipdnn_sdk::data_objects::DataType SCALE_BIAS_DATA_TYPE = ScaleBiasDT;
    static constexpr hipdnn_sdk::data_objects::DataType MEAN_VARIANCE_DATA_TYPE = MeanVarianceDT;

    BatchnormSignatureRegistryKey() = default;

    BatchnormSignatureKey toSignatureKey() const
    {
        return BatchnormSignatureKey{
            INPUT_DATA_TYPE,
            SCALE_BIAS_DATA_TYPE,
            MEAN_VARIANCE_DATA_TYPE
        };
    }
};

// Runtime key: can be constructed at runtime
struct BatchnormSignatureKey
{
    hipdnn_sdk::data_objects::DataType inputDataType;
    hipdnn_sdk::data_objects::DataType scaleBiasDataType;
    hipdnn_sdk::data_objects::DataType meanVarianceDataType;
};
```

BrianHarrisonAMD: Okay we looked at this, and I think we were able to get it down to a value struct template arg for the executor, and a single SignatureKey. 

Looks like we should be able to have a single array declared that gives us all the instances we need.

adickin-amd: i think we finally go this to work

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/GenericBatchnormExecutor.hpp
```diff
@@ -0,0 +1,87 @@
...
+        return {NativeToDataType<InputDataType>::value,
+                NativeToDataType<ScaleBiasDataType>::value,
+                NativeToDataType<MeanVarianceDataType>::value};
+    }
+
+    // bool isApplicable(const hipdnn_sdk::data_objects::Node& node) const override
+    // {
+    //     // Implementation to check if this executor can handle the given node
+    //     return true; // Placeholder
+    // }
```
BrianHarrisonAMD, line 47: Right now we have a (not really correct) IsApplicable on a few of the static ref impls.
I think this is where the IsApplicable check will need to live eventually.

That or our ref impls need to be split more, and have IsApplicable on each op (separate one for Bwds, Fwds etc).
We also need to update this to take in the Tensor details I believe, otherwise we can't really properly tell if it's supported.

I don't think that has to be decided here, but something we should think about as we expand the CpuGraphExecutor.

sdk/include/hipdnn_sdk/test_utilities/FlatbufferDatatypeMapping.hpp
jerehartAMD: I didn't realize that you could use a constexpr function to return different types!

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/GenericBatchnormExecutor.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+    //virtual bool isApplicable(const hipdnn_sdk::data_objects::Node& node) const = 0;
+
+    virtual void batchnormFwdInference(std::any& input,
+                                       std::any& scale,
+                                       std::any& bias,
+                                       std::any& mean,
+                                       std::any& variance,
+                                       std::any& output,
+                                       double epsilon)
```
jerehartAMD, line 31: Alright, I'm trying again to understand the use case for the `any` here. Would we expect something besides a TensorBase (of some type) to be passed into these? I noticed that the CPU executor and the GPU executor both operate on `TensorBase<T>` for some T. If the executors all do use TensorBases, is the goal to allow different executors subclasses to support different type combinations (One would only support `TensorBase<float>`, while another might support `TensorBase<half>` too for example), and have a single call site that could call for any hypothetical combination of types, but potentially fail at runtime?

BrianHarrisonAMD: Yea I believe that is correct.
Since we have limited type support that is extended in the Signature list, these are std::any&, and at runtime we will convert them into matching TensorBase<T> to what the graph needs.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/GenericBatchnormExecutor.hpp
adickin-amd: yea the next PR i think ill do the suggestion to build out how we decide if we can/cant execute a graph before we execute a graph.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/GenericBatchnormExecutor.hpp
jerehartAMD: I think that makes sense! How many executors do we expect to have? Is it just going to be one for the CPU and one for the GPU, or will there be different executors for different plugins that have different type supports?

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormRegistry.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+#include <hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistryKey.hpp>
+#include <hipdnn_sdk/test_utilities/cpu_graph_executor/GenericBatchnormExecutor.hpp>
+
+namespace hipdnn_sdk
+{
+namespace test_utilities
+{
+
+constexpr std::array allBatchnormSignatures
```
adickin-amd: probably needs to be all caps.  got to figure out why tidy isnt running on these header files

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/GenericBatchnormExecutor.hpp
BrianHarrisonAMD: I think we are starting with CPU, and eventually will want a reference impl using GPU to improve perf.

I think likely we will only have CPU / GPU  ref impls to start with, but it should be extensible.
We might want to be able to swap in reference impls from other sources over time.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormRegistry.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+constexpr std::array ALL_BATCHNORM_SIGNATURES
+    = {BatchnormSignatureRegistryKey{.inputDataType = hipdnn_sdk::data_objects::DataType::FLOAT,
+                                     .scaleBiasDataType = hipdnn_sdk::data_objects::DataType::FLOAT,
+                                     .meanVarianceDataType
+                                     = hipdnn_sdk::data_objects::DataType::FLOAT},
+       BatchnormSignatureRegistryKey{.inputDataType = hipdnn_sdk::data_objects::DataType::HALF,
+                                     .scaleBiasDataType = hipdnn_sdk::data_objects::DataType::HALF,
+                                     .meanVarianceDataType
+                                     = hipdnn_sdk::data_objects::DataType::HALF}};
```
BrianHarrisonAMD, line 25: Is bfp16 getting added as a follow-up, or just need to add still?

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormRegistry.hpp
```diff
@@ -0,0 +1,77 @@
...
+inline std::unordered_map<BatchnormSignatureRegistryKey,
+                          std::unique_ptr<IGenericBatchnormExecutor>,
+                          BatchnormSignatureRegistryKeyHash>&
+    batchnormRegistry()
+{
+    static std::unordered_map<BatchnormSignatureRegistryKey,
+                              std::unique_ptr<IGenericBatchnormExecutor>,
+                              BatchnormSignatureRegistryKeyHash>
+        registry;
+    return registry; //
```
BrianHarrisonAMD, line 36: Missing, or not needed comment it looks like.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistryKey.hpp
```diff
@@ -0,0 +1,40 @@
...
+    bool operator==(const BatchnormSignatureRegistryKey& other) const
+    {
+        return inputDataType == other.inputDataType && scaleBiasDataType == other.scaleBiasDataType
+               && meanVarianceDataType == other.meanVarianceDataType;
+    }
+
+    //fix these names
+    hipdnn_sdk::data_objects::DataType inputDataType;
+    hipdnn_sdk::data_objects::DataType scaleBiasDataType;
+    hipdnn_sdk::data_objects::DataType meanVarianceDataType;
```
BrianHarrisonAMD, line 26: Still need fixing?
Or comment removal?

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistryKey.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+struct BatchnormSignatureRegistryKeyHash
+{
+    std::size_t operator()(const BatchnormSignatureRegistryKey& k) const
+    {
+        return std::hash<int>()(static_cast<int>(k.inputDataType))
+               ^ (std::hash<int>()(static_cast<int>(k.scaleBiasDataType)) << 1)
+               ^ (std::hash<int>()(static_cast<int>(k.meanVarianceDataType)) << 2);
+    }
+};
```
BrianHarrisonAMD, line 37: Not a huge deal, but might want a few bits per since they have multiple values.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/GenericBatchnormExecutor.hpp
```diff
@@ -0,0 +1,78 @@
...
+{
+namespace test_utilities
+{
+
+class IGenericBatchnormExecutor
+{
+public:
+    virtual ~IGenericBatchnormExecutor() = default;
+
+    //virtual bool isApplicable(const hipdnn_sdk::data_objects::Node& node) const = 0;
```
BrianHarrisonAMD, line 23: I think this is where IsApplicable should be, but we need the tensor details as well.
We can leave this for now, and do in a follow-up.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/TensorVariant.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+namespace hipdnn_sdk
+{
+namespace test_utilities
+{
+
+using TensorVariant = std::variant<std::unique_ptr<hipdnn_sdk::utilities::TensorBase<float>>,
+                                   std::unique_ptr<hipdnn_sdk::utilities::TensorBase<half>>>;
+
+class TensorVariantUtils
```
BrianHarrisonAMD, line 22: I think these could be inlined utility methods instead of in a class.

sdk/tests/test_utilities/cpu_graph_executor/TestCpuReferenceGraphExecutor.cpp
```diff
@@ -0,0 +1,201 @@
...
+private:
+public:
+    static flatbuffers::FlatBufferBuilder createValidBatchnormGraph(
+        std::vector<int64_t> strides = {1, 3, 224, 224},
+        std::vector<int64_t> dims = {1, 3, 224, 224},
+        bool hasOptionalAttributes = true,
+        hipdnn_sdk::data_objects::DataType inputDataType = DataType::FLOAT,
+        hipdnn_sdk::data_objects::DataType scaleBiasDataType = DataType::FLOAT,
+        hipdnn_sdk::data_objects::DataType meanVarianceDataType = DataType::FLOAT)
+    {
```
BrianHarrisonAMD, line 31: I think this would be useful to move into the test_utilities class.
We have similar "createValidXGraphs" I think.

sdk/include/hipdnn_sdk/utilities/ShallowTensor.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <hipdnn_sdk/utilities/ShapeUtilities.hpp>
+#include <hipdnn_sdk/utilities/Tensor.hpp>
+
+namespace hipdnn_sdk
+{
+namespace utilities
+{
+
+template <class T>
+class ShallowTensor : public TensorBase<T>
```
BrianHarrisonAMD, line 21: Might be nice to have a way to construct one of these from another TensorBase<T>.
Can leave for now, but I think eventually we want a quick way to create a clone of variantpacks to use for reference impl tests.

sdk/tests/test_utilities/cpu_graph_executor/TestCpuReferenceGraphExecutor.cpp
```diff
@@ -0,0 +1,201 @@
...
+                                                         static_cast<MeanVarianceType>(0.0f),
+                                                         static_cast<MeanVarianceType>(1.0f),
+                                                         seed));
+
+        PinnedTensor<MeanVarianceType> varianceTensor(derivedDims);
+        deviceBuffers.push_back(generateRandomHostBuffer(varianceTensor,
+                                                         6,
+                                                         static_cast<MeanVarianceType>(0.1f),
+                                                         static_cast<MeanVarianceType>(1.0f),
+                                                         seed));
```
BrianHarrisonAMD, line 137: I don't think these need to be Pinned since it's host only :P.

sdk/include/hipdnn_sdk/utilities/ShallowTensor.hpp
BrianHarrisonAMD: I guess that's a HostTensor, and not a ShallowTensor in that case.

BrianHarrisonAMD: Why are these Shallow?
I think we should make these just HostTensors, and not shallow.
Then we can still allow for the fill versions, and a copy constructor from other TensorBase<T>.

BrianHarrisonAMD: Oh nevermind I see.

Hmm I guess we might want both HostTensor, and ShallowHostTensor eventually.

BrianHarrisonAMD (approved):
LGTM.

There are a few clean-up comments, and a few things that I think we want eventually, but I think we could merge this, and do that other work in follow-ups.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistryKey.hpp
adickin-amd: ah i fixed and left comment

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormRegistry.hpp
adickin-amd: need to add it after we fix the bfloat16 headers.  Theres an issue with implicit casting we need to deal with in the batchnorm fwd inference calc.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/BatchnormSignatureRegistryKey.hpp
adickin-amd: done, 4 and 8 instead

sdk/tests/test_utilities/cpu_graph_executor/TestCpuReferenceGraphExecutor.cpp
adickin-amd: yea I dont like this test much, I want to do a large refactor in the next ticket

sdk/tests/test_utilities/cpu_graph_executor/TestCpuReferenceGraphExecutor.cpp
adickin-amd: next PR, migrate to non pinned

sdk/include/hipdnn_sdk/utilities/ShallowTensor.hpp
adickin-amd: nothing to do here.  Possible add HostTensor in next PR.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/TensorVariant.hpp
adickin-amd: Combine most of these functions into one next PR

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/GenericBatchnormExecutor.hpp
adickin-amd: yes, next PR.

PR #212, opened 2025-09-17
MIOpen convolution forward integration

averinevg:
## Proposed changes

Add integration to the MIOpen plugin for convolution forward.

## Checklist


plugins/miopen_legacy_plugin/engines/plans/MiopenConvPlanBuilder.cpp
```diff
@@ -0,0 +1,316 @@
...
+    // Check convolution attributes
+
+    const auto prePadding = attr->pre_padding();
+    const auto postPadding = attr->post_padding();
+    const auto stride = attr->stride();
+    const auto dilation = attr->dilation();
+
+    const auto spatialDimCount = miopen_utils::getSpatialDimCount(tensorAttrX);
+
+    auto checkVectorSize = [&](const auto* vec, const char* name) {
```
adickin-amd: Can we refactor these lambdas into named private functions.

plugins/miopen_legacy_plugin/engines/plans/MiopenConvPlanBuilder.cpp
```diff
@@ -0,0 +1,316 @@
...
+    // Create MIOpen tensor descriptors
+    const MiopenTensor tensorX(tensorAttrX);
+    const MiopenTensor tensorW(tensorAttrW);
+    const MiopenTensor tensorY(tensorAttrY);
+
+    // Create MIOpen convolution descriptor
+    const MiopenConvDescriptor convDesc(spatialDimCount, *attr);
+
+    size_t solutionCount;
+    THROW_ON_MIOPEN_FAILURE(miopenConvolutionForwardGetSolutionCount(handle.miopenHandle,
```
adickin-amd: if we dont check the above and simply call this function, does MIOpen do any of the above checks?

plugins/miopen_legacy_plugin/engines/plans/MiopenConvFwdPlan.cpp
```diff
@@ -0,0 +1,142 @@
...
+    other._solution = nullptr;
+}
+
+ConvFwdPlan& ConvFwdPlan::operator=(ConvFwdPlan&& other) noexcept
+{
+    if(this != &other)
+    {
+        if(_solution != nullptr)
+        {
+            std::ignore = miopenDestroySolution(_solution);
```
adickin-amd: should we not validate the return?  if we fail to destroy we should at least log we couldnt destroy it.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuConvForward.cpp
```diff
@@ -0,0 +1,361 @@
...
+        {{1, 20, 20, 20}, {1, 1, 3, 3}, {1, 1}, {1, 1}, {1, 1}, {1, 1}},
+        {{1, 20, 20, 20}, {1, 1, 3, 3}, {0, 0}, {0, 0}, {1, 1}, {1, 1}},
+        {{1, 20, 20, 20}, {1, 1, 3, 3}, {1, 1}, {1, 1}, {2, 2}, {1, 1}},
+        {{1, 20, 20, 20}, {1, 1, 3, 3}, {2, 2}, {2, 2}, {1, 1}, {2, 2}},
+    };
+}
+
+} // namespace
+
+TEST_P(IntegrationGpuConvFwdNchwFp32, Correctness)
```
adickin-amd, line 216: would it be possible to refactor the tests such that NCHW/NHWC is a test param and not part of the suite?

plugins/miopen_legacy_plugin/tests/common/TestOperationsCommon.hpp
```diff
@@ -13,6 +14,98 @@ namespace test_operations_common
 
 using namespace hipdnn_sdk::test_utilities;
 
+struct ConvTestCase
```
adickin-amd, line 15: might be worth splitting this into two files.  One for conv, one for batch norm.  This also isnt a test itself so I wonder if something like ConvTestUtils and BatchnormConvUtils.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuConvForward.cpp
```diff
@@ -0,0 +1,361 @@
...
+#include <hipdnn_sdk/utilities/Tensor.hpp>
+
+using namespace hipdnn_frontend;
+using namespace hipdnn_sdk::utilities;
+using namespace hipdnn_sdk::test_utilities;
+
+namespace
+{
+
+struct ConvTestCase
```
adickin-amd: isnt this also defined in TestCommonOperations?

plugins/miopen_legacy_plugin/tests/mocks/MockEngine.hpp
```diff
@@ -18,7 +18,10 @@ class MockEngine : public IEngine
 {
 public:
     MOCK_METHOD(int64_t, id, (), (const, override));
-    MOCK_METHOD(bool, isApplicable, (const hipdnn_plugin::IGraph& opGraph), (const, override));
+    MOCK_METHOD(bool,
+                isApplicable,
+                (HipdnnEnginePluginHandle & handle, const hipdnn_plugin::IGraph& opGraph),
```
adickin-amd, line 23: is formatting not running on this correctly?

plugins/miopen_legacy_plugin/MiopenUtils.cpp
```diff
@@ -44,20 +44,41 @@ miopenDataType_t tensorDataTypeToMiopenDataType(const hipdnn_sdk::data_objects::
...
+{
+    if(attr.dims()->size() < 3)
+    {
+        throw hipdnn_plugin::HipdnnPluginException(
+            HIPDNN_PLUGIN_STATUS_BAD_PARAM,
+            "Tensor must have at least 3 dimensions, but got: "
+                + std::to_string(attr.dims()->size()));
+    }
+
+    return attr.dims()->size() - 2;
```
adickin-amd, line 79: why -2?

tests/frontend/IntegrationConvForward.cpp
```diff
@@ -0,0 +1,316 @@
...
+            ASSERT_EQ(result.code, ErrorCode::OK) << result.err_msg;
+        }
+    }
+
+    void runTest()
+    {
+        const auto& testCase = GetParam();
+
+        std::vector<int64_t> xDims = {2, 3, 14, 14}; // n, c, h, w
+        std::vector<int64_t> wDims = {3, 3, 3, 3}; // k, c, h, w
```
adickin-amd, line 255: k, c, h, w?

plugins/miopen_legacy_plugin/engines/plans/MiopenConvFwdPlan.cpp
```diff
@@ -0,0 +1,142 @@
...
+           miopenTensorArgument_t{
+               .id = miopenTensorConvolutionW, .descriptor = &wDesc, .buffer = wBuffer.ptr},
+           miopenTensorArgument_t{
+               .id = miopenTensorConvolutionY, .descriptor = &yDesc, .buffer = yBuffer.ptr}};
+
+    size_t workspaceSize = 0;
+    if(workspace != nullptr)
+    {
+        // Assume the provided workspace is large enough
+        THROW_ON_MIOPEN_FAILURE(miopenGetSolutionWorkspaceSize(_solution, &workspaceSize));
```
adickin-amd, line 143: Should we amend our EnginePluginApi to pass workspace size to hipdnnEnginePluginExecuteOpGraph. Seems like it would be beneficial to know the size of the void* we got passed is.

plugins/miopen_legacy_plugin/MiopenUtils.cpp
BrianHarrisonAMD: NC aren't part of spatial dims

plugins/miopen_legacy_plugin/engines/plans/MiopenConvFwdPlan.cpp
BrianHarrisonAMD: Hmm I think that might actually not be part of the frontend API.

plugins/miopen_legacy_plugin/engines/plans/MiopenConvPlanBuilder.cpp
```diff
@@ -0,0 +1,316 @@
...
+    size_t workSpaceSize;
+    THROW_ON_MIOPEN_FAILURE(miopenConvolutionForwardGetWorkSpaceSize(handle.miopenHandle,
+                                                                     params.w().tensorDescriptor(),
+                                                                     params.x().tensorDescriptor(),
+                                                                     params.conv().convDescriptor(),
+                                                                     params.y().tensorDescriptor(),
+                                                                     &workSpaceSize));
+
+    return workSpaceSize;
+}
```
BrianHarrisonAMD, line 120: I think this is the best we can do right now, but ill add a task to add an API for getting the workspace from the selected solution at the plan level.

BrianHarrisonAMD: Ah we did have this logged.
Alright we will need to prioritize when we can.

plugins/miopen_legacy_plugin/engines/plans/MiopenConvFwdPlan.cpp
adickin-amd: :(

BrianHarrisonAMD:
Looks like some test breaks in the CI for new conv tests.

tests/frontend/IntegrationConvForward.cpp
BrianHarrisonAMD: It's the filter dims:
[output channels, input channels, height, width]

plugins/miopen_legacy_plugin/MiopenConvDescriptor.cpp
```diff
@@ -0,0 +1,145 @@
...
+namespace
+{
+
+void copyWithCheck(const flatbuffers::Vector<int64_t>* src,
+                   std::vector<int>& dst,
+                   size_t expectedSize,
+                   const char* name,
+                   const char* expectedSizeName)
+{
+    if(src != nullptr)
```
BrianHarrisonAMD: Is this possible?

Might be worth making this a check and exception since I wouldn't expect this should occur.

plugins/miopen_legacy_plugin/MiopenConvDescriptor.cpp
```diff
@@ -0,0 +1,145 @@
...
+                "MiopenConvDescriptor: pre_padding and post_padding sizes must be equal");
+        }
+
+        if(!std::ranges::equal(*attributes.pre_padding(), *attributes.post_padding()))
+        {
+            throw hipdnn_plugin::HipdnnPluginException(
+                HIPDNN_PLUGIN_STATUS_BAD_PARAM,
+                "MiopenConvDescriptor: asymmetric padding is not supported");
+        }
+    }
```
BrianHarrisonAMD: I don't know if these can be null, but checking doesn't hurt I suppose.
Pretty sure they are required attributes let me check.

plugins/miopen_legacy_plugin/engines/plans/MiopenBatchnormBwdPlan.hpp
```diff
@@ -48,15 +53,21 @@ class BatchnormBwdParams
...
+    BatchnormBwdPlan& operator=(BatchnormBwdPlan&&) = default;
 
     void execute(const HipdnnEnginePluginHandle& handle,
                  const hipdnnPluginDeviceBuffer_t* deviceBuffers,
                  uint32_t numDeviceBuffers,
                  void* workspace = nullptr) const override;
 
 private:
-    std::unique_ptr<BatchnormBwdParams> _params;
+    BatchnormBwdParams _params;
```
BrianHarrisonAMD, line 70: Oh I thought we did this already awhile ago.
Surprising!

plugins/miopen_legacy_plugin/engines/plans/MiopenConvFwdPlan.cpp
BrianHarrisonAMD, line 143: I guess we only have maxWorkspaceSize API at this point, but we should leave a todo to cache the workspace size that we provide to the user when requested for the plan, and then we can check here it matches what we reported earlier, or maybe just remove this and assume the earlier workspace is the same.

Ill add a note in the future work to add the plan API for workspace.

plugins/miopen_legacy_plugin/engines/plans/MiopenConvFwdPlan.cpp
```diff
@@ -0,0 +1,142 @@
...
+        problem, miopenTensorConvolutionX, _params.x().tensorDescriptor()));
+    THROW_ON_MIOPEN_FAILURE(miopenSetProblemTensorDescriptor(
+        problem, miopenTensorConvolutionW, _params.w().tensorDescriptor()));
+    THROW_ON_MIOPEN_FAILURE(miopenSetProblemTensorDescriptor(
+        problem, miopenTensorConvolutionY, _params.y().tensorDescriptor()));
+
+    size_t numSolutions;
+    // Requesting only the best solution
+    THROW_ON_MIOPEN_FAILURE(
+        miopenFindSolutions(handle.miopenHandle, problem, nullptr, &_solution, &numSolutions, 1));
```
BrianHarrisonAMD, line 68: I think we also need to call compile at this this stage. 
Once the plan is created, it should be instantly runnable and not require additional work.

plugins/miopen_legacy_plugin/tests/common/TestOperationsCommon.hpp
averinevg: I split this file.

plugins/miopen_legacy_plugin/engines/plans/MiopenConvFwdPlan.cpp
BrianHarrisonAMD: Hmm since it's find2.0, maybe this guarantees it's already compiled at this point?
We should double check, but it does look like it should be precompiled after this from a quick look.

https://github.com/ROCm/MIOpen/blob/06977176afd94476c18d5290f21cb40745bb73a9/src/problem.cpp#L465-L544

BrianHarrisonAMD: If it's not, then we should compile here, and swap to use the immedate API when running.

plugins/miopen_legacy_plugin/engines/plans/MiopenConvFwdPlan.cpp
BrianHarrisonAMD: Actually, I guess this is cached with the solution.
Probably fine as is.

plugins/miopen_legacy_plugin/MiopenConvDescriptor.cpp
BrianHarrisonAMD: Yea I think these should always be set, but I guess maybe empty gets handled as nullptr?

https://github.com/ROCm/hipDNN/blob/develop/sdk/schemas/tensor_attributes.fbs#L41-L49

plugins/miopen_legacy_plugin/engines/plans/MiopenConvPlanBuilder.cpp
BrianHarrisonAMD: I think the MiopenConvDescriptor constructor also duplicates some of these checks. 
We can likely unify these checks into the MiopenConvDescriptor constructor. 
I guess it depends how we want to handle IsApplicable checks, but seems like we can unify these somewhere.

BrianHarrisonAMD: We might need some checks generally to ensure we don't have issues ourself, but I think some of these should be handled by MIOpen.

The valid dim checks for example seems like MIOpen should be handling.
The others for padding matching, and max size ints probably need to be handled by us since that can be problematic when trying to convert to something valid for the MIOpen API.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuBatchnormBackward.cpp
```diff
@@ -236,7 +236,7 @@ class BatchnormBackward : public ::testing::TestWithParam<Batchnorm2dTestCase>
                                              *invVarianceTensorAttr,
                                              graphTensorBundle);
 
-        result = graphObj->execute(_handle, variantPack, _stream);
+        result = graphObj->execute(_handle, variantPack, nullptr);
```
BrianHarrisonAMD, line 263: Why was this swapped?

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuConvForward.cpp
```diff
@@ -0,0 +1,368 @@
...
+        auto cOut = _wDims[N];
+        std::vector<int64_t> outputDims = {n, cOut};
+
+        for(size_t i = 0; i < spatialDims; ++i)
+        {
+            auto paddedInputSize = _xDims[2 + i] + _convPrePadding[i] + _convPostPadding[i];
+            auto effectiveKernelSize = ((_convDilation[i] * (_wDims[2 + i] - 1))) + 1;
+            auto dimOut = ((paddedInputSize - effectiveKernelSize) / _convStride[i]) + 1;
+            outputDims.push_back(dimOut);
+        }
```
BrianHarrisonAMD: You might be able to use the logic inside the frontend graph for inferring output dims. 
I think this is the same logic.

BrianHarrisonAMD: Guess it's a bit clunky to do it that way though. 
Maybe not a huge deal either way.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuConvForward.cpp
BrianHarrisonAMD: Yea I think this is a duplicate.
Maybe just needs to be moved?

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuConvForward.cpp
```diff
@@ -0,0 +1,368 @@
...
+};
+
+template <typename DataType>
+class ConvForward : public ::testing::TestWithParam<ConvTestCase>
+{
+    struct ConvTensorBundle
+    {
+        ConvTensorBundle(const ConvTestCase& testCase,
+                         unsigned int seed = 1,
+                         const TensorLayout& layout = TensorLayout::NCHW)
```
BrianHarrisonAMD, line 35: I think we need to add seed to the testcase, and include it in the stream output.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuConvForward.cpp
```diff
@@ -0,0 +1,368 @@
...
+        convAttrs.set_stride(testCase._convStride);
+        convAttrs.set_dilation(testCase._convDilation);
+
+        auto yTensorAttr = graphObj->conv_fprop(xTensorAttr, wTensorAttr, convAttrs);
+
+        if(!yTensorAttr->has_uid())
+        {
+            yTensorAttr->set_uid(uid++);
+        }
+        yTensorAttr->set_data_type(inputDataType);
```
BrianHarrisonAMD: I think y needs to be set as output.
It defaults virtual.

Also, do we need to set data type here?
I thought it would get defaulted automagically, but maybe we haven't done that work yet.

plugins/miopen_legacy_plugin/engines/plans/MiopenConvPlanBuilder.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+constexpr size_t MIN_SUPPORTED_TENSOR_DIMS = 4;
+constexpr size_t MAX_SUPPORTED_TENSOR_DIMS = 5;
+
+std::string getNodeName(const hipdnn_sdk::data_objects::Node& node)
+{
+    return node.name() != nullptr ? node.name()->str() : "";
+}
+
+bool isApplicableFwd(const HipdnnEnginePluginHandle& handle, const hipdnn_plugin::IGraph& opGraph)
```
BrianHarrisonAMD, line 28: We should add checks on the tensors to ensure that they are all non-virtual.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuConvForward.cpp
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    void runConvTest(DataType tolerance = 1e-4f, const TensorLayout& layout = TensorLayout::NCHW)
+    {
+        const ConvTestCase& testCase = GetParam();
+
+        auto inputDataType = getDataTypeEnumFromType<DataType>();
+
+        unsigned int seed = std::random_device{}();
+        HIPDNN_LOG_INFO("Test is using {} for its random seed", seed);
```
BrianHarrisonAMD: We should move these to the testcase's, and part of the string stream now.
The logging won't show unless you explicitly enable it, and it makes it hard to replicate random breaks.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuConvForward.cpp
```diff
@@ -0,0 +1,368 @@
...
+};
+
+std::vector<ConvTestCase> getConvFwdTestCases()
+{
+    return {
+        {{1, 20, 20, 20}, {1, 20, 3, 3}, {1, 1}, {1, 1}, {1, 1}, {1, 1}},
+        {{1, 20, 20, 20}, {1, 20, 3, 3}, {0, 0}, {0, 0}, {1, 1}, {1, 1}},
+        {{1, 20, 20, 20}, {1, 20, 3, 3}, {1, 1}, {1, 1}, {2, 2}, {1, 1}},
+        {{1, 20, 20, 20}, {1, 20, 3, 3}, {2, 2}, {2, 2}, {1, 1}, {2, 2}},
+    };
```
BrianHarrisonAMD: I think we should have a bit of diversity here in the tests.

Would be good to add:
- 1x1 filter
- grouped conv
- batched conv (N > 1)
- non-square size for filter & image.
- Separate 3D cases.

We will want to keep the cases small and fast, but I think adding the above will give us more confidence the integration is valid for general cases.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuConvForward.cpp
BrianHarrisonAMD: I think we set up batchnorm similar to this.
I think I prefer this way, since we test the same cases for both layouts, but we could also do a cartesian product of inputs to generate it if that's preferrable. The downside is that we lose the layout as part of the test name that way though.

plugins/miopen_legacy_plugin/tests/TestMiopenConvFwdOperations.cpp
```diff
@@ -0,0 +1,249 @@
...
+            hipdnnPluginStatus_t status = hipdnnEnginePluginDestroy(_handle);
+            ASSERT_EQ(status, HIPDNN_PLUGIN_STATUS_SUCCESS);
+        }
+    }
+
+    void runConvFwdGraph(const ConvTestCase& testCase,
+                         hipdnn_sdk::data_objects::DataType inputDataType,
+                         DataType epsilon)
+    {
+        unsigned int seed = std::random_device{}();
```
BrianHarrisonAMD: Seed here as well, but hopefully covered by moving to testcase.

plugins/miopen_legacy_plugin/tests/TestMiopenConvFwdOperations.cpp
```diff
@@ -0,0 +1,249 @@
...
+                         testing::ValuesIn(getConvTestCases()));
+INSTANTIATE_TEST_SUITE_P(,
+                         TestGpuMiopenConvFwdExecuteGraphNhwcFp32,
+                         testing::ValuesIn(getConvTestCases()));
+INSTANTIATE_TEST_SUITE_P(,
+                         TestGpuMiopenConvFwdExecuteGraphNhwcFp16,
+                         testing::ValuesIn(getConvTestCases()));
+INSTANTIATE_TEST_SUITE_P(,
+                         TestGpuMiopenConvFwdExecuteGraphNhwcBfp16,
+                         testing::ValuesIn(getConvTestCases()));
```
BrianHarrisonAMD: I think at the unit testing level we are just looking to ensure it runs without issue.
We should be able to do a different set of tests, or just a single shape. 

Integration tests is where I would expect to be running several shapes with broad coverage for the different ways of running the OP.

plugins/miopen_legacy_plugin/tests/common/BatchnormCommon.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace test_bn_common
+{
+
+struct Batchnorm2dTestCase
+{
+    int64_t n;
+    int64_t c;
+    int64_t h;
+    int64_t w;
```
BrianHarrisonAMD, line 19: Oh this might be conflicting with the seed changes @mousdahl-amd did.
It'll need some coordination.

plugins/miopen_legacy_plugin/tests/common/ConvolutionCommon.hpp
BrianHarrisonAMD, line 1: I guess the other comments should be on this file, and we should remove the other definition from the integration test file.

plugins/miopen_legacy_plugin/tests/engines/plans/TestMiopenConvFwdPlan.cpp
```diff
@@ -0,0 +1,72 @@
...
+    const auto& node = graph.getNode(0);
+    auto* attrs = node.attributes_as_ConvolutionFwdAttributes();
+    ASSERT_NE(attrs, nullptr);
+
+    // Construct params
+    ConvFwdParams params(*attrs, graph.getTensorMap());
+
+    // Create plan
+    ConvFwdPlan(_handle, std::move(params));
+}
```
BrianHarrisonAMD, line 238: We should add some more tests to cover the error cases for constructing a plan. 
Incorrect dim sizes, mismatched padding etc.

sdk/include/hipdnn_sdk/utilities/StringUtil.hpp
```diff
@@ -22,4 +25,20 @@ static void copyMaxSizeWithNullTerminator(char* destination, const char* source,
 #endif
     destination[maxSize - 1] = '\0';
 }
+
+template <typename T>
+void vecToStream(std::ostream& os, const std::vector<T>& vec)
```
BrianHarrisonAMD: I think this should have inline as well.

tests/frontend/IntegrationConvForward.cpp
```diff
@@ -0,0 +1,324 @@
...
+{
+    std::string pluginPath;
+    std::string description;
+    std::string graphName;
+    FailurePoint expectedFailure;
+    bool useManualUids;
+
+    friend std::ostream& operator<<(std::ostream& os, const IntegrationTestCase& tc)
+    {
+        os << "BatchnormTestCase{" << "plugin_path: " << tc.pluginPath
```
BrianHarrisonAMD: BN -> Conv

tests/frontend/IntegrationConvForward.cpp
```diff
@@ -0,0 +1,324 @@
...
+{
+
+enum class FailurePoint
+{
+    NONE, // No failure expected
+    CREATE_EXECUTION_PLAN, // Expect failure at create execution plan
+    EXECUTE // Expect failure at execute
+};
+
+struct IntegrationTestCase
```
BrianHarrisonAMD, line 34: We could move these to a common since I imagine this will be reused for each OP.

tests/frontend/IntegrationConvForward.cpp
BrianHarrisonAMD, line 1: Can probably be a separate task, but I think this is a good candidate to make into a reusable base with just the op specific set-up needing to be added.

BrianHarrisonAMD (commented):
Thanks for all this!
This task ended up being a lot more than I think we expected. 

Lets discuss to see how we should split these up more going forward.
Seems like the frontend integration testing could probably be done separately, and maybe makes more sense to do it with the frontend work.

plugins/miopen_legacy_plugin/engines/plans/MiopenBatchnormBwdPlan.hpp
SamuelReeder: I only changed the tensors but should have caught those too.

plugins/miopen_legacy_plugin/tests/common/BatchnormCommon.hpp
averinevg: I have fixed that.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuConvForward.cpp
averinevg: I removed duplicated code.

plugins/miopen_legacy_plugin/tests/mocks/MockEngine.hpp
averinevg: Is there something wrong with the formatting?

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuBatchnormBackward.cpp
averinevg: Workspace buffer should be passed instead of the stream.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuConvForward.cpp
averinevg: I've moved it.

plugins/miopen_legacy_plugin/tests/TestMiopenConvFwdOperations.cpp
averinevg: Done

plugins/miopen_legacy_plugin/engines/plans/MiopenConvFwdPlan.cpp
averinevg: Yes, I added a check.

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuConvForward.cpp
averinevg: Added this to the testcase.

plugins/miopen_legacy_plugin/tests/common/ConvolutionCommon.hpp
averinevg: Done

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuBatchnormBackward.cpp
BrianHarrisonAMD: Ah okay just an old bug!

plugins/miopen_legacy_plugin/engines/plans/MiopenConvFwdPlan.cpp
averinevg: Since benchmarking is happening there, the necessary kernels are compiled and cached.

tests/frontend/IntegrationConvForward.cpp
BrianHarrisonAMD: This is a new task now. No worries.

plugins/miopen_legacy_plugin/tests/common/ConvolutionCommon.hpp
```diff
@@ -0,0 +1,143 @@
...
+        // Dilation = 2
+        {{1, 16, 16, 16}, {1, 16, 3, 3}, {2, 2}, {2, 2}, {1, 1}, {2, 2}, seed},
+        {{1, 16, 16, 16, 16}, {1, 16, 3, 3, 3}, {2, 2, 2}, {2, 2, 2}, {1, 1, 1}, {2, 2, 2}, seed},
+        // Batched convolution
+        {{8, 16, 16, 16}, {1, 16, 1, 1}, {0, 0}, {0, 0}, {1, 1}, {1, 1}, seed},
+        {{8, 16, 16, 16, 16}, {1, 16, 1, 1, 1}, {0, 0, 0}, {0, 0, 0}, {1, 1, 1}, {1, 1, 1}, seed},
+        // Non-square
+        {{1, 16, 16, 8}, {1, 16, 3, 3}, {1, 1}, {1, 1}, {1, 1}, {1, 1}, seed},
+        {{1, 16, 16, 8, 4}, {1, 16, 3, 3, 3}, {1, 1, 1}, {1, 1, 1}, {1, 1, 1}, {1, 1, 1}, seed},
+    };
```
BrianHarrisonAMD, line 156: Ty!

BrianHarrisonAMD (approved):
LGTM thanks for all the updates!

Seems like some test breaks need to be resolved, but my initial comments are all resolved now.

PR #218, opened 2025-09-18
[ALMIOPEN-280] Test output improvements

mousdahl-amd:
## Proposed changes
- Added seed to test case definitions
- Did a sweep to make sure that any other places that were using `fillWithRandomValues()` or `std::random_device{}` were getting annotated appropriately
- Checked the samples, but decided to leave them alone
- Added logging to `CpuFpReferenceMiopenRmsValidation`, as the `CpuFpReferenceValidation` already had decent logging

## Checklist


sdk/include/hipdnn_sdk/test_utilities/CpuFpReferenceMiopenRmsValidation.hpp
```diff
@@ -82,6 +82,13 @@ class CpuFpReferenceMiopenRmsValidation : public IReferenceValidation<T>
         double relativeRmsError = std::sqrt(squareDifference)
                                   / (std::sqrt(static_cast<double>(elementCount)) * maxMagnitude);
 
+        if(relativeRmsError > _relativeTolerance)
+        {
+            HIPDNN_LOG_ERROR("Validation failed: relative rms error = {}, relative tolerance = {}",
+                             relativeRmsError,
+                             _relativeTolerance);
```
SamuelReeder, line 89: A tangent, but it takes some magic to get logging to work from the sdk since it is an `INTERFACE`. I will probably make it have dedicated macros like the FE.

SamuelReeder: It looks like more logs are being put in the SDK, so this will be a priority.

SamuelReeder (approved):
LGTM. I will update 3D cases to align.

bghimireamd:
LGTM

bghimireamd (approved):

adickin-amd (approved):

sdk/include/hipdnn_sdk/test_utilities/CpuFpReferenceMiopenRmsValidation.hpp
BrianHarrisonAMD: I think this gets called only in test binaries, and they all should have logging initalized.
It should hopefully be working fine, but an improvement could be turning on error logging by default for all test binaries.

BrianHarrisonAMD (approved):
These changes LGTM!

Was there no cases where we had a one off test that was using randomly generated data? 
Looks like everything was handled with test case output for errors, but thought there might be a couple one off cases as well.

sdk/include/hipdnn_sdk/test_utilities/CpuFpReferenceMiopenRmsValidation.hpp
SamuelReeder: Yeah I think all tests binaries initialize logging, so as long as `COMPONENT_NAME` is defined somewhere it will work. I think it's either defined for tests or the FE will propagate it as well.

But non-testing consumers will have to initialize still.

mousdahl-amd: I'd like that.  So far the test executables do it when they need it.

PR #219, opened 2025-09-18
[PointWise] Implement Pointwise cpu reference op for Add & Sub

bghimireamd:
## Proposed changes

Ref Pointwise with Add and Sub

## Checklist

### DIFF
--- a/sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuDeviceExecutor.hpp
+++ b/sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuDeviceExecutor.hpp
@@ -16,10 +16,64 @@ namespace test_utilities
 using namespace hipdnn_sdk::utilities;
 
 template <class DataType>
-class CpuExecutionPolicy
+class CpuDeviceExecutor
 {
 public:
-    // Broadcasting utility functions
+    template <typename Op>
+    void executeBinaryBroadcast(const std::vector<const TensorBase<DataType>*>& inputs,
+                                TensorBase<DataType>& output,
+                                Op op)
+    {
+        if(inputs.size() != 2)
+        {
+            throw std::runtime_error("Binary operation requires exactly 2 input tensors.");
+        }
+
+        for(size_t i = 0; i < inputs.size(); ++i)
+        {
+            if(inputs[i] == nullptr)
+            {
+                throw std::runtime_error("Input tensor " + std::to_string(i) + " is null.");
+            }
+        }
+
+        const auto& input1 = *inputs[0];
+        const auto& input2 = *inputs[1];
+
+        std::vector<std::vector<int64_t>> input_shapes = {
+            input1.dims(), input2.dims()
+        };
+
+        auto broadcast_shape = computeBroadcastShape(input_shapes);
+
+        if(output.dims() != broadcast_shape)
+        {
+            throw std::runtime_error("Output shape doesn't match computed broadcast shape");
+        }
+
+        auto func = [&](const std::vector<int64_t>& indices) {
+            auto input1_idx = getBroadcastableIndex(indices, input1.dims());
+            auto input2_idx = getBroadcastableIndex(indices, input2.dims());
+
+            auto input1Val = input1.getHostValue(input1_idx);
+            auto input2Val = input2.getHostValue(input2_idx);
+
+            DataType outputVal;
+            op(outputVal, input1Val, input2Val);
+            output.setHostValue(outputVal, indices);
+        };
+
+        auto parallelFunc = makeParallelTensorFunctor(func, broadcast_shape);
+        parallelFunc();
+    }
+
+
+    void markOutputModified(TensorBase<DataType>& output)
+    {
+        output.memory().markHostModified();
+    }
+private: 
+
     static std::vector<int64_t> computeBroadcastShape(
         const std::vector<std::vector<int64_t>>& input_shapes)
     {
@@ -28,7 +82,6 @@ public:
             throw std::runtime_error("Cannot compute broadcast shape for empty input shapes");
         }
 
-        // Find the maximum number of dimensions
         size_t max_dims = 0;
         for(const auto& shape : input_shapes)
         {
@@ -37,14 +90,12 @@ public:
 
         std::vector<int64_t> broadcast_shape(max_dims, 1);
 
-        // Process each dimension from right to left (NumPy broadcasting rules)
         for(size_t dim = 0; dim < max_dims; ++dim)
         {
             int64_t max_size = 1;
             
             for(const auto& shape : input_shapes)
             {
-                // Get dimension size (1 if dimension doesn't exist)
                 int64_t dim_size = 1;
                 if(dim < shape.size())
                 {
@@ -52,7 +103,6 @@ public:
                     dim_size = shape[shape_dim_idx];
                 }
 
-                // Check broadcasting compatibility
                 if(max_size == 1)
                 {
                     max_size = dim_size;
@@ -82,75 +132,17 @@ public:
             throw std::runtime_error("Output index has fewer dimensions than tensor");
         }
 
-        // Return index vector with same size as tensor dimensions
         std::vector<int64_t> broadcasted_index(tensor_dims.size());
         
-        // Handle dimension alignment (right-aligned)
         size_t dim_offset = output_index.size() - tensor_dims.size();
         
         for(size_t i = 0; i < tensor_dims.size(); ++i)
         {
             size_t output_dim_idx = dim_offset + i;
-            // If dimension is 1, always broadcast to index 0
             broadcasted_index[i] = (tensor_dims[i] == 1) ? 0 : output_index[output_dim_idx];
         }
 
         return broadcasted_index;
-    }
-
-    template <typename Op>
-    void executeBinaryBroadcast(const std::vector<const TensorBase<DataType>*>& inputs,
-                                TensorBase<DataType>& output,
-                                Op op)
-    {
-        if(inputs.size() != 2)
-        {
-            throw std::runtime_error("Binary operation requires exactly 2 input tensors.");
-        }
-
-        const auto& input1 = *inputs[0];
-        const auto& input2 = *inputs[1];
-
-        // Get input shapes
-        std::vector<std::vector<int64_t>> input_shapes = {
-            input1.dims(), input2.dims()
-        };
-
-        // Compute broadcast output shape
-        auto broadcast_shape = computeBroadcastShape(input_shapes);
-
-        // Verify output matches broadcast shape
-        if(output.dims() != broadcast_shape)
-        {
-            throw std::runtime_error("Output shape doesn't match computed broadcast shape");
-        }
-
-        // Iterate over output shape
-        auto func = [&](const std::vector<std::size_t>& indices) {
-            std::vector<int64_t> output_idx(indices.begin(), indices.end());
-
-            // Get broadcasted indices for each input
-            auto input1_idx = getBroadcastableIndex(output_idx, input1.dims());
-            auto input2_idx = getBroadcastableIndex(output_idx, input2.dims());
-
-            // Get values and compute
-            auto input1Val = input1.getHostValue(input1_idx);
-            auto input2Val = input2.getHostValue(input2_idx);
-
-            DataType outputVal;
-            op(outputVal, input1Val, input2Val);
-            output.setHostValue(outputVal, output_idx);
-        };
-
-        // Use dynamic parallel functor that supports any number of dimensions
-        auto parallelFunc = makeParallelTensorFunctor(func, broadcast_shape);
-        parallelFunc();
-    }
-
-
-    void markOutputModified(TensorBase<DataType>& output)
-    {
-        output.memory().markHostModified();
     }
 };
 
--- a/sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuDeviceExecutor.hpp
+++ b/sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuDeviceExecutor.hpp
@@ -40,9 +40,7 @@ public:
         const auto& input1 = *inputs[0];
         const auto& input2 = *inputs[1];
 
-        std::vector<std::vector<int64_t>> input_shapes = {
-            input1.dims(), input2.dims()
-        };
+        std::vector<std::vector<int64_t>> input_shapes = {input1.dims(), input2.dims()};
 
         auto broadcast_shape = computeBroadcastShape(input_shapes);
 
@@ -67,15 +65,14 @@ public:
         parallelFunc();
     }
 
-
     void markOutputModified(TensorBase<DataType>& output)
     {
         output.memory().markHostModified();
     }
-private: 
 
-    static std::vector<int64_t> computeBroadcastShape(
-        const std::vector<std::vector<int64_t>>& input_shapes)
+private:
+    static std::vector<int64_t>
+        computeBroadcastShape(const std::vector<std::vector<int64_t>>& input_shapes)
     {
         if(input_shapes.empty())
         {
@@ -93,7 +90,7 @@ private:
         for(size_t dim = 0; dim < max_dims; ++dim)
         {
             int64_t max_size = 1;
-            
+
             for(const auto& shape : input_shapes)
             {
                 int64_t dim_size = 1;
@@ -123,9 +120,8 @@ private:
         return broadcast_shape;
     }
 
-    static std::vector<int64_t> getBroadcastableIndex(
-        const std::vector<int64_t>& output_index,
-        const std::vector<int64_t>& tensor_dims)
+    static std::vector<int64_t> getBroadcastableIndex(const std::vector<int64_t>& output_index,
+                                                      const std::vector<int64_t>& tensor_dims)
     {
         if(output_index.size() < tensor_dims.size())
         {
@@ -133,9 +129,9 @@ private:
         }
 
         std::vector<int64_t> broadcasted_index(tensor_dims.size());
-        
+
         size_t dim_offset = output_index.size() - tensor_dims.size();
-        
+
         for(size_t i = 0; i < tensor_dims.size(); ++i)
         {
             size_t output_dim_idx = dim_offset + i;
--- a/sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuDeviceExecutor.hpp
+++ b/sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuDeviceExecutor.hpp
@@ -56,9 +56,7 @@ public:
             auto input1Val = input1.getHostValue(input1_idx);
             auto input2Val = input2.getHostValue(input2_idx);
 
-            DataType outputVal;
-            op(outputVal, input1Val, input2Val);
-            output.setHostValue(outputVal, indices);
+            output.setHostValue(static_cast<DataType>(op(input1Val, input2Val)), indices);
         };
 
         auto parallelFunc = makeParallelTensorFunctor(func, broadcast_shape);
--- a/sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuExecutionPolicy.hpp
+++ b/sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuExecutionPolicy.hpp
@@ -20,40 +20,6 @@ class CpuExecutionPolicy
 {
 public:
     template <typename Op>
-    void executeUnary(const std::vector<const TensorBase<DataType>*>& inputs,
-                      TensorBase<DataType>& output,
-                      Op op)
-    {
-        if(inputs.size() != 1)
-        {
-            throw std::runtime_error("Unary operation requires exactly 1 input tensor.");
-        }
-
-        const auto& input = *inputs[0];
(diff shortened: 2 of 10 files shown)


sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuExecutionPolicy.hpp
```diff
@@ -0,0 +1,65 @@
...
+            DataType outputVal;
+            op(outputVal, input1Val, input2Val);
+            output.setHostValue(outputVal, idxVec);
+        };
+
+        if(dims.size() != 4)
+        {
+            throw std::runtime_error("Only 4D tensors are supported, got "
+                                     + std::to_string(dims.size()) + "D");
+        }
```
SamuelReeder: Can we make this support Nd tensors?

SamuelReeder: Can look at `iterateChannelElements` in the batchnorm reference but do it over all elements, but I think there's other ways.

sdk/tests/test_utilities/TestCpuReferencePointwise.cpp
```diff
@@ -0,0 +1,322 @@
...
+        std::vector<const TensorBase<T>*> inputs = {&input1, &input2};
+        CpuReferencePointwiseImpl<T>::pointwiseForward(inputs, output, PointwiseMode::ADD);
+
+        T expected = static_cast<T>(0.375); // 3/8 - exactly representable
+        EXPECT_EQ(output.getHostValue(0, 0, 0, 0), expected);
+    }
+};
+
+using TestTypes = ::testing::Types<float, double, half, hip_bfloat16>;
+TYPED_TEST_SUITE(CpuReferencePointwiseTemplate, TestTypes);
```
SamuelReeder, line 585: We should try to use this pattern more!

sdk/include/hipdnn_sdk/test_utilities/pointwise/PointwiseOperationFunctors.hpp
```diff
@@ -0,0 +1,33 @@
...
+};
+
+struct Subtract
+{
+    template <typename Y, typename X0, typename X1>
+    void operator()(Y& y, const X0& x0, const X1& x1) const
+    {
+        y = static_cast<Y>(x0 - x1);
+    }
+};
```
SamuelReeder, line 29: Is this file for all types of operations? (ie. all of unary, binary, ternary)

sdk/include/hipdnn_sdk/test_utilities/pointwise/ReferencePointwiseBase.hpp
```diff
@@ -0,0 +1,93 @@
...
+    static bool isApplicable(const hipdnn_sdk::data_objects::Node& node)
+    {
+        using namespace hipdnn_sdk::data_objects;
+
+        if(node.attributes_type() != NodeAttributes::PointwiseAttributes)
+        {
+            return false;
+        }
+
+        return true;
```
SamuelReeder, line 44: Could be less verbose: `return node.attributes_type() == NodeAttributes::PointwiseAttributes;`

sdk/include/hipdnn_sdk/test_utilities/pointwise/ReferencePointwiseBase.hpp
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+private:
+    static void validateTensorDimensions(const std::vector<const TensorBase<DataType>*>& inputs,
+                                         const TensorBase<DataType>& output)
+    {
+        if(inputs.empty())
+        {
+            throw std::runtime_error("No input tensors provided.");
+        }
```
SamuelReeder: Since this is a private helper you can maybe remove this check since it's already in `pointwiseForward`.

SamuelReeder: Actually, I'd say remove the one in `pointwiseForward` instead so it's part of validation.

sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuExecutionPolicy.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+namespace hipdnn_sdk
+{
+namespace test_utilities
+{
+
+using namespace hipdnn_sdk::utilities;
+
+template <class DataType>
+class CpuExecutionPolicy
```
BrianHarrisonAMD: Why is it called CpuExecutionPolicy?

sdk/include/hipdnn_sdk/test_utilities/pointwise/ReferencePointwiseBase.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+template <class DataType, class ExecutionPolicy>
+class ReferencePointwiseBase
+{
+public:
+    static bool isApplicable(const hipdnn_sdk::data_objects::Node& node)
+    {
+        using namespace hipdnn_sdk::data_objects;
+
+        if(node.attributes_type() != NodeAttributes::PointwiseAttributes)
```
BrianHarrisonAMD, line 27: We either need to update these checks to include the tensor details as well, or we need to support N-dimensional ops.

Additionally, we should be rejecting nodes for pointwise ops we don't have support for yet.

sdk/include/hipdnn_sdk/test_utilities/pointwise/ReferencePointwiseBase.hpp
```diff
@@ -0,0 +1,93 @@
...
+            break;
+        case hipdnn_sdk::data_objects::PointwiseMode::SUB:
+            policy.executeBinary(inputs, output, pointwise::Subtract{});
+            break;
+        default:
+            throw std::runtime_error("Unsupported pointwise operation: "
+                                     + std::to_string(static_cast<int>(operation)));
+        }
+
+        policy.markOutputModified(output);
```
BrianHarrisonAMD: I think policy here is not quite right.
Or at least, I am used to execution policy referring to things like seq / parallel execution.

In this case, it seems like ExecutionPolicy is to handle running generic elementwise ops.

sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuExecutionPolicy.hpp
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        const auto& input1 = *inputs[0];
+        const auto& input2 = *inputs[1];
+        const auto& dims = output.dims();
+
+        auto func = [&](const auto&... indices) {
+            std::vector<int64_t> idxVec = {static_cast<int64_t>(indices)...};
+            auto input1Val = input1.getHostValue(idxVec);
+            auto input2Val = input2.getHostValue(idxVec);
```
BrianHarrisonAMD: I added a version of tensor accessors that use the array instead of ...
I think we could just take in dims as a std::vector<int64_t>, and then use that directly and avoid some extra manipulation here.

sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuExecutionPolicy.hpp
```diff
@@ -0,0 +1,65 @@
...
+            DataType outputVal;
+            op(outputVal, input1Val, input2Val);
```
BrianHarrisonAMD: I think we could change these to be OP's that return a DataType instead of take in a DataType&.
I guess this is nice since you get automatic type args for the op, but if it returned DataType then you could make it a oneliner. 

I guess either way works.

sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuExecutionPolicy.hpp
BrianHarrisonAMD: Oh hmm, I guess ParallelTensorFunctor might be only working off ...
I guess it depends how we want that to work.

sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuExecutionPolicy.hpp
BrianHarrisonAMD: Looks like ParallelTensorFunctor works off ..., and that's why it's limited to 4D currently (the rest is Nd).

We should add a makeParallelTensorFunctor that can split off the dims for any length to support Nd.

BrianHarrisonAMD: I think we could swap it to something like:

```
// Dynamic dimension version of ParallelTensorFunctor
template <typename F>
struct ParallelTensorFunctorDynamic
{
    F _func;
    std::vector<std::size_t> _lengths;
    std::vector<std::size_t> _strides;
    std::size_t _totalElements;

    ParallelTensorFunctorDynamic(F f, const std::vector<int64_t>& dimensions)
        : _func(f)
        , _lengths(dimensions.begin(), dimensions.end())
        , _strides(dimensions.size())
        , _totalElements(1)
    {
        if(_lengths.empty())
        {
            _totalElements = 0;
            return;
        }

        // Calculate strides in the same way as the static version
        _strides.back() = 1;
        for(int i = static_cast<int>(_lengths.size()) - 2; i >= 0; --i)
        {
            _strides[i] = _strides[i + 1] * _lengths[i + 1];
        }
        _totalElements = _strides[0] * _lengths[0];
    }

    std::vector<std::size_t> getNdIndices(std::size_t i) const
    {
        std::vector<std::size_t> indices(_lengths.size());

        for(std::size_t idim = 0; idim < _lengths.size(); ++idim)
        {
            indices[idim] = i / _strides[idim];
            i -= indices[idim] * _strides[idim];
        }

        return indices;
    }

    void operator()(std::size_t numThreads = 1) const
    {
        if(numThreads == 0 || _totalElements == 0)
        {
            return;
        }

        std::size_t workPerThread = (_totalElements + numThreads - 1) / numThreads;

        std::vector<JoinableThread> threads(numThreads);

        for(std::size_t threadIdx = 0; threadIdx < numThreads; ++threadIdx)
        {
            std::size_t workBegin = threadIdx * workPerThread;
            std::size_t workEnd = std::min((threadIdx + 1) * workPerThread, _totalElements);

            auto threadFunc = [=, *this] {
                for(std::size_t workIdx = workBegin; workIdx < workEnd; ++workIdx)
                {
                    _func(getNdIndices(workIdx));
                }
            };
            threads[threadIdx] = JoinableThread(threadFunc);
        }
    }
};

// Factory function for dynamic dimension version
template <typename F>
static auto makeParallelTensorFunctor(F f, const std::vector<int64_t>& dimensions)
{
    return ParallelTensorFunctorDynamic<F>(f, dimensions);
}
```

Not sure we need to be using ..., and force NDIM to be constexpr, but I might be missing a reason why.

BrianHarrisonAMD: Oh, and the current version is wrong for strides.
We actually need to update it or else it's only going to work for NCHW.

Edit:
Oh maybe not, since strides for ParallelTensorFunctor are only used for the index calculations, and it uses the actual tensor strides instead when accessing values.

sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuExecutionPolicy.hpp
BrianHarrisonAMD: Okay see comment below, I think we can make ParallelTensorFunctor work off dims only, and then this just takes in std::vector<int64_t> to represent the indices.

sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuExecutionPolicy.hpp
BrianHarrisonAMD: Maybe the intent is that this is representing CpuParallelExecution, and we could add alternatives for like  GpuExecution or, CpuSequentialExecution?

sdk/tests/test_utilities/TestCpuReferencePointwise.cpp
```diff
@@ -0,0 +1,322 @@
...
+template <typename T>
+class CpuReferencePointwiseTemplate : public ::testing::Test
+{
+protected:
+    T getTolerance() const
+    {
+        if constexpr(std::is_same_v<T, half> || std::is_same_v<T, hip_bfloat16>)
+        {
+            return static_cast<T>(1e-2f);
+        }
```
BrianHarrisonAMD: I would think half could probably have a more strict tolerance.
But it depends on the data range being tested.

sdk/tests/test_utilities/TestCpuReferencePointwise.cpp
```diff
@@ -0,0 +1,322 @@
...
+        input2.fillWithValue(static_cast<T>(2.0));
+
+        std::vector<const TensorBase<T>*> inputs = {&input1, &input2};
+
+        CpuReferencePointwiseImpl<T>::pointwiseForward(inputs, output, PointwiseMode::ADD);
+
+        T expected = static_cast<T>(3.0);
+        EXPECT_EQ(output.getHostValue(0, 0, 0, 0), expected);
+        EXPECT_EQ(output.getHostValue(0, 1, 1, 1), expected);
+        EXPECT_EQ(output.getHostValue(0, 2, 0, 1), expected);
```
BrianHarrisonAMD: These should probably do a full check instead of a few samples. 
Should be able to use AllClose with a tensor init'd to 3 in this case.

sdk/tests/test_utilities/TestCpuReferencePointwise.cpp
```diff
@@ -0,0 +1,322 @@
...
+        input2.fillWithValue(static_cast<T>(2.0));
+
+        std::vector<const TensorBase<T>*> inputs = {&input1, &input2};
+
+        CpuReferencePointwiseImpl<T>::pointwiseForward(inputs, output, PointwiseMode::SUB);
+
+        T expected = static_cast<T>(3.0);
+        EXPECT_EQ(output.getHostValue(0, 0, 0, 0), expected);
+        EXPECT_EQ(output.getHostValue(0, 1, 1, 1), expected);
+        EXPECT_EQ(output.getHostValue(0, 2, 0, 1), expected);
```
BrianHarrisonAMD: Same here swap to AllClose.

sdk/tests/test_utilities/TestCpuReferencePointwise.cpp
```diff
@@ -0,0 +1,322 @@
...
+                    static_cast<float>(tolerance));
+        EXPECT_NEAR(static_cast<float>(output.getHostValue(0, 0, 0, 1)),
+                    2.2f,
+                    static_cast<float>(tolerance));
+        EXPECT_NEAR(static_cast<float>(output.getHostValue(0, 0, 1, 0)),
+                    3.3f,
+                    static_cast<float>(tolerance));
+        EXPECT_NEAR(static_cast<float>(output.getHostValue(0, 0, 1, 1)),
+                    4.4f,
+                    static_cast<float>(tolerance));
```
BrianHarrisonAMD: This is fine, but I also think all close would be nice here.

BrianHarrisonAMD: There is a few more in this file, but I think just globally we should probably use AllClose for these, and reduce tolerance since these are small numbers.

sdk/tests/test_utilities/TestCpuReferencePointwise.cpp
BrianHarrisonAMD: Yea this is nice!

sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuExecutionPolicy.hpp
BrianHarrisonAMD, line 1: I think this will only work if both inputs have matching dims, but I think generally pointwise should support broadcastable dims. 

I.E

If you try this with dims p1 = {2, 2, 2, 2}  p2 = {1, 2, 1, 1}, then the current version won't work, but we should support that since it's broad castable.

That's what Bias would look like for Conv + Bias + Activ I believe.

BrianHarrisonAMD: If we modified this to get the output dims, and then operate on those, I think that would work.

We would have to have a method that would convert the current Index into the broadcasted version.

For an example:

P1 = {1, 2, 2, 1}
P2 = {2, 1, 1, 2}
Output = {2, 2, 2, 2}

We iterate over {2, 2 ,2 ,2}

For converting, we would need to take each dim, and recalc:
```
inline std::vector<int64_t> getBroadcastableIndex(const std::vector<int64_t>& index,
                                                  const std::vector<int64_t>& dims)
{
    if(index.size() != dims.size())
    {
        throw std::invalid_argument("Index and dims must have the same size");
    }

    std::vector<int64_t> broadcasted(index.size());
    for(std::size_t i = 0; i < index.size(); ++i)
    {
        // If dimension is 1, always broadcast to 1
        broadcasted[i] = (dims[i] == 1) ? 1 : index[i];
    }
    return broadcasted;
}
```

BrianHarrisonAMD: Probably also need to convert non-matching dims into a version that has 1's prepended.
Some of this logic is in the PointwiseNode checks.

BrianHarrisonAMD: The above example is generic, but it makes more sense if you think of bias:

P1 = {N, K, H, W}
P2 = {1, K, 1, 1}
Output = {N, K, H, W}

P2 only cares about K index when doing the add.

sdk/include/hipdnn_sdk/test_utilities/pointwise/PointwiseOperationFunctors.hpp
bghimireamd: I was planning to place other (unary, binary, ternary) in there too. Should I have sep file or namespace for them?

sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuExecutionPolicy.hpp
bghimireamd: yes in future there could be `GpuExecutionPolicy`

sdk/include/hipdnn_sdk/test_utilities/pointwise/ReferencePointwiseBase.hpp
bghimireamd: I could do N-dimensional ops.

sdk/include/hipdnn_sdk/test_utilities/pointwise/ReferencePointwiseBase.hpp
bghimireamd: I will rename to `DeviceExecutor` to handles CPU/GPU specifics

sdk/include/hipdnn_sdk/test_utilities/pointwise/PointwiseOperationFunctors.hpp
SamuelReeder: It's probably fine. Could consider a separate file in the future if there's a lot.

sdk/include/hipdnn_sdk/test_utilities/pointwise/ReferencePointwiseBase.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+        if(attrs == nullptr)
+        {
+            return false;
+        }
+
+        if(attrs->in_0_tensor_uid() == 0 || attrs->out_0_tensor_uid() == 0)
+        {
+            return false;
+        }
```
BrianHarrisonAMD, line 88: Probably don't need to worry about id checks at this spot.

sdk/include/hipdnn_sdk/test_utilities/pointwise/CpuDeviceExecutor.hpp
```diff
@@ -0,0 +1,144 @@
...
+        const auto& input2 = *inputs[1];
+
+        std::vector<std::vector<int64_t>> input_shapes = {input1.dims(), input2.dims()};
+
+        auto broadcast_shape = computeBroadcastShape(input_shapes);
+
+        if(output.dims() != broadcast_shape)
+        {
+            throw std::runtime_error("Output shape doesn't match computed broadcast shape");
+        }
```
BrianHarrisonAMD, line 50: I think this could be replaced by using the existing areDimensionsBroadcastCompatible inside ShapeUtilities with both inputs, and then if they both pass we can use output dims for the broadcast shape.

Similar to how PointwiseNode [currently checks](https://github.com/ROCm/hipDNN/blob/develop/frontend/include/hipdnn_frontend/node/PointwiseNode.hpp#L160-L190) that it's a valid op.

ammallya:
Imported to ROCm/rocm-libraries

PR #220, opened 2025-09-18
Json conversions of graphs

jerehartAMD:
## Proposed changes

Provides json serialization functions for graphs and their required types.

Currently unhandled: serialization of the actual tensors. 

Example of a serialized graph:
``` json
{
    "compute_type": "float",
    "intermediate_type": "half",
    "io_type": "bfloat16",
    "name": "test",
    "nodes":
    [
        {
            "inputs":
            {
                "bias": 4,
                "inv_variance": 6,
                "mean": 5,
                "scale": 3,
                "x": 1
            },
            "name": "batchnorm",
            "outputs":
            {
                "y": 2
            },
            "type": "BatchnormInferenceAttributes"
        }
    ],
    "tensors":
    [
        {
            "data_type": "float",
            "dims": [1, 3, 224, 224],
            "name": "x",
            "strides": [ 1, 3, 224, 224],
            "uid": 1,
            "virtual": false
        },
        {
            "data_type": "float",
            "dims": [ 1, 3, 224, 224],
            "name": "y",
            "strides": [ 1, 3, 224, 224],
            "uid": 2,
            "virtual": false
        },
        {
            "data_type": "float",
            "dims": [ 1, 3, 1, 1 ],
            "name": "scale",
            "strides": [ 1, 3, 1, 1 ],
            "uid": 3,
            "virtual": false
        },
        {
            "data_type": "float",
            "dims": [ 1, 3, 1, 1 ],
            "name": "bias",
            "strides": [ 1, 3, 1, 1 ],
            "uid": 4,
            "virtual": false
        },
        {
            "data_type": "float",
            "dims": [ 1, 3, 1, 1 ],
            "name": "est_mean",
            "strides": [ 1, 3, 1, 1 ],
            "uid": 5,
            "virtual": false
        },
        {
            "data_type": "float",
            "dims": [ 1, 3, 1, 1 ],
            "name": "est_variance",
            "strides": [ 1, 3, 1, 1 ],
            "uid": 6,
            "virtual": false
        }
    ]
}
```

## Checklist


cmake/Dependencies.cmake
```diff
@@ -192,6 +192,19 @@ function(_fetch_spdlog VERSION HASH)
     _mark_targets_as_system(${spdlog_SOURCE_DIR})
 endfunction()
 
+# Doesn't conform with the others and ignores the VERSION and HASH arguments, but this will change very soon
```
BrianHarrisonAMD, line 195: Why? hehe

sdk/CMakeLists.txt
```diff
@@ -79,18 +81,18 @@ configure_file(
...
     EXPORT hipdnn_sdk_targets
 )
 
 export(
-    TARGETS hipdnn_sdk FlatBuffers spdlog_header_only
+    TARGETS hipdnn_sdk FlatBuffers spdlog_header_only nlohmann_json
     FILE  ${CMAKE_BINARY_DIR}/lib/cmake/hipdnn_sdk/hipdnn_sdkConfig.cmake
 )
 
 #todo, get spdlog to install its headers
```
BrianHarrisonAMD: Not from this PR, but I think this was todone right?

sdk/include/hipdnn_sdk/test_utilities/FlatbufferGraphTestUtils.hpp
```diff
@@ -8,7 +8,7 @@
 #include <hipdnn_sdk/data_objects/graph_generated.h>
 #include <hipdnn_sdk/plugin/PluginApiDataTypes.h>
 
-namespace hipdnn_backend::test_utilities
+namespace hipdnn_sdk::test_utilities
```
BrianHarrisonAMD, line 13: Ah, I was wondering why. 
Unfortunate.

sdk/tests/utilities/TestJson.cpp
BrianHarrisonAMD, line 1: I think we should expand the tests to cover all the node types, and also some with optional elements.
(Don't use BN inference though since those are not optional soon.)

sdk/include/hipdnn_sdk/utilities/Json.hpp
BrianHarrisonAMD, line 1: I think we should probably split this up into a few files.
It's not too bad right now, but I expect soon it'll be too long hehe.

BrianHarrisonAMD (commented):
Looks good to me, but a couple minor comments that I think are worth doing.

cmake/Dependencies.cmake
jerehartAMD: When we migrate to TheRock, as far as I can tell, it will manage fetching dependencies itself, and we'll just be linking to them. The URL fetch isn't obvious to me how you'd incorporate the HASH (if you even could), although the VERSION is a little more obvious. I can't imagine us changing the version before then, so I thought it would be fine to leave this hardcoded until it's replaced with something else

sdk/CMakeLists.txt
jerehartAMD: I'm not sure I follow, was this dependency added in another PR?

sdk/tests/utilities/TestJson.cpp
jerehartAMD: Adding the json serialization for the other node types too? Currently the only one implemented is the BatchnormInference, but it shouldn't be hard to implement the others too

sdk/include/hipdnn_sdk/utilities/Json.hpp
jerehartAMD: I was thinking about that too, I think it would make a lot of sense to have a json folder and have their filenames correspond to the components they're providing serialization for

cmake/Dependencies.cmake
BrianHarrisonAMD: Yea I think we probably leave this in case people want to build and use standalone generally, but we will want to have it use the bootstrapped env one when building in TheRock.

sdk/CMakeLists.txt
BrianHarrisonAMD: The code right near this todo looks to be doing it hehe.

BrianHarrisonAMD:
```
install(
    DIRECTORY ${HIP_DNN_SDK_INCLUDE_DIR}/ ${HIP_DNN_FLATBUFFERS_INCLUDE_DIR}/ ${HIP_DNN_SPDLOG_INCLUDE_DIR}/
    DIRECTORY ${HIP_DNN_SDK_INCLUDE_DIR}/ ${HIP_DNN_FLATBUFFERS_INCLUDE_DIR}/ ${HIP_DNN_SPDLOG_INCLUDE_DIR}/ ${HIP_DNN_NLOHMANN_JSON_INCLUDE_DIR}/
    DESTINATION ${HIP_DNN_SDK_INSTALL_INCLUDE_DIR}
)
```

sdk/tests/utilities/TestJson.cpp
BrianHarrisonAMD: Ah yea, sorry I thought we would add it for the others as well since it's small. 
I guess since this is a POC we can leave it as is for now, or we could add another to test how optional serializes.

sdk/CMakeLists.txt
jerehartAMD: Ahh, I totally missed reading the comment! It looks like it is done, I'll remove that

sdk/tests/utilities/TestJson.cpp
jerehartAMD: It's easy enough I'll add the others! My original intent was to get a vertical slice and get review feedback before adding the others, but if it looks good I might as well do the others now

BrianHarrisonAMD (approved):
LGTM

(You can merge whenever you are done with the small remaining changes)

sdk/include/hipdnn_sdk/utilities/json/BatchnormInferenceAttributes.hpp
```diff
@@ -0,0 +1,42 @@
...
+// SPDX-License-Identifier:  MIT
+#pragma once
+
+#include <hipdnn_sdk/data_objects/batchnorm_inference_attributes_generated.h>
+#include <hipdnn_sdk/utilities/json/Common.hpp>
+
+namespace hipdnn_sdk::data_objects
+{
+// NOLINTNEXTLINE(readability-identifier-naming)
+void to_json(nlohmann::json& batchnormJson, BatchnormInferenceAttributes const& bn)
```
SamuelReeder, line 11: Curious why these function names are snake_case?

jerehartAMD: nlohmann::json will find these functions and use them to create automatic constructors to json. That's what lets me do direct assignments like `jsonEntry["batchnorm_node"] = BatchnormInferenceAttributes{};`

SamuelReeder (approved):
LGTM

PR #221, opened 2025-09-19
Remove optional tensors from BN inference

BrianHarrisonAMD:
## Proposed changes

Remove optional tensors from BN fwd inference schema, and update relevant areas.
Frontend API was treating them as non-optional, and having them optional was left over from the original POC work.

## Checklist

### DIFF
--- a/plugins/miopen_legacy_plugin/engines/plans/MiopenBatchnormFwdInferencePlan.cpp
+++ b/plugins/miopen_legacy_plugin/engines/plans/MiopenBatchnormFwdInferencePlan.cpp
@@ -18,16 +18,9 @@ BatchnormFwdInferenceParams::BatchnormFwdInferenceParams(
     , _y(miopen_utils::createTensor(tensorMap, attributes.y_tensor_uid()))
     , _scale(miopen_utils::createTensor(tensorMap, attributes.scale_tensor_uid()))
     , _bias(miopen_utils::createTensor(tensorMap, attributes.bias_tensor_uid()))
+    , _estMean(miopen_utils::createTensor(tensorMap, attributes.mean_tensor_uid()))
+    , _estVariance(miopen_utils::createTensor(tensorMap, attributes.inv_variance_tensor_uid()))
 {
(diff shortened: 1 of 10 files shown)


plugins/miopen_legacy_plugin/engines/plans/MiopenBatchnormFwdInferencePlan.cpp
```diff
@@ -18,16 +18,9 @@ BatchnormFwdInferenceParams::BatchnormFwdInferenceParams(
     , _y(miopen_utils::createTensor(tensorMap, attributes.y_tensor_uid()))
     , _scale(miopen_utils::createTensor(tensorMap, attributes.scale_tensor_uid()))
     , _bias(miopen_utils::createTensor(tensorMap, attributes.bias_tensor_uid()))
+    , _estMean(miopen_utils::createTensor(tensorMap, attributes.mean_tensor_uid()))
+    , _estVariance(miopen_utils::createTensor(tensorMap, attributes.inv_variance_tensor_uid()))
```
SamuelReeder, line 22: Huh, this is strange that we call it `_estVariance` instead of `_invVariance`.

SamuelReeder (approved):
LGTM!

plugins/miopen_legacy_plugin/engines/plans/MiopenBatchnormFwdInferencePlan.cpp
SamuelReeder: I'd be in favour of changing this. Doesn't necessarily need `est` prefix either.

BrianHarrisonAMD: I think [MIOpen](https://github.com/ROCm/MIOpen/blob/develop_deprecated/include/miopen/miopen.h#L3042-L3059) calls it that in the API.

Since this is for the MIOpen plugin, I thought it made sense like this.

jerehartAMD (approved):
LGTM!

PR #222, opened 2025-09-19
Conv bwd data frontend & sdk integration

SamuelReeder:
## Proposed changes

Schema files, frontend headers, and tests for conv bwd data.

## Checklist


frontend/include/hipdnn_frontend/node/ConvolutionDgradNode.hpp
```diff
@@ -0,0 +1,358 @@
...
+            // dxDims[1] / wDims[1] is group count
+            // weightChannels = inputChannels / groups
+            // groups = inputChannels / weightChannels
+            auto groupCount = dxDims[1] / wDims[1];
+            HIPDNN_RETURN_IF_NE(
+                dxDims[1] % wDims[1],
+                0,
+                ErrorCode::INVALID_VALUE,
+                "ConvolutionDgradNode: dx tensor channels must be divisible by weight "
+                "tensor input channels");
```
BrianHarrisonAMD: Nit:

Probably makes more sense to flip these.
Error if they aren't divisible before doing it.

frontend/include/hipdnn_frontend/node/ConvolutionDgradNode.hpp
```diff
@@ -0,0 +1,358 @@
...
+
+            HIPDNN_RETURN_IF_TRUE(
+                dyStrides.empty(),
+                ErrorCode::ATTRIBUTE_NOT_SET,
+                "ConvolutionDgradNode: Cannot infer dx strides - missing dy strides");
+
+            HIPDNN_RETURN_IF_TRUE(
+                dxDimsFinal.empty(),
+                ErrorCode::ATTRIBUTE_NOT_SET,
+                "ConvolutionDgradNode: Cannot infer dx strides - missing dx dimensions");
```
BrianHarrisonAMD: Probably can skip this since it's guaranteed above, and we already check for matching size.

frontend/include/hipdnn_frontend/node/ConvolutionDgradNode.hpp
```diff
@@ -0,0 +1,358 @@
...
+
+            // Sort indices by their corresponding stride values (ascending)
+            std::sort(indices.begin(), indices.end(), [&dyStrides](size_t a, size_t b) {
+                return dyStrides[a] < dyStrides[b];
+            });
+
+            for(size_t i = 0; i < indices.size(); ++i)
+            {
+                strideOrder[indices[i]] = static_cast<int64_t>(i);
+            }
```
BrianHarrisonAMD: We should move this to the shape_utilties.

I feel like I have seen this logic before somewhere else.
Might be worth double checking if we can reduce duplication after moving to shape_utilities. 

This is basically figure out stride order from strides.
We might even have this already?

frontend/tests/TestConvolutionDgradNode.cpp
```diff
@@ -0,0 +1,988 @@
...
+    convAttributes.set_dy(std::make_shared<TensorAttributes>());
+    convAttributes.set_w(std::make_shared<TensorAttributes>());
+    GraphAttributes graphAttributes;
+    ConvolutionDgradNode node(std::move(convAttributes), graphAttributes);
+
+    auto error = node.infer_properties_node();
+    EXPECT_EQ(error.code, error_code_t::ATTRIBUTE_NOT_SET);
+}
+
+TEST(TestConvolutionDgradNode, InferPropertiesNode2DConvolutionSuccess)
```
BrianHarrisonAMD, line 220: We should double check that these are the same outputs you are getting in MIOpenDriver.
The math looked correct, but nice just to ensure our infer is working same as MIOpen's infer does.

BrianHarrisonAMD: Same for 3D.

frontend/tests/TestConvolutionDgradNode.cpp
```diff
@@ -0,0 +1,988 @@
...
+    // Check inferred strides
+    auto inferredStrides = dxTensor->get_stride();
+    EXPECT_EQ(inferredStrides.size(), 4);
+    EXPECT_EQ(inferredStrides[0], 2883); // N stride: 3 * 31 * 31 = 2883
+    EXPECT_EQ(inferredStrides[1], 961); // C stride: 31 * 31 = 961
+    EXPECT_EQ(inferredStrides[2], 31); // H stride: 31
+    EXPECT_EQ(inferredStrides[3], 1); // W stride: 1
+}
+
+TEST(TestConvolutionDgradNode, InferPropertiesNodeWithDilation2x2)
```
BrianHarrisonAMD, line 342: We need an infer with group as well I think.
Can do just stride infer, but dims represent a group conv.

sdk/schemas/convolution_common.fbs
BrianHarrisonAMD, line 1: Ah nice yea we needed to split.

tests/frontend/IntegrationConvolutionBackwardData.cpp
BrianHarrisonAMD, line 1: We really should unify this file since they are all copy pastas, but I guess we can do a follow-up to clean this up later if it's not trivial now.

BrianHarrisonAMD (approved):
LGTM with some minor comments.

frontend/include/hipdnn_frontend/node/ConvolutionDgradNode.hpp
SamuelReeder: We have the inverse, but strider order -> strides was needed still.

BrianHarrisonAMD: I swear we do this somewhere else hehe.

frontend/tests/TestConvolutionDgradNode.cpp
SamuelReeder: I'm so far finding it's not easy to 1-to-1 align the test to the driver. The driver seems to treat dy as output and so it only infers the dy dims. The `in_` prefixed parameters describe the dx tensor.

Basically this is where it calls the API:
```c
rc = miopenConvolutionBackwardData(GetHandle(),
                                           &alpha,
                                           outputTensor, // this is supposed to be where dy is passed, but dy should be an input tensor
                                           dout.GetDevicePtr(),
                                           weightTensor,
                                           wei.GetDevicePtr(),
                                           convDesc,
                                           algo,
                                           &beta,
                                           inputTensor, // this is supposed to be where dx is passed, but dx should be an output tensor
                                           ...
```
So the dimensions are only inferred for the `outputTensor` (dy) which is not what we want.

SamuelReeder: The calculation appears the same though:
```c
out_c = wei_k / wDesc.GetVectorLength();

            for(int i = 0; i < spatial_dim; ++i)
            {
                out_spatial[i] = std::max<std::ptrdiff_t>(
                    1,
                    (ptrdiff_t(in_spatial[i]) -
                     (1 + GetConvDilations()[i] * (std::ptrdiff_t(wei_spatial[i]) - 1)) +
                     2 * static_cast<std::ptrdiff_t>(GetConvPads()[i])) /
                            GetConvStrides()[i] +
                        1);
            }
```

PR #226, opened 2025-09-23
Swap to cxx17 instead of cxx20 standard

BrianHarrisonAMD:
## Proposed changes

Updating hipDNN to use CXX17 instead of CXX20 to ensure clients still on CXX17 can consume frontend header API and SDK.

## Checklist


backend/src/HipdnnBackend.cpp
BrianHarrisonAMD, line 1: Need to double check the extra logging params don't look silly.

frontend/tests/TestTensorValueAttributes.cpp
BrianHarrisonAMD, line 1: Changed this file a bit.
It was using std::numbers which isn't allowed for C++17, and then it also didn't really need the different values.

Swapped to just constant values for the tests, and removed usage of cmath.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/GenericBatchnormExecutor.hpp
```diff
@@ -32,14 +32,16 @@ class IGenericBatchnormExecutor
...
+          hipdnn_sdk::data_objects::DataType MeanVarianceDataTypeEnum>
 class BatchnormExecutor : public IGenericBatchnormExecutor
 {
-public:
-    using InputDataType = DataTypeToNative<Key.inputDataType>;
-    using ScaleBiasDataType = DataTypeToNative<Key.scaleBiasDataType>;
-    using MeanVarianceDataType = DataTypeToNative<Key.meanVarianceDataType>;
+    using InputDataType = DataTypeToNative<InputDataTypeEnum>;
+    using ScaleBiasDataType = DataTypeToNative<ScaleBiasDataTypeEnum>;
+    using MeanVarianceDataType = DataTypeToNative<MeanVarianceDataTypeEnum>;
```
BrianHarrisonAMD, line 42: I guess technically this changed (due to public location), but I don't think it matters.

sdk/include/hipdnn_sdk/utilities/ScopedResource.hpp
```diff
@@ -34,18 +38,12 @@ class ScopedResource
...
-        if(other._empty)
-        {
-            return;
-        }
-
-        _resource = std::move(other._resource);
-        _destructor = std::move(other._destructor);
         other._empty = true;
-        _empty = false;
     }
```
BrianHarrisonAMD, line 47: I guess this could be flipped back to the old way now that I fixed the move issue with std::function, but I think this version is preferrable.

sdk/include/hipdnn_sdk/utilities/UtilsBfp16.hpp
```diff
@@ -3,6 +3,9 @@
 
 #pragma once
 
+// Workaround: Operator<< in hip_bfloat16 is missing the std::ostream type info without this include.
+#include <ostream>
+
 #include <hip/hip_bfloat16.h>
```
BrianHarrisonAMD, line 9: This is a bit unfortunate.
For some reason when I swapped to cxx17, I was getting std::ostream errors without forcing this include order.

Thinking ill just leave this since we are planning on swapping bfloat16 headers once we are in the mono-repo, and we can revisit if this is still an issue.

tests/frontend/IntegrationConvForward.cpp
```diff
@@ -319,6 +319,8 @@ INSTANTIATE_TEST_SUITE_P(
     // Provide a custom name for each test instance
     [](const ::testing::TestParamInfo<IntegrationTestCase>& info) {
         std::string name = info.param.description;
-        std::ranges::replace_if(name, [](char c) { return !std::isalnum(c); }, '_');
+        std::transform(name.begin(), name.end(), name.begin(), [](char c) {
```
BrianHarrisonAMD: cbegin()

backend/src/HipdnnBackend.cpp
SamuelReeder: Have you tried doing `, ##__VA_ARGS__` instead? It might let us keep passing only two args since it should remove the preceding comma in those cases.

backend/tests/TestEnginePlugin.cpp
```diff
@@ -15,8 +15,8 @@
 
 using namespace hipdnn_backend;
 
-template <typename T, typename Destructor>
-using ScopedResource = hipdnn_sdk::utilities::ScopedResource<T, Destructor>;
+template <typename T>
+using ScopedResource = hipdnn_sdk::utilities::ScopedResource<T>;
```
adickin-amd, line 19: was the destructor just not being used here?

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuBatchnormForwardInference.cpp
```diff
@@ -353,14 +353,14 @@ std::vector<Batchnorm2dTestCase> getBnFwdInferenceTestCases()
...
     return {
-        {.n = 1, .c = 3, .h = 14, .w = 14, .seed = seed},
-        {.n = 1, .c = 256, .h = 1, .w = 1, .seed = seed},
-        {.n = 2, .c = 3, .h = 1, .w = 1, .seed = seed},
-        {.n = 32, .c = 1, .h = 14, .w = 14, .seed = seed},
-        {.n = 32, .c = 3, .h = 1, .w = 14, .seed = seed},
-        {.n = 32, .c = 3, .h = 14, .w = 1, .seed = seed},
-        {.n = 64, .c = 64, .h = 112, .w = 112, .seed = seed},
-        {.n = 64, .c = 512, .h = 14, .w = 14, .seed = seed},
+        {1, 3, 14, 14, seed},
```
adickin-amd, line 356: what a bummer, this was so much better named.

sdk/include/hipdnn_sdk/test_utilities/cpu_graph_executor/GenericBatchnormExecutor.hpp
adickin-amd: yea its fine, i also delete this whole file soon anyway.

adickin-amd (approved):
nothing to crazy in here.  LGTM

backend/tests/TestEnginePlugin.cpp
adickin-amd: Ah, i see the change below.

backend/src/descriptors/VariantDescriptor.cpp
```diff
@@ -46,7 +46,7 @@ void VariantDescriptor::getAttribute(hipdnnBackendAttributeName_t attributeName,
                       "VariantDescriptor::getAttribute(): elementCount is null");
         *elementCount
             = std::min<int64_t>(requestedElementCount, static_cast<int64_t>(_dataPointers.size()));
```
jerehartAMD, line 48: Wouldn't hurt to have ~~an assert~~ a throw here or something that ensures that `requestedElementCount` isn't negative. Speaking of which, why is this an int64_t?

frontend/tests/TestTensorValueAttributes.cpp
```diff
@@ -1,24 +1,25 @@
...
 #include <hipdnn_sdk/data_objects/tensor_attributes_generated.h>
 #include <limits>
-#include <numbers>
 #include <vector>
 
 // using namespace hipdnn_frontend::graph;
 using namespace hipdnn_sdk::data_objects;
 
+constexpr float PIFLOAT = 3.14159265358979323846f;
+constexpr double PIDOUBLE = 3.14159265358979323846;
```
jerehartAMD: We should probably have these constants defined in a file so we can use them elsewhere

Nit: Maybe PI_FLOAT and PI_DOUBLE instead?

frontend/tests/TestTensorValueAttributes.cpp
mousdahl-amd: Seems fine to me

sdk/include/hipdnn_sdk/utilities/UtilsBfp16.hpp
mousdahl-amd: Leave a note in the follow-on bfloat16 story

plugins/miopen_legacy_plugin/integration_tests/IntegrationGpuBatchnormForwardInference.cpp
mousdahl-amd: Yeah, I'm sad about this.

backend/src/HipdnnBackend.cpp
```diff
@@ -22,7 +22,7 @@ using namespace hipdnn_backend;
     HIPDNN_LOG_INFO("API called: [{}] " format, __func__ __VA_OPT__(, ) __VA_ARGS__)
 
 #define LOG_API_SUCCESS(func_name, format, ...) \
```
mousdahl-amd, line 24: Could we just split this into a single-arg version `LOG_API_SUCCESS()` and a multi-arg version `LOG_API_SUCCESS_ARGS()`

mousdahl-amd (approved):
I'd like to see another swing at LOG_API_SUCCESS, but I'm not devastated if you don't.  LGTM otherwise.

sdk/include/hipdnn_sdk/utilities/ScopedResource.hpp
jerehartAMD: The code is definitely more readable IMO. The only way the other would be better would be if the moves were costly, which they shouldn't be given that the only way you save on the moves is if you've already moved the function and resource out

jerehartAMD (approved):
LGTM, though I am curious about that int64_t thing

backend/src/descriptors/VariantDescriptor.cpp
adickin-amd: Likely because public API for hipdnn_backend is int64_t so we end up using that type.

adickin-amd: https://github.com/ROCm/hipDNN/blob/develop/backend/include/hipdnn_backend.h#L188-L194

backend/src/HipdnnBackend.cpp
BrianHarrisonAMD: I did have that, but didn't really like having the different macros.
I wish you could overload.

Ill see if there is a better way to handle it though.

frontend/tests/TestTensorValueAttributes.cpp
BrianHarrisonAMD: They are just random values really in this file for testing.
I hope we don't need PI_FLOAT & PI_DOUBLE generally hehe.

Ill do the rename, but I thought that was flagging when I tried it originally.
Let me see. constexpr are a bit finicky in our current version.

backend/src/HipdnnBackend.cpp
```diff
@@ -22,7 +22,7 @@ using namespace hipdnn_backend;
     HIPDNN_LOG_INFO("API called: [{}] " format, __func__ __VA_OPT__(, ) __VA_ARGS__)
```
SamuelReeder: Why wasn't this one changed too? I thought `__VA_OPT__(, )` had to be removed.

SamuelReeder (approved):
LGTM but I think you may have to update the other backend logging macro. This is a sad PR :(

backend/src/HipdnnBackend.cpp
BrianHarrisonAMD: Huh

BrianHarrisonAMD: I think it's because everywhere was providing additional args.

BrianHarrisonAMD: Ill remove it just to make it clear.

